PR #19, opened 2018-10-20
Default start to ./dcrcli -l

### LINKED ISSUE #18: Display command list when ran with no parameters
```
[x@rose dcrcli]$ ./dcrcli
No command specified
Usage:
  dcrcli [OPTIONS] <command> <args...>

Specify -h to show available options
Specify -l to list available commands
```

When ran with no parameters, what is currently being output with -l  should be shown here.

Jujhar:
If no arguments are passed it will default to listcommands output.

Fixes #18

### DIFF
--- a/main.go
+++ b/main.go
@@ -68,6 +68,10 @@ func main() {
 		os.Exit(1)
 	}
 
+	// get an instance of walletrpcclient without connecting
+	// because there are commands we can run at this stage without needing to connect to dcrwallet
+	client := walletrpcclient.New()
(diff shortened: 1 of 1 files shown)


yaowork (changes requested):
I'm skeptical about instantiating the client within an if block. you might want to do that outside the block, and run the `listcommands` command within the block

Jujhar:
Updated.

yaowork (approved):

PR #29, opened 2018-10-29
Restructure app to allow for multiple entry points

yaowork:
This restructures the application creating an abstraction layer such that commands can be called from cli mode and web mode providing the same interface for commands and response


main.go
```diff
@@ -68,6 +68,9 @@ func main() {
 		os.Exit(1)
 	}
 
+	params := make(map[string]interface{}, 1)
+	params["caller"] = "cli"
```
itswisdomagain: Why is `params` a map? Why not just a simple string that informs the source of the call (aka `caller`)? Do you anticipate other data being passed into `params`?

main.go
```diff
@@ -97,7 +100,7 @@ func main() {
 	if !serveHTTP {
 		command = args[0]
 		if args[0] == "listcommands" {
-			res, err := client.RunCommand("listcommands", nil)
+			res, err := client.RunCommand("listcommands", nil, params)
```
itswisdomagain: Does the `listcommands` handler need to know the "caller"? Currently, the handler for this command returns a Response object with the array of available commands; and this response can be processed by cli/http modules to display proper feedback to the user. Don't see a need to pass params to this handler.

Compare the implementation of the `listcommands` handler in #28

server/handlers.go
```diff
@@ -25,7 +25,7 @@ func (s *Server) renderError(w http.ResponseWriter, req *http.Request, err error
 }
 
 func (s *Server) accounts(w http.ResponseWriter, req *http.Request) {
-	res, err := s.walletClient.RunCommand("accounts", nil)
+	res, err := s.walletClient.RunCommand("accounts", nil, nil)
```
itswisdomagain: Currently, all web server calls to `walletClient.RunCommand` passes nil as params. And when this is done, `"caller"` defaults to "cli". If the command handlers always end up using `caller = "cli"`, then `caller` and the additional `params` parameter are not required.

Ideally, if we can find a way to make the command handlers work without knowing the caller, I think it'd make it easier for us to introduce other "callers" going forward without having to modify the implementation of each handler to account for the newly added caller.

Because I suspect the key reason for passing `params` to each handler is to enable the handler know the source of the call and perform its operations accordingly depending on "who" the caller is.

server/handlers.go
```diff
@@ -46,29 +46,11 @@ func (s *Server) balance(w http.ResponseWriter, req *http.Request) {
 	}
 
 	opts := []string{accountNumber, minConf}
-	res, err := s.walletClient.RunCommand("balance", opts)
+	res, err := s.walletClient.RunCommand("balance", opts, nil)
```
itswisdomagain: The `opts` array passed here is ignored in the implementation for the balance command handler. See line 191 of `walletrpcclient/walletrpcclient.go`

server/handlers.go
```diff
@@ -46,29 +46,11 @@ func (s *Server) balance(w http.ResponseWriter, req *http.Request) {
...
-func (s *Server) overview(w http.ResponseWriter, req *http.Request) {
-	res, err := s.walletClient.RunCommand("overview", nil)
-	if err != nil {
-		s.renderError(w, req, err)
-		return
-	}
-	s.renderJSON(w, req, res)
-}
-
-func (s *Server) version(w http.ResponseWriter, req *http.Request) {
```
itswisdomagain, line 67: Why are we taking this out? Won't hurt to have a page in the web app showing users the current version of dcrwallet?

walletrpcclient/walletrpcclient.go
```diff
@@ -90,242 +82,129 @@ func (c *Client) IsCommandSupported(command string) bool {
 // RunCommand takes a command and tries to call the appropriate handler to call a gRPC service
 // This should only be called after verifying that the command is supported using the IsCommandSupported
 // function.
-func (c *Client) RunCommand(command string, opts []string) (*Response, error) {
+func (c *Client) RunCommand(command string, args []string, params map[string]interface{}) (*Response, error) {
```
itswisdomagain: I want to propose we use consistent variable names for `args`/`opts`. Can we stick to `commandArgs` everywhere so its immediately obvious that this variable holds additional arguments for the command about to be proceed?

walletrpcclient/cmd.go
```diff
@@ -0,0 +1,185 @@
...
+		Result: [][]interface{}{
+			[]interface{}{r.VersionString},
+		},
+	}
+	return res, nil
+}
+
+// listCommands lists all supported commands
+// requires no parameter
+func (c *Client) cmdListCommands(ctx context.Context) (*Response, error) {
```
itswisdomagain: `ctx` is never used. No need to create and pass a context that's never used

walletrpcclient/walletrpcclient.go
```diff
@@ -90,242 +82,129 @@ func (c *Client) IsCommandSupported(command string) bool {
...
-func (c *Client) RegisterHandler(key, command, description string, h Handler) {
-	if _, ok := c.funcMap[key]; ok {
-		panic("trying to register a handler twice: " + key)
-	}
-
-	c.funcMap[key] = h
-	c.commands[key] = command
-	c.descriptions[key] = description
+// listCommands calls the cmdListCommands. This requires no parameters to run
+func (c *Client) listCommands(args []string, params map[string]interface{}) (*Response, error) {
```
itswisdomagain: Since this requires no params to run, can we take off the arguments in the function signature? Compare the implementation of the listcommands handler in #28

walletrpcclient/walletrpcclient.go
```diff
@@ -90,242 +82,129 @@ func (c *Client) IsCommandSupported(command string) bool {
...
-	if err != nil {
-		return nil, fmt.Errorf("error signing transaction: %s", err.Error())
-	}
-
-	// publish transaction
-	pReq := &pb.PublishTransactionRequest{
-		SignedTransaction: sRes.Transaction,
+	if caller == "cli" {
+		if len(args) == 0 {
+			return nil, errors.New("command 'receive' requires at least 1 param. 0 found \nUsage:\n  receive \"accountnumber\"")
```
itswisdomagain: I'm thinking of an idea where this function behaves like the send command, displaying a list of accounts and prompting for a selection if no account number is passed. But should **definitely belong in a different PR**, just thought to mention

walletrpcclient/walletrpcclient.go
```diff
@@ -90,242 +82,129 @@ func (c *Client) IsCommandSupported(command string) bool {
...
-	obj := qrcodeTerminal.New()
-	obj.Get(r.Address).Print()
-
-	return res, nil
+func (c *Client) balance(args []string, params map[string]interface{}) (*Response, error) {
+	return c.cmdBalance(context.Background())
 }
 
-// walletVersion fetches and returns version of wallet we are connected to
-func (c *Client) walletVersion(ctx context.Context, opts []string) (*Response, error) {
```
itswisdomagain, line 313: Please let's keep this. Unless there's an advantage to taking it away...

itswisdomagain (changes requested):
Hi @codemaestro64, please give thoughts to this. To confirm, the major change this PR brings to the table is the caller functionality that allows different command handlers "behave" differently depending on the caller. Am I correct in my deduction?

main.go
yaowork: depending on the caller, other data can be passed to the called function through the map

walletrpcclient/walletrpcclient.go
yaowork: @raedah  wants it taken off

server/handlers.go
yaowork: all code in the server module are being refactored.

main.go
yaowork: yes. at some point in development. The listcommands shouldn't function in web ad desktop mode

walletrpcclient/walletrpcclient.go
yaowork: all of those functions have that signature. can't take of the arguments

server/handlers.go
yaowork: server module is a WIP

walletrpcclient/walletrpcclient.go
itswisdomagain: Oh. Okay then

server/handlers.go
yaowork: @raedah 's instructions

walletrpcclient/walletrpcclient.go
```diff
@@ -106,230 +99,129 @@ func (c *Client) IsCommandSupported(command string) bool {
...
-func (c *Client) RegisterHandler(key, command, description string, h Handler) {
-	if _, ok := c.funcMap[key]; ok {
-		panic("trying to register a handler twice: " + key)
-	}
-
-	c.funcMap[key] = h
-	c.commands[key] = command
-	c.descriptions[key] = description
+// listCommands calls the cmdListCommands. This requires no parameters to run
+func (c *Client) listCommands(commandArgs []string, params map[string]interface{}) (*Response, error) {
```
itswisdomagain: There's another function like this in this package: `ListSupportedCommands`, and seeing as the arguments here are not used, I recommend taking this off. I guess the reason we have this, is because this function is registered as a command.

I don't think `-l` is a command like `send` is, usually `-abc` are flags/options. Same reason we won't list `-h` as a command. So this function can go, and the code at the bottom of this file that registers this function as a command.

If you run `dcrcli -h`, `-l` is listed there as an option/flag

walletrpcclient/walletrpcclient.go
```diff
@@ -106,230 +99,129 @@ func (c *Client) IsCommandSupported(command string) bool {
...
+		if len(commandArgs) == 0 {
+			return nil, errors.New("command 'receive' requires at least 1 param. 0 found \nUsage:\n  receive \"accountnumber\"")
+		}
+		acc, err := strconv.ParseUint(commandArgs[0], 0, 32)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing account number. err:%s", err.Error())
+		}
+		accountNumber = uint32(acc)
+	} else {
+		if acc, ok := params["accountNumber"]; ok {
```
itswisdomagain: Remember my comment earlier:
>Ideally, if we can find a way to make the command handlers work without knowing the caller, I think it'd make it easier for us to introduce other "callers" going forward without having to modify the implementation of each handler to account for the newly added caller

The cli sends command arguments in an array. For this handler, line 129 above reads the account number from the `commandArgs` array at position 0. Can't we maintain that for the web server and any other module that calls the handlers?

The web server already follows this pattern, line 48-49 of `server/handlers.go` uses an array to send arguments to the `balance` command.

Following this line of thought, I'm thinking we create 3 different packages: `core`, `cli` and `server`. The `core` would be like the current `walletrpcclient` package which will primarily be concerned about interacting with the dcrwallet rpc server. The `core` package should have a fixed way it receives arguments for each command and a fixed response it returns.

Both the `cli` and `server` package interact with the `core` package, passing in the required fields/values and processing the response received.

The `cli` package gets user input from the terminal, passes the input received to the `core` package in the format the `core` package understands, receives response from the `core` package and writes messages/prompts/outputs back to the terminal.

The `server` package gets user input from the web app interface (mostly through forms), passes the input received to the `core` package in the format the `core` package understands, receives response from the `core` package and displays outputs to the user in the web page.

This way, we make the `core` package "caller-agnostic" and we can add more "callers" like the desktop interface on the fly.

main.go
```diff
@@ -74,29 +72,30 @@ func enterHttpMode(config *config) {
 }
 
 func enterCliMode(config *config, args []string) {
-	client := walletrpcclient.New()
+	//client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
+	cli := cli.New()
```
itswisdomagain: variable name `cli` clashes with the name of the imported `cli` package

main.go
```diff
@@ -74,29 +72,30 @@ func enterHttpMode(config *config) {
...
 		invalidCommandReceived(command)
 		os.Exit(1)
 	}
 
-	cliExecuteCommand(client, command, config, args)
+	cliExecuteCommand(cli, command, config, args)
 }
 
-func cliExecuteCommand(client *walletrpcclient.Client, command string, config *config, args []string) {
+func cliExecuteCommand(cli *cli.CLI, command string, config *config, args []string) {
```
itswisdomagain: this function is completely related to the cli package, that's where it belongs. Access this function by calling cli.ExecuteCommand rather than passing the cli instance into this function

main.go
```diff
@@ -106,52 +105,14 @@ func cliExecuteCommand(client *walletrpcclient.Client, command string, config *c
...
 
-func showAvailableCommands(client *walletrpcclient.Client) {
-	commands := client.ListSupportedCommands()
-	printResult(commands)
+func showAvailableCommands(cli *cli.CLI) {
+	cli.RunCommand("listcommands", nil, nil)
 }
 
-func noCommandReceived(client *walletrpcclient.Client) {
+func noCommandReceived(client *cli.CLI) {
```
itswisdomagain: This function too, should go into the cli package,. Access this function by calling cli.NoCommandReceived rather than passing the cli instance into this function

main.go
```diff
@@ -74,29 +72,30 @@ func enterHttpMode(config *config) {
...
+	cli := cli.New()
 
 	if len(args) == 0 {
-		noCommandReceived(client)
+		noCommandReceived(cli)
 		os.Exit(0)
 	}
 
 	command := args[0]
 	if command == "-l" {
```
itswisdomagain: while it's okay to check if no command was passed in (aka `len(args) == 0`, I don't think processing **any** cli command should be done in main.go, the cli command (`args[0]`) and any additional command args (`args[1:]`) should be passed to the `cli` package via `cli.ExecuteCommand`

main.go
```diff
@@ -74,29 +72,30 @@ func enterHttpMode(config *config) {
...
-		showAvailableCommands(client)
+		showAvailableCommands(cli)
 	}
 
-	if !client.IsCommandSupported(command) {
+	if !cli.IsCommandSupported(command) {
```
itswisdomagain: same here, checking whether a command is valid or not belongs in the cli package. Anything that has to do with **processing** a cli command should please go in the cli package

main.go
```diff
@@ -106,52 +105,14 @@ func cliExecuteCommand(client *walletrpcclient.Client, command string, config *c
...
-	}
-
-	w.Flush()
+	cli.RunCommand(command, commandArgs, client)
 }
 
-func showAvailableCommands(client *walletrpcclient.Client) {
-	commands := client.ListSupportedCommands()
-	printResult(commands)
+func showAvailableCommands(cli *cli.CLI) {
```
itswisdomagain: 1. this function belongs in the cli package as I've explained for other similar cases
2. why not just call `cli.RunCommand("listcommands", nil, nil)` directly?

cli/cli.go
```diff
@@ -0,0 +1,231 @@
...
+		descriptions: make(map[string]string),
+	}
+
+	// register handlers
+	client.registerHandlers()
+
+	return client
+}
+
+func (c *CLI) RunCommand(command string, commandArgs []string, walletrpcclient *walletrpcclient.Client) {
```
itswisdomagain: `walletrpcclient` is registered as a property of `CLI`, see the CLI struct definition in line  25 above. CLI.walletrpcclient should be set in `func New()` and then accessed here as `c.walletrpcclient`

itswisdomagain: alternatively, take off the property declaration in line 25

cli/cli.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func (c *CLI) registerHandlers() {
+	c.RegisterHandler("listcommands", "-l", "List all supported commands", c.listCommands)
+	c.RegisterHandler("receive", "receive", "Generate address to receive funds", c.receive)
+	c.RegisterHandler("send", "send", "Send DCR to address. Multi-step", c.send)
+	c.RegisterHandler("balance", "balance", "Check balance of an account", c.balance)
+}
+
+func (c *CLI) printResult(res *response) {
```
itswisdomagain: this func does not need to extend CLI as `c` is not used, it can just be an ordinary func in this package

walletrpcclient/walletrpcclient.go
```diff
@@ -168,168 +87,118 @@ func (c *Client) sendTransaction(ctx context.Context, opts []string) (*Response,
...
+	response := &SendResult{
+		TransactionHash: string(publishResponse.TransactionHash),
 	}
 
-	res.Result = [][]interface{}{resultRow}
+	return response, nil
+}
 
-	return res, nil
+type AccountBalanceResult struct {
```
itswisdomagain: I love the introduction of these types! Can we put all type definitions in a different file in this package (walletrpcclienttypes)?

itswisdomagain (changes requested):
Really, really nice work. Just a little more refactoring to firmly establish separation of concerns.

main.go
itswisdomagain: I'm thinking you could call the variable `cliInstance` or something

cli/cli.go
```diff
@@ -0,0 +1,261 @@
...
+	command := commandArgs[0]
+	if !c.IsCommandSupported(command) {
+		if command == "-l" {
+			command = "listcommands"
+		} else {
+			c.invalidCommandReceived(command)
+			os.Exit(1)
+		}
+	}
+	if command == "-l" {
```
itswisdomagain: lines 52-62 can be represented in a simpler way:
```

	if command == "-l" {
		command = "listcommands"
	} else if !c.IsCommandSupported(command) {
		c.invalidCommandReceived(command)
		os.Exit(1)
	}
```

itswisdomagain, line 52: lines 52-62 can be represented in a simpler way:
```

	if command == "-l" {
		command = "listcommands"
	} else if !c.IsCommandSupported(command) {
		c.invalidCommandReceived(command)
		os.Exit(1)
	}
```

main.go
```diff
@@ -74,90 +72,13 @@ func enterHttpMode(config *config) {
...
-func noCommandReceived(client *walletrpcclient.Client) {
-	fmt.Printf("usage: %s [OPTIONS] <command> [<args...>]\n\n", AppName)
-	fmt.Printf("available %s commands:\n", AppName)
-	showAvailableCommands(client)
-	fmt.Printf("\nFor available options, see '%s -h'\n", AppName)
-}
-
-func invalidCommandReceived(command string) {
-	fmt.Fprintf(os.Stderr, "%s: '%s' is not a valid command. See '%s -h'\n", AppName, command, AppName)
+	cli := cli.New(client, AppName)
```
itswisdomagain: This variable name clash is still unresolved, `cli` is used as a variable name and as a package name in the same scope.

itswisdomagain (changes requested):
Just 2 small changes and we're good. Just semantic corrections

itswisdomagain (approved):

PR #30, opened 2018-11-01
Web ui

yaowork:
This adds a web a web interface to the application.


web/server.go
```diff
@@ -0,0 +1,118 @@
...
+	}
+
+	for i, v := range tpls {
+		tpl, err := template.New(i).ParseFiles(v, layout)
+		if err != nil {
+			log.Fatalf("error loading templates: %s", err.Error())
+		}
+
+		s.lock.Lock()
+		s.templates[i] = tpl
```
itswisdomagain, line 61: This code is called once when the server is started. I don't see multiple goroutines trying to access this map; ergo don't see a need for mutex lock here.

web/server.go
```diff
@@ -0,0 +1,118 @@
...
+		s.lock.Unlock()
+	}
+}
+
+func (s *Server) render(tplName string, data map[string]interface{}, res http.ResponseWriter) {
+	s.lock.RLock()
+	defer s.lock.RUnlock()
+
+	if tpl, ok := s.templates[tplName]; ok {
+		err := tpl.Execute(res, data)
```
itswisdomagain, line 67: same here, template execution does not need mutex lock
if you take off lock from the server struct in line 22, and remove all its usages in lines 63,65,70 and 71, the server starts and runs fine. Unless there's something I'm missing?

web/server.go
```diff
@@ -0,0 +1,118 @@
...
+
+func (s *Server) loadTemplates() {
+	layout := "web/views/layout.html"
+	tpls := map[string]string{
+		"balance.html": "web/views/balance.html",
+		"send.html":    "web/views/send.html",
+		"receive.html": "web/views/receive.html",
+	}
+
+	for i, v := range tpls {
```
itswisdomagain, line 55: descriptive variable names is usually very helpful, `for key, value := range tpls` is okay but we can and should go the extra mile to add meaning to key and value, so I'm thinking `for name, path : range tpls`

itswisdomagain: and while we're here, can we use template names//keys without the `.html` extension?

itswisdomagain (changes requested):

itswisdomagain:
Some issues I noticed with the send functionality:

1. Send amount says unit is in DCR but I'm not sure it is. I had a balance of 200 DCR, sent 100 "DCR" (twice) and my balance is 199.999492 DCR.

2. The transaction hash displayed after sending is scrambled
screenshot 2018-11-03 at 10 20 06 am

3. Refreshing the page after a send transaction re-submits the previous request (even though it was clearly successful). Feels like bad UX. After a user sends a particular amount to a particular address **successfully**, if he opts to refresh the page, it's likely he wants to make another transaction not to repeat the previous transaction.

PR #33, opened 2018-11-12
Empty imported wallet fix

### LINKED ISSUE #32: Generate Receive Address Error with imported account
When dcrwallet creates a wallet on first run, the wallet has 2 accounts
- default (account number 0)
- imported (account number 2147483647)

Attempting to generate receive address for the imported account (for example via cli: `dcrcli receive 2147483647`) returns the following error: 

`wallet.NewExternalAddress: item does not exist: account 2147483647`

Jujhar:
Removed imported wallet if empty on flaky basis that system falsely automatically — always — creates an entry of an imported wallet even when no wallet is imported

Fixes #32

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -243,12 +243,6 @@ func (c *CLI) balance(commandArgs []string) (*response, error) {
 		result: make([][]interface{}, len(balances)),
 	}
 	for i, v := range balances {
-
-		// bug fix - hide imported wallet by default if no wallet imported
-		if v.AccountName == "imported" && v.Total == 0 {
-			break
-		}
-
 		res.result[i] = []interface{}{
 			v.AccountName,
 			v.Total,
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -243,6 +243,12 @@ func (c *CLI) balance(commandArgs []string) (*response, error) {
 		result: make([][]interface{}, len(balances)),
 	}
 	for i, v := range balances {
+
+		// bug fix - hide imported wallet by default if no wallet imported
+		if v.AccountName == "imported" && v.Total == 0 {
+			break
+		}
+
 		res.result[i] = []interface{}{
 			v.AccountName,
 			v.Total,
--- a/walletrpcclient/walletrpcclient.go
+++ b/walletrpcclient/walletrpcclient.go
@@ -128,7 +128,31 @@ func (c *Client) Balance() ([]AccountBalanceResult, error) {
 		return nil, fmt.Errorf("error fetching accounts: %s", err.Error())
 	}
 
+	// bug fix - hide imported wallet by default if no wallet imported (pt 1)
+	var minusRange = 0  // used to return correct number of wallet entries to make
+	for i, v := range accounts.Accounts {
+		req := &pb.BalanceRequest{
+			AccountNumber:         v.AccountNumber,
+			RequiredConfirmations: 0,
+		}
+		res, err := c.walletServiceClient.Balance(ctx, req)
+		if err != nil {
+			return nil, fmt.Errorf("error fetching balance for account: %d :%s", v.AccountNumber, err.Error())
+		}
+
+		if v.AccountName == "imported" && dcrutil.Amount(res.Total) == 0 {
+			minusRange = 1
+		}
+		i++
+	}
+
 	balanceResult := make([]AccountBalanceResult, len(accounts.Accounts))
+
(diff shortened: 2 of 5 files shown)


cli/cli.go
```diff
@@ -243,6 +243,12 @@ func (c *CLI) balance(commandArgs []string) (*response, error) {
 		result: make([][]interface{}, len(balances)),
 	}
 	for i, v := range balances {
+
+		// bug fix - hide imported wallet by default if no wallet imported
+		if v.AccountName == "imported" && v.Total == 0 {
```
itswisdomagain: Why not do this from the `walletrpcclient` package instead? The `Balance()` handler in that package is called multiple times from the web and cli packages. It's called when displaying the balance on both packages, when sending an amount on both interfaces and when generating a receive address on the web interface. Soon enough, it will be called when generating a receive address on the cli interface. And there could be other interfaces.

It would be better to discard the imported account at the source where all the interfaces connect to, that way we won't have to write same/similar code in multiple places to do that.

itswisdomagain (changes requested):

cli/cli.go
Jujhar: Updated

walletrpcclient/walletrpcclient.go
```diff
@@ -128,7 +128,31 @@ func (c *Client) Balance() ([]AccountBalanceResult, error) {
 		return nil, fmt.Errorf("error fetching accounts: %s", err.Error())
 	}
 
+	// bug fix - hide imported wallet by default if no wallet imported (pt 1)
+	var minusRange = 0  // used to return correct number of wallet entries to make
```
itswisdomagain: This is quite unnecessary as you can make a slice in go without knowing the actual length of the slice. More so, this attempt to determine the number of "invalid" accounts to disregard comes with an extra cost - you're making 2 RPCs for each account to get the account balance which introduces redundancy.

You can initialise the `balanceResult` slice without specifying the length, like so:
`balanceResult := make([]AccountBalanceResult, 0, len(accounts.Accounts))`

Then you can run your accounts loop once, get the balance and ignore accounts with name as "imported" and balance equal zero. The code sinppet below should work:

```
balanceResult := make([]AccountBalanceResult, 0, len(accounts.Accounts))

for _, v := range accounts.Accounts {
	req := &pb.BalanceRequest{
		AccountNumber:         v.AccountNumber,
		RequiredConfirmations: 0,
	}

	res, err := c.walletServiceClient.Balance(ctx, req)
	if err != nil {
		return nil, fmt.Errorf("error fetching balance for account: %d :%s", v.AccountNumber, err.Error())
	}

	if v.AccountName == "imported" && dcrutil.Amount(res.Total) == 0 {
		continue
	}

	accountBalance := AccountBalanceResult{
		AccountNumber:   v.AccountNumber,
		AccountName:     v.AccountName,
		Total:           dcrutil.Amount(res.Total),
		Spendable:       dcrutil.Amount(res.Spendable),
		LockedByTickets: dcrutil.Amount(res.LockedByTickets),
		VotingAuthority: dcrutil.Amount(res.VotingAuthority),
		Unconfirmed:     dcrutil.Amount(res.Unconfirmed),
	}

	balanceResult = append(balanceResult, accountBalance)
}
```

Notice also the usage of `continue` not `break` when you want to skip further processing in a loop and jump to the next item in the loop. Also notice there's no need for incrementing the loop index yourself with `i++`

I think with this, we should be good to go.

itswisdomagain: This article and SO answer might help to explain more about go slices
- https://blog.golang.org/go-slices-usage-and-internals
- https://stackoverflow.com/a/28636712

itswisdomagain (changes requested):

walletrpcclient/walletrpcclient.go
Jujhar: OK this kind of makes sense did not know you could do 0 like that.

itswisdomagain (approved):

itswisdomagain:
For future reference, the `imported` account has its purpose in dcrwallet. It's meant to hold details for accounts a user imports into the wallet.

However, because it's not part of the wallet's account, the wallet cannot generate a receive address for it.

But if a user does import an external account, he should be able to see the balance for such account and possibly create transactions from the imported account.

This PR does ensure that if the imported account's balance is > 0, the account will show up in the account balances section of the interfaces.

Issue #10, opened 2018-10-15
Several cleanups of how the send cmd works


oregonisaac:
1. send cmd asks for source account, should list off account names, show and use 'default' as the default account. 
2. send cmd Needs to return a readable error if input is not valid.  
3. send cmd, If error is shown, it should after that again show the input prompt. 
4. send cmd, I get 'error: Invalid addressDestination address'   even when inputting a valid address. Does it support testnet?  Could specify which net is being used in the input address question.

PR #36, opened 2018-11-13
Replace promptui with stdlib package functions.

### LINKED ISSUE #34: remove "github.com/manifoldco/promptui" dependency and replace its usages with standard packages

barisere:
Fixes #34


cli/cli.go
```diff
@@ -42,6 +42,12 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
 	return client
 }
 
+// RunCommand invokes the handler function registered for the given
+// command in `commandArgs`.
+//
+// If no command, or an unsupported command is passed to RunCommand,
+// the program exits with an error.
+// commandArgs[0] is the command to run. commandArgs[1:] are the arguments to the command.
 func (c *CLI) RunCommand(commandArgs []string) {
```
itswisdomagain, line 51: Nice work for adding comments for these public functions 👍

cli/prompter.go
```diff
@@ -0,0 +1,91 @@
...
+
+type promptOption struct {
+	Label    interface{}
+	Options  []interface{}
+	Default  string
+	Validate validatorFunc
+	Secure   bool
+}
+
+func (p promptOption) Prompt() (string, error) {
```
itswisdomagain: This doesn't make semantic sense, having an action `Prompt()` in an `Option` type. Might be more meaningful to make this function extend `prompter` instead.

cli/prompter.go
```diff
@@ -0,0 +1,91 @@
...
+	fmt.Println(p.Label)
+	for _, opt := range p.Options {
+		fmt.Println(opt)
+	}
+	input, err := readInput(label, p.Secure)
+	if err != nil {
+		return "", err
+	}
+	if p.Validate != nil {
+		if vErr := p.Validate(input); vErr != nil {
```
itswisdomagain: I'm thinking you can reuse the `err` variable declared earlier here, no need to create another variable...

cli/prompter.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func (p promptOption) Prompt() (string, error) {
+	label := ""
+	if p.Default != "" {
+		label += fmt.Sprintf("[default: %s]", p.Default)
+	}
+	label += "=> "
+	fmt.Println(p.Label)
+	for _, opt := range p.Options {
```
itswisdomagain: Adding a single space before this for loop makes this code a little more readable

itswisdomagain (changes requested):
Preliminary review. I'll share more thoughts in a bit

cli/helpers.go
```diff
@@ -17,26 +16,34 @@ func getSendSourceAccount(c *walletrpcclient.Client) (uint32, error) {
...
 		return 0, err
 	}
-
-	promptItems := []string{}
-	accountItems := map[string]uint32{}
-	for _, v := range accounts {
-		itemStr := fmt.Sprintf("%s (%s)", v.AccountName, dcrutil.Amount(v.Total).String())
-		promptItems = append(promptItems, itemStr)
-		accountItems[itemStr] = v.AccountNumber
+	validateInt := func(v string) error {
```
itswisdomagain: the validate function above is not complete, and you're not only validating if the input is an integer, you also should check if the int falls within the range of 0 to `len(accounts) - 1`.

cli/helpers.go
```diff
@@ -17,26 +16,34 @@ func getSendSourceAccount(c *walletrpcclient.Client) (uint32, error) {
...
+		}
+		return nil
+	}
+	promptItems := promptOption{
+		Label:    "Select source account",
+		Validate: validateInt,
+	}
+	accountItems := map[int]uint32{}
+	for idx, v := range accounts {
+		promptItems.Options = append(promptItems.Options, fmt.Sprintf("%3d. %s (%s)",
```
itswisdomagain: Please use a separate variable to hold these options and then assign to promptItems same way you did for the validate function above. Makes the code more readable

cli/helpers.go
```diff
@@ -17,26 +16,34 @@ func getSendSourceAccount(c *walletrpcclient.Client) (uint32, error) {
...
-		Label: "Select source account",
-		Items: promptItems,
+	result, err := promptItems.Prompt()
+	if err != nil {
+		return 0, fmt.Errorf("error getting selected account: %s", err.Error())
 	}
 
-	_, result, err := prompt.Run()
+	choice, err := strconv.Atoi(result)
 	if err != nil {
```
itswisdomagain, line 46: notice a partial repetition of the earlier validate command here. I want to recommend performing validation outside the `prompter.Prompt()` function

Read the input without validating and then validate here. Same thing for other places you need to call `Prompt()`. It'll make more sense to someone reading your code why you're performing the validation when you validate **after** receiving the input

More so, if you validate here, you'd be able to repeat the prompt if the user's input is invalid which is an important requirement (compare the comment on issue #10)

itswisdomagain (changes requested):

itswisdomagain:
Nice job @barisere. The code works almost as expected (the only missing requirement is repeating the prompt if a user provides a wrong input)

That said, I believe the code can be easier to interprete if you abstract just a little bit less and don't attempt to plan for features we may never need or scenarios we may never accomplish. I'm sure you know about the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)?

Here are some suggestions to improve the code quality. Doing these should take care of most if not all of the review comments I dropped earlier.

- rename `prompter` to `terminal-prompter` or something more descriptive (using descriptive names for files, variables, structs, interfaces, packages etc makes your code a lot more easier to understand)
- remove `Prompt()` function and replace with the following 3 functions (we can add more as the need arises, we don't want to plan for what we may never need)
- first function: `terminalPrompt.RequestInput(message, validateFn)`, displays message, waits for input and validates using the validateFn, if input fails validation, tell user the input was invalid (and why) then repeat prompt
- second function: `terminalPrompt.RequestInput(message, validateFn)`, much like the first function, except it masks the user's input
- third function: `terminalPrompt.RequestSelection(message, options, validateFn)`, display the message, then the options, and wait for input. Before running the provided validate function, first check if the user's selection is within the range of the options provided, if not, show an error message and run prompt again, else, run the validate function

barisere:
@itswisdomagain I've applied your suggestions. New commits added.

barisere:
Current behaviour (with the send command) uses ordinal values to select options

Current behaviour

Previous behaviour (with the send command) uses arrow navigation to select options.

Previous behaviour
Previous behaviour

itswisdomagain:
This looks like good work! See some thoughts below, I haven't looked at the code though. Will do so and leave more thoughts where appropriate
> Current behaviour

The spacing between the prompt `"Select source account"` and the options seems to suggest a disconnect for me. Look at the previous implementation, you'd notice the prompt message, instruction and the options all form a "paragraph", with no empty line between any of them. 

Whitespaces are generally used to indicate the end of a section and start of another not very related section. If they're not related at all, then the space is bigger, if there's some relationship but not very big, then a single space or no space will do. If they're very related, no need for white space.

I believe the only spacing that would make visual sense is the one after showing an error message, which does not currently exist in your implementation. The prompt message, options, user selection and any error arising from the user's input should be close together

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
 	"fmt"
 	"strconv"
 
-	"github.com/manifoldco/promptui"
+	"github.com/raedahgroup/dcrcli/cli/terminalprompt"
+
```
itswisdomagain: why the space though?

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
 	"github.com/raedahgroup/dcrcli/walletrpcclient"
```
itswisdomagain: same space here, there are not so many imports, I think doing away with the empty line spacing won't harm readability

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
-	promptItems := []string{}
-	accountItems := map[string]uint32{}
-	for _, v := range accounts {
-		itemStr := fmt.Sprintf("%s (%s)", v.AccountName, dcrutil.Amount(v.Total).String())
-		promptItems = append(promptItems, itemStr)
-		accountItems[itemStr] = v.AccountNumber
+	// validateInput ensures that the input received is a number that corresponds to an account
+	validateInput := func(value string) error {
+		if choice, err = strconv.Atoi(value); err != nil {
+			return fmt.Errorf("could not recognize input: not an allowed option")
```
itswisdomagain: This error message could be more informative. I'm thinking: `Invalid selection. Please enter a number between 1 and N`

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
+		}
+		choiceFloor, choiceCeiling := 0, len(accounts)-1
+		if choice < choiceFloor || choice > choiceCeiling {
+			return fmt.Errorf("%d is not an allowed option", choice)
```
itswisdomagain: Same message above would work here as well

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
+		}
+		return nil
 	}
 
-	prompt := promptui.Select{
-		Label: "Select source account",
-		Items: promptItems,
+	accountItems := map[int]uint32{}
```
itswisdomagain: This is a map of account numbers. Wouldn't a more descriptive variable name be `accountNumbers`? When I first saw it, I first thought to my self, what are account items? Do accounts have items? Or is this a map of accountItem objects? Had to defer to your usage of the variable below to determine its purpose

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
+	options := make([]string, len(accounts))
+
+	for idx, v := range accounts {
```
itswisdomagain: Forgive my obsession with variable naming, but while you're here, can you use `for index, account : range accounts`? Helps almost anyone familiar with programming generally instantly have an idea of what you're doing

itswisdomagain: Or `i` instead of `index`/`idx`

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
 	}
-
-	promptItems := []string{}
-	accountItems := map[string]uint32{}
-	for _, v := range accounts {
-		itemStr := fmt.Sprintf("%s (%s)", v.AccountName, dcrutil.Amount(v.Total).String())
-		promptItems = append(promptItems, itemStr)
-		accountItems[itemStr] = v.AccountNumber
+	// validateInput ensures that the input received is a number that corresponds to an account
+	validateInput := func(value string) error {
```
itswisdomagain: We could call this function, `validateAccountSelectionInput` instead

cli/helpers.go
```diff
@@ -5,43 +5,47 @@ import (
...
+		if choice, err = strconv.Atoi(value); err != nil {
```
itswisdomagain: and rename `choice` to `selection`

cli/helpers.go
itswisdomagain: and while at it, we could use `input` instead of `value` as the name of the parameter to the function

cli/helpers.go
```diff
@@ -57,13 +61,10 @@ func getSendDestinationAddress(c *walletrpcclient.Client) (string, error) {
 		return nil
 	}
 
-	prompt := promptui.Prompt{
-		Label:    "Destination Address",
-		Validate: validate,
-	}
-
-	result, err := prompt.Run()
+	var result string
```
itswisdomagain: this declaration is unnecessary

cli/helpers.go
```diff
@@ -82,26 +83,20 @@ func getSendAmount() (int64, error) {
...
 	}
 
 	return amount, nil
 }
 
 func getWalletPassphrase() (string, error) {
-	prompt := promptui.Prompt{
-		Label: "Wallet Passphrase",
-		Mask:  '*',
+	emptyValidator := func(v string) error {
```
itswisdomagain: Instead of having to create an empty validator, why not pass nil to the `terminalprompt.RequestInputSecure` function and in the function's handler, check if the validateFn is nil before attempting to call it

itswisdomagain: Or you could create a public `EmptyValidator` function in `terminalprompt` and whenever you don't need to perform custom validation, you call that function. All the function does could be to check if the input is not null/empty.  But please don't space trim inputs. "  " (double spaces) may be a valid input in some scenarios

cli/terminalprompt/terminalprompt.go
```diff
@@ -0,0 +1,128 @@
...
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"os"
+	"os/signal"
+	"strings"
+	"syscall"
+
+	"golang.org/x/crypto/ssh/terminal"
```
itswisdomagain: Surprised I'm just seeing this. The aim of this PR effort is to remove external dependencies for the use cases we have here. Adding this new dependency invalidates our efforts

itswisdomagain (changes requested):
Good improvements made. Just noticed our objective for this PR hasn't really even been met, so you should give attention to that

barisere:
I've made changes to reflect your suggestions. Please take a look.

cli/helpers.go
```diff
@@ -5,47 +5,51 @@ import (
...
+			return errWrongInput
+		}
+		selection--
+		return nil
 	}
 
-	prompt := promptui.Select{
-		Label: "Select source account",
-		Items: promptItems,
+	accountNumbers := map[int]uint32{}
```
itswisdomagain: `accountNumbers` is redundant and unnecessary. You can always use `accounts[selection].AccountNumber` to get account number after selection

cli/helpers.go
```diff
@@ -5,47 +5,51 @@ import (
...
-	promptItems := []string{}
-	accountItems := map[string]uint32{}
-	for _, v := range accounts {
-		itemStr := fmt.Sprintf("%s (%s)", v.AccountName, dcrutil.Amount(v.Total).String())
-		promptItems = append(promptItems, itemStr)
-		accountItems[itemStr] = v.AccountNumber
+	// validateAccountSelection  ensures that the input received is a number that corresponds to an account
+	validateAccountSelection := func(input string) error {
+		minAllowed, maxAllowed := 1, len(accounts)
+		errWrongInput := fmt.Errorf("Error: input must be between %d and %d", minAllowed, maxAllowed)
```
itswisdomagain, line 28: What if `len(accounts) == 1`? The error message would read: `"Error: input must be between 1 and 1"`. Not exactly grammatically correct.

@raedah suggested proceeding with the default account if there's no other account. In line with that, if `len(accounts) == 1` , this `getSendSourceAccount` function should return `accounts[0]. AccountNumber`

cli/terminalprompt/getinput.go
```diff
@@ -14,42 +14,53 @@ import (
...
 )
 
-// getPassword - Prompt for password.
-func getPassword(prompt string) (string, error) {
+// getTextInput - Prompt for text input.
+func getTextInput(prompt string) (string, error) {
+	fmt.Println(prompt)
+
+	reader := bufio.NewReader(os.Stdin)
+	text, err := reader.ReadString('\n')
```
itswisdomagain, line 22: This can be replaced with the `ReadLine` function of bufio reader:
`inputBytes, _, err := reader.ReadLine()`

Then your return statement can convert the bytes to string without needing to trim:
`return string(inputBytes), nil`

itswisdomagain: I'm thinking that doing this will render the `skipEOFError` function in `terminalprompt.go` obsolete? If so, that function should go. I really don't see it's use case but will appreciate some enlightenment if it's useful

cli/cli.go
```diff
@@ -43,15 +43,22 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
...
 	signalChannel := make(chan os.Signal, 1)
 	signal.Notify(signalChannel, os.Interrupt)
 	go func() {
-		for _ = range signalChannel {
-			fmt.Println("\n^C interrupt, caught on main")
-			//os.Exit(1)
-		}
+		// Capture SIGINT and do nothing. This allows other functions that need to clean up
+		// on SIGINT to do so and exit. If we exit here, the other functions may not have
+		// finished their clean up.
```
itswisdomagain: if we don't exit here and the current process does not capture SIGINT and exits, then the program keeps running even after user tries to terminate

I understand the issue this present. For now, let's take off this SIGINT listener, as it doesn't perform any cleanup work. The implication is that if a user attempts to cancel and there is no SIGINT listener, the program will simply exit. If there is a SIGINT listener like we have in `getPasswordInput`, then that listener has to exit the process manually, which it does currently.

Capturing SIGINT without exiting will only keep the program running if there is no other SIGINT listener that exits the program. To reproduce this error, run `dcrcli send` and ^C before getting to the part where it asks for wallet passphrase

itswisdomagain: While testing this, also remove the SIGINT listener in `getPasswordInput` and confirm if it's even needed. Testing on a Mac terminal works fine without the listener - the terminal's echo setting is automatically restored. If we can confirm this behaviour across windows cmd prompt and linux terminal, we can safely take this out. Until then, having the listener in that function is fine.

cli/terminalprompt/terminalprompt.go
```diff
@@ -0,0 +1,85 @@
...
+
+// ValidatorFunction  validates the input string according to its custom logic.
+type ValidatorFunction func(string) error
+
+// EmptyValidator is a noop validator that can be used if no validation is needed.
+var EmptyValidator = func(v string) error {
+	return nil
+}
+
+func skipEOFError(value string, err error) (string, error) {
```
itswisdomagain, line 16: Reconfirm the necessity of this function. I think the program works fine without it. More so when you use `reader.ReadLine` to get user input as pointed out in an earlier comment

cli/terminalprompt/terminalprompt.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+}
+
+// RequestSelection prompts the user to select from a list of options. The user is expected to enter
+// a number that corresponds to an item in the list.
+// If an error other than EOF occurs while requesting input, the error is returned.
+// It calls `validate` on the received input. If `validate` returns an error, the user is prompted
+// again for a correct input.
+func RequestSelection(message string, options []string, validate ValidatorFunction) (string, error) {
+	var label = message + "\n"
```
itswisdomagain: label may make more sense as `prompt` or `promptMessage`

cli/terminalprompt/getinput.go
```diff
@@ -14,42 +14,53 @@ import (
...
+	fmt.Println()
```
itswisdomagain: Introducing a new line after reading user input may not be best for displaying messages directly related to the user input, such as validation error messages.

Compare the 2 screenshots below, the first has an empty line space after reading input, the second does not. Do you agree that the error message sitting closer to the user input, makes the error message easier to understand? Especially if the user were to get distracted from his terminal and comes back to see the error message.

screenshot 2018-11-17 at 7 09 27 am
screenshot 2018-11-17 at 7 08 37 am

If you agree, then take off `fmt.Println()` and use it only where appropriate such as **after** reading a valid input and before prompting for the next information

cli/terminalprompt/getinput.go
```diff
@@ -14,42 +14,53 @@ import (
...
+		os.Exit(1)
 	}()
 
 	// disable terminal echo
-	termEcho(false)
+	setTerminalEcho(false)
 
 	// Echo is disabled, now grab the data.
 	reader := bufio.NewReader(os.Stdin)
 	text, err := reader.ReadString('\n')
```
itswisdomagain, line 49: This can also be replaced with `ReadLine`

itswisdomagain (changes requested):
Much better, seeing as everything now works with standard libraries. There's a key concern though. Trying to cancel the send command before it gets to the wallet passphrase stage doesn't cancel. The fix to this is mentioned in one of the comments. While at it, give attention to the other matters.

cli/terminalprompt/getinput.go
barisere: > I'm thinking that doing this will render the `skipEOFError` function in `terminalprompt.go` obsolete? If so, that function should go. I really don't see it's use case but will appreciate some enlightenment if it's useful

`skipEOFError` will remain necessary, because the underlying function is `ReadSlice` which returns `EOF` if it encounters it.
`EOF` is returned if the user enters `Ctrl+D` (on Linux). That should be treated equivalent to pressing `Enter` (it means the user is done entering data).

barisere: > This can be replaced with the `ReadLine` function of bufio reader:
> `inputBytes, _, err := reader.ReadLine()`
> 
> Then your return statement can convert the bytes to string without needing to trim:
> `return string(inputBytes), nil`

I'm wary of the second return value from `ReadLine`. If it's set, it means that there was a `bufio.ErrBufferFull` error, and you should retry for the next fragment of the line. If we are expecting long inputs that can fill the input buffer, and would like to continue until the rest of the line is read, then `ReadLine` is okay. But I think that is not the case. In fact, `ReadString` may be just okay here.

cli/terminalprompt/terminalprompt.go
barisere: See my reply here: https://github.com/raedahgroup/dcrcli/pull/36#discussion_r234402626

cli/terminalprompt/getinput.go
barisere: See https://github.com/raedahgroup/dcrcli/pull/36#discussion_r234402626

cli/cli.go
barisere: On Linux my terminal setting was not restored when I removed the SIGINT listener.

cli/terminalprompt/getinput.go
itswisdomagain: That's fine then

cli/cli.go
itswisdomagain: I figured!

itswisdomagain (approved):

PR #39, opened 2018-11-16
Added interactive receive prompt

### LINKED ISSUE #38: make cli receive command interactive if account number is not provided
Currently, `dcrcli receive` command requires account number argument and returns error message if no account number argument is supplied.

Seeing as users should not really be expected to memorize the account numbers for all the accounts in their wallet, it might help to give users a prompt to select the account they want to receive funds for. This functionality is already implemented in the web server package and similar functionality exists for the `dcrcli send` command as described in #10 

The `dcrcli receive` command should retain its ability to accept an account number argument though and should only prompt user for selection if no argument is provided or an invalid argument was provided

Jujhar:
Passing no parameters to `./dcrcli receive` brings up interactive console to select account to generate receive address for previously displayed error.

Fixes #38


Jujhar:
Please hold off on merging for a bit until I update and follow up.
Raedah wants third party plugin Promptui removed from system.

cli/cli.go
```diff
@@ -161,11 +161,26 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
+		sourceAccount, err := getSendSourceAccount(c.walletrpcclient)
+		if err != nil {
+			return nil, err
+		}
+
+		recieveAddress = fmt.Sprint(sourceAccount)
+	}
+
+	// if an address was passed in eg. ./dcrcli receive 0 use that address
+	if recieveAddress == "" {
```
itswisdomagain: Shouldn't this be an else? `if len(commandArgs) == 0`, prompt user to select an account, **`else`**, use the value present in `commandArgs[0]`. Pretty straight forward.

cli/cli.go
```diff
@@ -161,11 +161,26 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
 	if len(commandArgs) == 0 {
-		return nil, errors.New(" account number is required.  \nUsage:\n  receive \"accountnumber\"")
+
+		// display menu options to select account
+		sourceAccount, err := getSendSourceAccount(c.walletrpcclient)
+		if err != nil {
+			return nil, err
+		}
+
+		recieveAddress = fmt.Sprint(sourceAccount)
```
itswisdomagain: This is unnecessary, what we actually need for receiveAddress is a `uint32` which is what we already get from the `getSendSourceAccount` function.

You should instead make a direct assignment of the result of the `getSendSourceAccount` function to `receiveAddress`

Then there'd be no need for the string parsing in line 183. Instead, you should parse the string in the else block where you read the argument passed in; and assign the result of the parsing to `receiveAddress`

itswisdomagain (changes requested):
This looks good, just some improvements required.

itswisdomagain:
After taking another look and conferring with @raedah, you do not need to worry about removing the promptui dependency. That's already done in a separate PR that will get merged soon. After that PR is merged and promptui removed, your code will still work fine, so you can safely proceed to just making the corrections above for now.

Jujhar:
Ready to go I think.

cli/cli.go
```diff
@@ -161,16 +161,30 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
-	acc, err := strconv.ParseUint(commandArgs[0], 0, 32)
-	if err != nil {
-		return nil, fmt.Errorf("Error parsing account number: %s", err.Error())
+		// display menu options to select account
+		sourceAccount, err := getSendSourceAccount(c.walletrpcclient)
+		if err != nil {
+			return nil, err
+		}
+
+		recieveAddress = sourceAccount
```
itswisdomagain: I think you can straight-forward assign `recieveAddress` in line 170 instead of using a new "temporary" variable `sourceAccount`

cli/cli.go
```diff
@@ -161,16 +161,30 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
+		x, err := strconv.ParseUint(commandArgs[0], 0, 32)
+		if err != nil {
+			return nil, fmt.Errorf("Error parsing account number: %s", err.Error())
+		}
+
+		recieveAddress = uint32(x)
 	}
 
-	r, err := c.walletrpcclient.Receive(uint32(acc))
+	r, err := c.walletrpcclient.Receive(uint32(recieveAddress))
```
itswisdomagain: This conversion to uint32 is unnecessary. `recieveAddress` was declared in line 164 above as unit32 already.

itswisdomagain (changes requested):
Just these little code cleanup.

cli/cli.go
```diff
@@ -161,16 +161,30 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
+		sourceAccount, err := getSendSourceAccount(c.walletrpcclient)
+		if err != nil {
+			return nil, err
+		}
+
+		recieveAddress = sourceAccount
+	} else {
+
+		// if an address was passed in eg. ./dcrcli receive 0 use that address
+		x, err := strconv.ParseUint(commandArgs[0], 0, 32)
```
itswisdomagain: Shouldn't you use base 10 for this conversion? Like:
`x, err := strconv.ParseUint(commandArgs[0], 10, 32)`

Jujhar:
Updated and tested.

cli/cli.go
```diff
@@ -161,16 +161,30 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
 	if len(commandArgs) == 0 {
-		return nil, errors.New(" account number is required.  \nUsage:\n  receive \"accountnumber\"")
-	}
 
-	acc, err := strconv.ParseUint(commandArgs[0], 0, 32)
-	if err != nil {
-		return nil, fmt.Errorf("Error parsing account number: %s", err.Error())
+		// display menu options to select account
+		var err error
+		recieveAddress, err = getSendSourceAccount(c.walletrpcclient)
```
itswisdomagain, line 168: If you do `recieveAddress, err := getSendSourceAccount(c.walletrpcclient)`, there'd be no need for the declaration of the err variable in the preceding line.

cli/cli.go
```diff
@@ -161,16 +161,30 @@ func (c *CLI) listCommands(commandArgs []string) (*response, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+	} else {
+
```
itswisdomagain: Maybe take off the whitespaces in lines 175 and 177?

itswisdomagain (changes requested):
This looks good @Jujhar! Please rebase your branch with master and fix merge conflicts. Or it might be easier/faster to update your local master branch by merging with this remote master branch, creating a new branch off the updated master and then re-introducing your changes.

Either ways, while doing that, do you mind giving some attention to these minor issues?

cli/cli.go
Jujhar: That would make recieveAddress a new variable and we want to use the one located a step up the scope (line 164)

Jujhar:
Updated and tested.
Interactive prompt is removed but selection (if more than one account is present) is possible through inputting a number.

itswisdomagain (approved):

PR #40, opened 2018-11-18
[WIP] Send utxo

yaowork:
This creates functions that sends UTXOs to an address.

### DIFF
--- a/walletrpcclient/walletrpcclient.go
+++ b/walletrpcclient/walletrpcclient.go
@@ -60,33 +60,8 @@ func (c *Client) connect(address, cert string, noTLS bool) (*grpc.ClientConn, er
 	return conn, nil
 }
 
-func (c *Client) Send(amount int64, sourceAccount uint32, destinationAddress, passphrase string) (*SendResult, error) {
-	// decode destination address
-	addr, err := dcrutil.DecodeAddress(destinationAddress)
-	if err != nil {
-		return nil, fmt.Errorf("error decoding destination address: %s", err.Error())
-	}
-
-	// get script
-	pkScript, err := txscript.PayToAddrScript(addr)
-	if err != nil {
-		return nil, err
-	}
-
-	// construct transaction
-	constructRequest := &pb.ConstructTransactionRequest{
-		SourceAccount: sourceAccount,
-		NonChangeOutputs: []*pb.ConstructTransactionRequest_Output{{
-			Destination: &pb.ConstructTransactionRequest_OutputDestination{
-				Script:        pkScript,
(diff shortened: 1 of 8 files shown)


walletrpcclient/walletrpcclient.go
```diff
@@ -121,6 +97,74 @@ func (c *Client) Send(amount int64, sourceAccount uint32, destinationAddress, pa
...
+				Address:      k.Address,
+				OutputScript: k.OutputScript,
+			}
+			outputs = append(outputs, output)
+		}
+	}
+
+	constructRequest := &pb.ConstructTransactionRequest{
+		SourceAccount:    sourceAccount,
+		NonChangeOutputs: make([]*pb.ConstructTransactionRequest_Output, len(outputs)),
```
itswisdomagain, line 151: This is what is holding this PR. According to the docs for [dcrwallet ConstructTransaction](https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md#constructtransaction), the `NonChangeOutputs` property in the `ConstructTransactionRequest` object is to describe the **destination** of the funds, not the source.

My next comment below will expatiate.

walletrpcclient/walletrpcclient.go
```diff
@@ -121,6 +97,74 @@ func (c *Client) Send(amount int64, sourceAccount uint32, destinationAddress, pa
...
+	}
+
+	for i, v := range outputs {
+		constructRequest.NonChangeOutputs[i] = &pb.ConstructTransactionRequest_Output{
+			Destination: &pb.ConstructTransactionRequest_OutputDestination{
+				Address:       destinationAddress,
+				Script:        v.OutputScript,
+				ScriptVersion: 0,
```
itswisdomagain, line 159: When adding an `OutputDestination` to `ConstructTransactionRequest.NonChangeOutputs`, the doc says to use **either** Address **or** Script, but not both.

Regardless of the one you use, the `NonChangeOutputs` is only meant to provide information about the destination/recipient of the funds, not the source.

Even more so, when you provide an Address in the `OutputDestination` object, the script is ignored. At least that's[ what the doc says](https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md#constructtransaction).

itswisdomagain (changes requested):
The part of this PR that attempts to send funds using the selected UTXOs is incorrectly implemented. I'd suggest to clean up this PR to focus on extracting the UTXOs for selection, while a different PR handles the implementation for the actual transaction construction.

Alternative would be to fix the implementation or close the PR.

PR #42, opened 2018-11-25
Add utilities to fetch transactions

yaowork:
This adds functions that fetches transactions of a wallet. 

The following are yet to be implemented:

1. CLI interface 
2. Pagination on the web ui

### DIFF
--- a/walletrpcclient/helpers.go
+++ b/walletrpcclient/helpers.go
@@ -1,0 +1,117 @@
+package walletrpcclient
+
+import (
+	"github.com/decred/dcrd/chaincfg/chainhash"
+	pb "github.com/decred/dcrwallet/rpc/walletrpc"
+)
+
+func AtomsToCoin(amount int64) float64 {
+	return float64(amount) / float64(100000000)
+}
+
+func getOutputs(credits []*pb.TransactionDetails_Output) []*TransactionOutput {
+	outputs := make([]*TransactionOutput, len(credits))
+
+	for i, v := range credits {
+		output := &TransactionOutput{
+			Index:        v.Index,
+			Account:      v.Account,
+			Internal:     v.Internal,
+			Amount:       AtomsToCoin(v.Amount),
+			Address:      v.Address,
+			OutputScript: v.OutputScript,
+		}
+		outputs[i] = output
+	}
+
+	return outputs
+}
+
+func getInputs(debits []*pb.TransactionDetails_Input) []*TransactionInput {
+	inputs := make([]*TransactionInput, len(debits))
+
+	for i, v := range debits {
+		input := &TransactionInput{
+			Index:           v.Index,
+			PreviousAccount: v.PreviousAccount,
+			PreviousAmount:  AtomsToCoin(v.PreviousAmount),
+		}
+		inputs[i] = input
+	}
+
+	return inputs
+}
+
+func getTransactionDetails(txDetails *pb.TransactionDetails) *TransactionDetails {
+	hash, _ := chainhash.NewHash(txDetails.Hash)
+
+	tx := &TransactionDetails{
+		Hash:            hash.String(),
+		Transaction:     txDetails.Transaction,
+		Debits:          getInputs(txDetails.Debits),
+		Credits:         getOutputs(txDetails.Credits),
+		Fee:             txDetails.Fee,
+		Timestamp:       txDetails.Timestamp,
+		TransactionType: int(txDetails.TransactionType),
+	}
+
+	return tx
+}
+
+func getTransactionsDetails(txDetails []*pb.TransactionDetails) []*TransactionDetails {
+	txns := make([]*TransactionDetails, len(txDetails))
+
+	for i, v := range txDetails {
+		txns[i] = getTransactionDetails(v)
+	}
+
+	return txns
+}
+
+func getBlockDetails(blockDetails *pb.BlockDetails) *BlockDetails {
+	hash, _ := chainhash.NewHash(blockDetails.Hash)
+
+	b := &BlockDetails{
+		Hash:           hash.String(),
+		Height:         blockDetails.Height,
+		Timestamp:      blockDetails.Timestamp,
+		ApprovesParent: blockDetails.ApprovesParent,
+		Transactions:   getTransactionsDetails(blockDetails.Transactions),
+	}
+
+	return b
+}
+
+func getSummary(blockDetails *BlockDetails) []*TransactionSummary {
+	summaries := []*TransactionSummary{}
+
+	for _, v := range blockDetails.Transactions {
+		for _, j := range v.Credits {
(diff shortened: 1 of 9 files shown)


yaowork:
ready for merge

raedah:
Pagination is not part of any current spec. Do not over engineer.

walletrpcclient/walletrpcclient.go
```diff
@@ -208,3 +211,61 @@ func (c *Client) NextAccount(accountName string, passphrase string) (uint32, err
 
 	return r.AccountNumber, nil
 }
+
+func (c *Client) FetchHeaders() (*FetchHeadersResult, error) {
```
itswisdomagain: This function is never called. Why do we have it? If it doesn't relate to this PR (fetching transactions), let's do without it. If it does, please explain it's relevance.

If taking off this function, the following will have to be removed as well, as they'd become obsolete:
- `FetchHeadersResult` struct in line 23 of `walletrpcclient/types.go`
-  `c.walletLoaderServiceClient` defined in line 18 and assigned in line 30 of `walletrpcclient/walletrpcclient.go`

walletrpcclient/types.go
```diff
@@ -19,3 +19,60 @@ type AccountBalanceResult struct {
 type SendResult struct {
 	TransactionHash string `json:"transaction_hash"`
 }
+
+type FetchHeadersResult struct {
```
itswisdomagain: This might be unnecessary. Refer to my comment on line 215 of `walletrpcclient/walletrpcclient.go`

walletrpcclient/walletrpcclient.go
```diff
@@ -13,7 +14,8 @@ import (
 )
 
 type Client struct {
-	walletServiceClient pb.WalletServiceClient
+	walletServiceClient       pb.WalletServiceClient
+	walletLoaderServiceClient pb.WalletLoaderServiceClient
```
itswisdomagain: This might be unnecessary. Refer to my comment on line 215 of `walletrpcclient/walletrpcclient.go`

walletrpcclient/walletrpcclient.go
```diff
@@ -25,6 +27,7 @@ func New(address, cert string, noTLS bool) (*Client, error) {
 
 	// register clients
 	c.walletServiceClient = pb.NewWalletServiceClient(conn)
+	c.walletLoaderServiceClient = pb.NewWalletLoaderServiceClient(conn)
```
itswisdomagain: This might be unnecessary. Refer to my comment on line 215 of `walletrpcclient/walletrpcclient.go`

walletrpcclient/helpers.go
```diff
@@ -0,0 +1,117 @@
...
+		Height:         blockDetails.Height,
+		Timestamp:      blockDetails.Timestamp,
+		ApprovesParent: blockDetails.ApprovesParent,
+		Transactions:   getTransactionsDetails(blockDetails.Transactions),
+	}
+
+	return b
+}
+
+func getSummary(blockDetails *BlockDetails) []*TransactionSummary {
```
itswisdomagain: This is unnecessary and problematic. What you're doing here is digging deep into each transaction to extract the inputs and outputs.

Then in the web and cli interface, this extracted info is what you display. You're displaying the individual inputs and outputs across all the transactions instead of the transactions themselves.

I'm confident we don't need this. Even if we need to display the individual inputs and outputs in a transaction, it shouldn't be on the transactions page. More so, the decred block explorer site provides this info already so instead of adding another feature here to show individual inputs and outputs, we could just link to the transaction detail page of dcrd explorer (an example: https://testnet.dcrdata.org/tx/3cb931653122a5cbd4ca0874b676ecbd6be3d5178ef587af87a1dc49c044c380)

If you agree that focus should be on **transactions** not **inputs/outputs**, then `MinedTransactions` and `UnminedTransactions` fields in `GetTransactionsResult` would be sufficient. `MinedTransactions` is of type `BlockDetails` which has `Transactions` field, which is an array of `TransactionDetails`. `UnminedTransactions` is itself an array of `TransactionDetails`.

itswisdomagain (changes requested):
You have the core concept, but work isn't (well) done yet. Focus is on displaying individual **transactions**, not **inputs/outputs**.

What information should be displayed for each transaction? The hash, transaction date, amount (sum of all output amounts), maybe from and to addresses, fee and any other potentially relevant info. Compare the Transactions section of the decred block explorer (https://testnet.dcrdata.org/block/00000000bbfafc943e01935d0bb9ae2bf2df204b0c12a8ba9a4cab908f2b6e7c?ind=3) for a clue.

And remember to keep it simple. We do not need plenty info, no need to attempt to extract every possible information. For example, I don't think we need the **details** of the inputs and outputs for each transaction, so the `TransactionSummary`, `TransactionInput` and `TransactionOutput` structs could be removed. We can instead display **number** of inputs and outputs.

itswisdomagain:
This is what we currently have for the transactions page on the web interface:
screen shot 2561-12-04 at 8 49 54 am

Similar to what we have on cli:
screen shot 2561-12-04 at 8 51 49 am

itswisdomagain:
We're making progress! Please remember to include screenshots when you update subsequently to give a visual appraisal of what was done. Here's what I capture:

screen shot 2561-12-08 at 5 50 10 pm

The progress I observe is that the table now displays `transaction hash`, `total amount` and `transaction date`

itswisdomagain:
1 immediate observation is that the hash displayed are not transaction hashes but block hashes. With the implication that the amount displayed is the total amount mined in the block which usually will contain a number of other transactions not related to the querying wallet. Same thing with the dates.

The [dcrwallet rpc doc](https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md#gettransactions) explains that the `getTransactions` call returns `mined_transactions` and `unmined_transactions`

`mined_transactions` is a `BlockDetails` object, that's not what we're looking for exactly. We're looking for `TransactionDetails` objects, which `BlockDetails` contains in `BlockDetails.transactions`.
`unmined_transactions` is itself an array of `TransactionDetails` objects, so yes we need it.

We should now have access to 2 arrays of `TransactionDetails` objects; in `mined_transactions.transactions` and `unmined_transactions`.

If you loop through both arrays, you can get the `hash`, `timestamp` directly from each `TransactionDetails` object. The `TransactionDetails` object also has `outputs` property. If you loop through this property, you can get the total amount in involved in the transaction. There we have what we need, `hash`, `date` and `amount` for each **transaction** belonging to the wallet.

Remember to loop through both `mined_transactions.transactions` and `unmined_transactions`

itswisdomagain:
Another thing I notice is that each `Output` object in `TransactionDetails.credits` has a boolean property `internal` that informs whether that particular output returned to the wallet. If this property is true, the amount in that output should not be included in the total transaction amount.

PR #45, opened 2018-12-04
Cleanup usage message

### LINKED ISSUE #44: Simplify dcrcli command line usage
### Display available commands if the program is run without a command argument.

Currently the user is shown the following message if the `dcrcli` program is run without any command argument.

```
$ ./dcrcli
usage: dcrcli [OPTIONS] <command> [<args...>]

available dcrcli commands:

For available options, see 'dcrcli -h'
```

Instead of asking the user to type `dcrcli -h` to get a list of commands, the available commands should be displayed immediately as follows.
```
$ ./dcrcli
usage: dcrcli [OPTIONS] <command> [<args...>]

available dcrcli commands:
     Command | Description 
             |   
     receive | Generate address to receive funds 
        send | Send DCR to address. Multi-step 
     balance | Check balance of an account 
 nextaccount | Generate next account for wallet 
          -l | List all supported commands 
```

### Replace the `--mode` flag with `--http`

The default mode for the program is CLI. To run in HTTP mode, the `--http` flag can be used. This is simpler and more concise. Take the `godoc` program.

`$ ./dcrcli --mode=http balance` becomes `$ ./dcrcli --http balance`
`$ ./dcrcli --mode=cli balance` becomes `$ ./dcrcli balance`

### Display available commands in alphabetical order

Running `$ ./dcrcli -l` (to list available commands) displays the commands in random order. This is because the commands are displayed by ranging over a map: Go maps are not ordered.
The commands should be listed in the same order each time. Below are outputs from three runs of the program.

```
$ ./dcrcli 
usage: dcrcli [OPTIONS] <command> [<args...>]

available dcrcli commands:
     Command | Description 
             |   
     receive | Generate address to receive funds 
        send | Send DCR to address. Multi-step 
     balance | Check balance of an account 
 nextaccount | Generate next account for wallet 
          -l | List all supported commands

$ ./dcrcli 
usage: dcrcli [OPTIONS] <command> [<args...>]

available dcrcli commands:
     Command | Description 
             |   
          -l | List all supported commands 
     receive | Generate address to receive funds 
        send | Send DCR to address. Multi-step 
     balance | Check balance of an account 
 nextaccount | Generate next account for wallet

$ ./dcrcli 
usage: dcrcli [OPTIONS] <command> [<args...>]

available dcrcli commands:
     Command | Description 
             |   
     balance | Check balance of an account 
 nextaccount | Generate next account for wallet 
          -l | List all supported commands 
     receive | Generate address to receive funds 
        send | Send DCR to address. Multi-step
```

barisere:
Fixes #44 
Makes breaking changes to the configuration file format.


config.go
```diff
@@ -34,9 +34,8 @@ type config struct {
 	RPCPassword       string `short:"P" long:"rpcpass" default-mask:"-" description:"RPC password"`
 	WalletRPCServer   string `short:"w" long:"walletrpcserver" description:"Wallet RPC server to connect to"`
 	RPCCert           string `short:"c" long:"rpccert" description:"RPC server certificate chain for validation"`
-	HTTPServerAddress string `short:"p" long:"httpserveraddress" description:"Http address to serve if mode is set to http"`
+	HTTPServerAddress string `long:"http" description:"HTTP address to serve. This turns HTTP mode on."`
```
itswisdomagain: After discussing these configuration changes with @raedah, here are some thoughts on a more user-friendly approach:

- separate server address config variable from http mode variable so users can run dcrcli on http mode without having to supply an address argument to the `dcrcli --http` command each time
- as a follow up to the above, we should retain `HttpServerAddress` to hold the http server config value and prolly use `s` for short and `server` for long
- as another follow up to the first thought, once we have a different config variable providing server address information for http mode, we can allow users toggle between different modes by using `boolean` flags for each mode viz `--cli` for cli, `--http` for http server, and introduce more such flags as other interfaces are added
- alternatively, we could (should we?) retain the mode flag to serve the purpose above (whichever approach is more user-friendly)
- if any mode flag is supplied it should take precedence, if no mode flag is provided, use cli by default even if server address is provided or set in config file
- the reason for the previous is this: Assume the following 2 scenarios for a configuration with server address set but mode not supplied
- scenario a: we run the program using web interface, the only way a user can "switch" to cli is by appending the `--cli` flag. To use the program in cli mode, every command has to be typed into the terminal for execution. That means user has to include the `--cli` flag for each command he wants to execute
- scenario b: we run the program using cli interface, a user can easily "switch" to web by appending `--http` flag (having already set the address) and this only needs to be done once since once the server is started, it stays active till it's shut down

After making these adjustments, update `sample-dcrcli.conf` to reflect the new variables that can be configured

itswisdomagain (changes requested):

config.go
barisere: > as a follow up to the above, we should retain HttpServerAddress to hold the http server config value and prolly use s for short and server for long

I think `httpserveraddress` (the current key) is more descriptive. Maybe we can shorten it to `httpserver`. With this we can remove the need for a short form of the flag (the short form `-s` is less intuitive. In some CLI programs I am familiar with, `-s` can mean "silent mode". See?)

> as another follow up to the first thought, once we have a different config variable providing server address information for http mode, we can allow users toggle between different modes by using boolean flags for each mode viz --cli for cli, --http for http server, and introduce more such flags as other interfaces are added

Yes, this is better. I'll change `--http` to be a boolean flag.

> alternatively, we could (should we?) retain the mode flag to serve the purpose above (whichever approach is more user-friendly)

Keeping the `mode` flag is less user-friendly: it creates dependent options that must be supplied together to achieve an effect that can be achieved with only one flag.

barisere: @itswisdomagain See my new commits.

config.go
```diff
@@ -34,9 +34,9 @@ type config struct {
 	RPCPassword       string `short:"P" long:"rpcpass" default-mask:"-" description:"RPC password"`
 	WalletRPCServer   string `short:"w" long:"walletrpcserver" description:"Wallet RPC server to connect to"`
 	RPCCert           string `short:"c" long:"rpccert" description:"RPC server certificate chain for validation"`
-	HTTPServerAddress string `short:"p" long:"httpserveraddress" description:"Http address to serve if mode is set to http"`
+	HTTPServerAddress string `long:"httpserveraddress" description:"HTTP address to serve. This turns HTTP mode on."`
```
itswisdomagain: Setting `httpserveraddress` doesn't turn http mode on. The value is only used if the `--http` flag is suppliedd or set to true/1 in config

cli/cli.go
```diff
@@ -141,20 +145,20 @@ func printResult(res *response) {
 	w.Flush()
 }
 
-func (c *CLI) listCommands(commandArgs []string) (*response, error) {
+func (c *CLI) listCommands(commandArgs []string) *response {
```
itswisdomagain: the `commandArgs` function parameter is no longer necessary... It was only added to maintain the command `Handler` signature

cli/cli.go
```diff
@@ -77,12 +82,12 @@ func (c *CLI) RunCommand(commandArgs []string) {
 func (c *CLI) noCommandReceived() {
 	fmt.Printf("usage: %s [OPTIONS] <command> [<args...>]\n\n", c.appName)
 	fmt.Printf("available %s commands:\n", c.appName)
-	c.listCommands(nil)
-	fmt.Printf("\nFor available options, see '%s -h'\n", c.appName)
+	printResult(c.listCommands(nil))
```
itswisdomagain: Similar to the behaviour of the `noCommandReceived` and `invalidCommandReceived` functions, make the `listCommands` function also print its result to the terminal and return nothing.

That way, simply calling `c.listCommands()` will display the commands on the terminal. This simpler method call will prove useful when you consider it's usage else where (in the RunCommand function explained in a separate comment)

cli/cli.go
```diff
@@ -141,20 +145,20 @@ func printResult(res *response) {
...
-			v,
-			c.descriptions[i],
+			c.commands[key],
+			c.descriptions[key],
 		}
 
 		res.result = append(res.result, item)
 	}
-	return res, nil
+	return res
```
itswisdomagain: call `printResult(res)` here and don't return anything

itswisdomagain (changes requested):
Currently, `dcrcli -l` and `dcrcli listcommands` return `invalid command` message. In inspecting the cause, I found the following:

- The flag `-l` is already accounted for in dcrcli options, see [line 31](https://github.com/raedahgroup/dcrcli/blob/master/config.go#L31) and [134](https://github.com/raedahgroup/dcrcli/blob/master/config.go#L134) of `config.go`
- The only thing we need to account for in the cli package, is to watch out for the `listcommands` command and execute `c.listCommands()`. The following change to lines 63 to 68 of `cli/cli.go` does the trick:
```
if command == "listcommands" {
	c.listCommands()
	os.Exit(1)
} else if !c.IsCommandSupported(command) {
	c.invalidCommandReceived(command)
	os.Exit(1)
}
```

While making these adjustments, see if you can reorder the function definitions in `cli/cli/.go`. The following order makes sense to me (and makes it easier/faster to find implementations of different functions):
- `New` calls `registerHandlers` which in turn calls `RegisterHandler`, they should probably appear in that order. I'm also thinking we do not need to expose (make public) the `RegisterHandler` function.  What do you think?
- Then we can have `RunCommand`, which calls `noCommandReceived`, `IsCommandSupported`, `invalidCommandReceived` and `listCommands`. If they could appear in that order following `RunCommand`, that makes sense too
- `printResult` is usually called at the end of every cli command, so it might make sense to keep at the bottom
- Then it would make sense too to have the different function handlers arranged in the order they are registered

itswisdomagain:
One more thing, edit `sample-dcrcli.conf` to reflect changes to the config.go file.
- Remove `; mode=cli` from the sample conf file
- Add `; http=1` to the sample conf file

Also update the comment in line 24 to `; HTTP server address to serve dcrcli via http, works with http=1 or http=true`

config.go
barisere: > Currently, dcrcli -l and dcrcli listcommands return invalid command message.

listcommands is to be removed in this PR. I missed that. I'll remove it.

itswisdomagain:
One last thing, currently, there's no way to view supported cli commands except if no command is passed or an invalid command is passed.

Can we add that to the output of `dcrcli -h`

cli/cli.go
```diff
@@ -17,16 +18,29 @@ type (
...
+		descriptions     map[string]string
+		commandListOrder []string
+		appName          string
+		walletrpcclient  *walletrpcclient.Client
+	}
+
+	// Command is an action that can be requested at the CLI
+	Command struct {
+		Name        string
+		Description string
```
itswisdomagain: Does `Name` and `Description` need to be public?

cli/cli.go
```diff
@@ -204,107 +205,123 @@ func (c *CLI) receive(commandArgs []string) (*response, error) {
...
-			"Account Number",
-		},
-		result: [][]interface{}{
-			[]interface{}{
-				accountName,
-				accountNumber,
-			},
-		},
-	}
+	printResult(stderrHelpWriter, res)
```
itswisdomagain: Why are we printing successful command execution output to stderr?

cli/cli.go
```diff
@@ -42,122 +56,109 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
...
-
-	handler := c.funcMap[command]
-	res, err := handler(commandArgs[1:])
-	if err != nil {
-		fmt.Fprintf(os.Stderr, "Error executing command '%s'\n", command)
-		fmt.Fprintln(os.Stderr, err.Error())
-		os.Exit(1)
+// commands provides the commands available from the cli
+func commands() []command {
+	cli := &cli{}
```
itswisdomagain: Instead of creating a new `cli` object here, why not make the `commands` function a method on the `cli` type like so:
`func (c *cli) commands() ...`

cli/cli.go
```diff
@@ -204,107 +205,123 @@ func (c *CLI) receive(commandArgs []string) (*response, error) {
 	return res, nil
 }
 
-func (c *CLI) send(commandArgs []string) (*response, error) {
-	sourceAccount, err := getSendSourceAccount(c.walletrpcclient)
-	if err != nil {
-		return nil, err
-	}
+var (
```
itswisdomagain: I'm thinking we can move these declarations and the 3 following helper functions to a different file in the same package (maybe helpers.go, keep this file focused on cli commands...

itswisdomagain (changes requested):
Some thoughts on the recent changes to `cli.go`

cli/cli.go
itswisdomagain: I'm thinking if we call this function supportedCommands, it'll make more sense seeing as this only lists currently supported commands and could/would take in more as time goes.

config.go
```diff
@@ -91,30 +92,37 @@ func cleanAndExpandPath(path string) string {
...
+	}
+}
+
 func loadConfig() (*config, []string, error) {
 	cfg := config{
 		ConfigFile: defaultConfigFile,
 		RPCCert:    defaultRPCCertFile,
 	}
+	stdinReadMessage := "The special parameter `-` indicates that a parameter should be read " +
+		"from the\nnext unread line from standard input."
```
itswisdomagain: been thinking... do we _really_ need to display this message? I've been seeing it on `dcrcli -h` but frankly have no idea why it's there...

config.go
```diff
@@ -91,30 +92,37 @@ func cleanAndExpandPath(path string) string {
 	return filepath.Join(homeDir, path)
 }
 
+var helpString = cli.UsageString()
+
+func addParserSettings(parser *flags.Parser) {
+	parser.Usage = helpString
```
itswisdomagain: `helpString` is only used here, why not limit the variable scope to this function only and even better, assign the result of `cli.UsageString()` to `parser.usage` directly

itswisdomagain (changes requested):
some thoughts on `config.go`

cli/cli.go
```diff
@@ -204,107 +205,123 @@ func (c *CLI) receive(commandArgs []string) (*response, error) {
...
-				accountNumber,
-			},
-		},
-	}
+	printResult(stderrHelpWriter, res)
+	os.Exit(0)
+}
 
-	return res, nil
+func (c *cli) noCommandReceived() {
```
itswisdomagain: `c *cli` is not used and can be removed

cli/cli.go
```diff
@@ -204,107 +205,123 @@ func (c *CLI) receive(commandArgs []string) (*response, error) {
...
-	}
+	usagePrefix = "Usage:\n  dcrcli "
+)
 
-	result, err := c.walletrpcclient.Send(sendAmount, sourceAccount, destinationAddress, passphrase)
-	if err != nil {
-		return nil, err
-	}
+// UsageString returns the cli usage message as a string.
+func UsageString() string {
```
itswisdomagain: This function is only used in `config.go`, the whole code in this section can be much more simplified. I've noticed with this update, the cli program displays available commands and options together in one message, which is used as help message and also displayed when no command is passed, invalid command is passed and invalid option is passed.

Can we simplify everything to just 2 functions (1 produces the message described above as string) and the other prints same message to the standard output?

itswisdomagain (changes requested):
Further thoughts on the message displayed when no command is passed, invalid command is passed or invalid option is passed.

Talking about simplifying the help message codes, here's a thought, a `help.go` file in the `cli` package of the following nature. Notice the change to variable and method names too

```
package cli

import (
	"bytes"
	"io"
	"os"
	"text/tabwriter"
)

// HelpMessage returns the cli usage message as a string.
func HelpMessage() string {
	buf := bytes.NewBuffer([]byte{})
	writer := tabWriter(buf)
	printHelpMessage("", writer)
	return buf.String()
}

// PrintHelp outputs the help message to os.Stderr
func PrintHelp() {
	stderrTabWriter := tabWriter(os.Stderr)
	printHelpMessage("Usage:\n  dcrcli ", stderrTabWriter)
}

func tabWriter(w io.Writer) *tabwriter.Writer {
	return tabwriter.NewWriter(w, 0, 0, 1, ' ', tabwriter.AlignRight|tabwriter.Debug)
}

func printHelpMessage(prefix string, w *tabwriter.Writer) {
	res := &response{
		columns: []string{prefix + "[OPTIONS] <command> [<args...>]\n\nAvailable commands:"},
	}
	commands := Commands()

	for _, command := range commands {
		item := []interface{}{
			command.Name,
			command.Description,
		}
		res.result = append(res.result, item)
	}

	printResult(w, res)
}
```

Look simpler? The important functions are maintained `HelpMessage()` used in `config.go` and `PrintHelp()` used in both `cli.go` and `config.go`

This has the added advantage of doing away with all the previously declared variables: `stderrHelpWriter`, `PrintHelp` wrapper function, `HelpMessageRecorder` that was only used in `UsageString` which was only used in `config.go` and `usagePrefix` which is later trimmed out in `UsageString` when providing the help message as string.

`stderrHelpWriter` was also used in `RunCommand` to output the result of cli commands execution. Here's the fix:
```
stdoutTabWriter := tabWriter(os.Stdout)
printResult(stdoutTabWriter, res)
```

May need some tweaking to get it to work as expected but the concept is there.

cli/cli.go
barisere: No. I've made them private

config.go
barisere: It's for piping commands. I don't think we need it now. All the commands we have now are interactive, so piping won't work. I'll remove it.

cli/cli.go
barisere: The reason I have two functions for this is that the library (`go-flags`) thinks it's cool to print part of the help message for me and have me fill the rest: it prints `Usage:\n  ` before printing whatever help message I supply. Worse, it does that only when responding to `--help`. If I pass just the part after the `Usage:\n ` prefix, I get inconsistent help messages.
When printing the help message myself, I have to add the prefix. But when supplying it as a usage string to the library, I have to strip the prefix.
Comparing the `writeHelpMessage` function to the `UsageString` function should make this clearer.

cli/cli.go
barisere: The entire method is now removed.

cli/cli.go
barisere: Ouch! Fixed.

cli/cli.go
barisere: Thanks, that made things simpler.

cli/cli.go
```diff
@@ -42,269 +39,61 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
...
 
+func (c *cli) registerHandlers() {
+	commands := supportedCommands()
+	for _, command := range commands {
+		c.registerHandler(command.name, command.name, command.description, command.handler)
+	}
+}
+
+// registerHandler registers a command, its description and its handler
+func (c *cli) registerHandler(key, command, description string, h handler) {
```
itswisdomagain: the `description` parameter is not used and can be removed...

cli/cli.go
```diff
@@ -42,269 +39,61 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
...
+	if _, ok := c.funcMap[key]; ok {
+		panic("trying to register a handler twice: " + key)
+	}
+
+	c.funcMap[key] = h
+	c.commandListOrder = append(c.commandListOrder, key)
```
itswisdomagain: this `commandListOrder` slice is assigned to here but is never read from or used anywhere else, the code works fine without it, even maintaining the order of commands when displaying available commands on the terminal

cli/cli.go
itswisdomagain: `command` is also not used and can be removed

cli/cli.go
```diff
@@ -42,269 +39,61 @@ func New(walletrpcclient *walletrpcclient.Client, appName string) *CLI {
...
-			v.LockedByTickets,
-			v.VotingAuthority,
-			v.Unconfirmed,
-		}
-	}
-
-	return res, nil
+func (c *cli) invalidCommandReceived(command string) {
+	fmt.Fprintf(os.Stderr, "%s: '%s' is not a supported command.\n\n", c.appName, command)
+	PrintHelp()
```
itswisdomagain: I'm suggesting adding a `appName` string parameter to the `PrintHelp()` function, instead of using the hardcoded `dcrcli` value in the `usagePrefix` variable in that function. More about this is explained in that function (line 19 of help.go specifically)

`PrintHelp()` is called twice in this file, and both instances can be replaced with `PrintHelp(c.appName)` to **ensure** uniformity

cli/help.go
```diff
@@ -0,0 +1,39 @@
...
+func HelpMessage() string {
+	buf := bytes.NewBuffer([]byte{})
+	writer := tabWriter(buf)
+	writeHelpMessage("", writer)
+	return buf.String()
+}
+
+// PrintHelp outputs help message to os.Stderr
+func PrintHelp() {
+	usagePrefix := "Usage:\n  dcrcli "
```
itswisdomagain: A thought... instead of using the hardcoded string `dcrcli` here, it might be "safer" in the long run to have `PrintHelp` receive a `appName` parameter, and `usagePrefix` defined instead as `usagePrefix := fmt.Sprintf("Usage:\n  %s ", appName)`

The goal of this change is to maintain uniformity when referring to the app name as used in terminal. For this goal to be properly achieved, I also recommend that the hardcoded definition of `AppName` in main.go be removed.

`config.go` has a "foolproof" way of getting the actual app name using the following code:
```
appName := filepath.Base(os.Args[0])
appName = strings.TrimSuffix(appName, filepath.Ext(appName))
```

In `main.go`,  the code above can be placed at the beginning of the `main` function and removed from `config.go` (avoid redundancies).  Instead of repeating the same code in `config.go` to get the appName, the `loadConfig` function in `config.go` can be modified to accept a `appName` string parameter (`func loadConfig(appName string)`) and the `appName` variable declared and assigned in function `main()` of `main.go` can be passed into the call to `loadConfig()` in `main.go`

Also, function `enterCliMode` in `main.go` is dependent on the `AppName` variable hardcoded at the top of `main.go` which I've recommended be **removed**. You can modify the function to instead accept a `appName` string parameter (`func enterCliMode(appName string, config *config, args []string)`) and pass in `appName` declared in the `main` function above into the call to the `enterCliMode` function.

Pretty long comment, hope my message is well communicated

cli/cli.go
itswisdomagain: `PrintHelp` is also called twice in `loadConfig` function of `config.go`, the function currently has a way to get the "actual" app name and that can be passed as argument into the modified `PrintHelp` function.

A later comment suggests having the `loadConfig` function receive `appName` as a string argument. Either ways, you'd have what you need to successfully call `PrintHelp` by passing an `appName` argument.

itswisdomagain (changes requested):
This change set is terrific! I love what you've done with the whole code restructuring!

Just some minor code cleanups/improvement and this should be good to merge.

I noticed in a previous commit that executing `dcrcli` without a command or with a wrong command displayed both available commands and options. Currently it only displays available commands. It'd be nice to have it display both, especially as we tell the user that the expected usage is `dcrcli [OPTIONS] <command> [<args...>]`, it might make sense that as we tell the user the commands we accept, we tell them what options are accepted.

Of course, if that's gonna be complicated to implement or will mess the code up some bit, we can sure keep things the way they are now.

Great work!

cli/cli.go
barisere: Yes, `commandListOrder` was used previously, but since I introduced `supportedCommands` it became obsolete.

barisere:
> I noticed in a previous commit that executing dcrcli without a command or with a wrong command displayed both available commands and options. Currently it only displays available commands. It'd be nice to have it display both, especially as we tell the user that the expected usage is dcrcli [OPTIONS] <command> [<args...>], it might make sense that as we tell the user the commands we accept, we tell them what options are accepted.

raedah prefers it that way. And that makes sense. If I type a wrong command, I should get command help. It's a subjective thing, but I think we should stick with this for now. If users find it okay, nothing changes; if they want the verbose output, good for them.

itswisdomagain (approved):

PR #46, opened 2018-12-04
Transfer funds using selected UTXOs

itswisdomagain:
Add functionality to specify unspent outputs to use in creating a transaction. Functionality is implemented in web interface.

This PR builds on @codemaestro64's work on #40, especially wrt to fetching unspent outputs from an account using drcwallet rpc.

Changes/additions to core `walletrpcclient` package:
- `UnspentOutputs` function in `walletrpcclient` core package to fetch ALL unspent outputs from specified account in wallet
- 2 new `Send...` functions in `walletrpcclient` replace previous `Send` function, viz:
- `SendFromAccount` function which automatically selects unspent outputs from an account to use as inputs for new transaction; and
- `SendFromUTXOs` function which accepts unique utxo keys to use as inputs for new transaction
- `NewUnsignedTx` function in new `walletcore` package (for direct wallet-related functions outside/without rpc)

On the send page of the web interface,
- Select source account, enter amount to send and destination address
- Click Next to see a list of unspent outputs in the sending account
- Select (check) outputs you want to use as inputs for new transaction or leave all unselected to automatically select inputs when creating transaction (default)
- Click Send
- After successful sending, transaction hash is displayed
- Use the hash to confirm the input used for the transaction on https://testnet.dcrdata.org

**EDIT: The web interface flow is highlighted in the following screenshots:**

Step 1: Begin send process
screen shot 2561-12-08 at 2 00 49 am

Step 2: Select unspent outputs to use (**9424fca59887e7d2e9ac7188c41ab8f02ba600eded74ba3763284903c19d7e0c:1** and **6a6890c084c23e9e8a99ebfb444081c8f314d25767fa8990c517fe2dd7049acd:1** are selected in the screenshot below). The unique key for each utxo is a combination of the transaction hash and the output index (hash:index).
screen shot 2561-12-08 at 2 01 13 am

Step 3: Send and get transaction hash if successful. Hash in screenshot is c5aa98007f0b35c8252ecd787a1a3a2de37ff1d5dcfccc165c3e41d477dc14f2
screen shot 2561-12-08 at 2 01 42 am

Step 4: Verify that the selected utxos were used as inputs for recently concluded transaction. [Transaction info can be found](https://testnet.dcrdata.org/tx/c5aa98007f0b35c8252ecd787a1a3a2de37ff1d5dcfccc165c3e41d477dc14f2) on testnet.dcrdata.org.
screen shot 2561-12-08 at 2 02 08 am


walletrpcclient/walletrpcclient.go
```diff
@@ -92,17 +90,99 @@ func (c *Client) Send(amountInDCR float64, sourceAccount uint32, destinationAddr
...
+		item, err := stream.Recv()
+		if err == io.EOF {
+			break
+		}
+
+		if err != nil {
+			return nil, err
+		}
+
+		// todo could _ be an error
```
barisere: I doubt _ can be an error: the input is from a previous transaction. Handling it is fine anyway.

itswisdomagain: ah, yes. I added the todo comment to confirm the function signature, I'll do that and update the code accordingly

walletrpcclient/walletrpcclient.go
```diff
@@ -92,17 +90,99 @@ func (c *Client) Send(amountInDCR float64, sourceAccount uint32, destinationAddr
...
+		}
+		if !useUtxo {
+			continue
+		}
+
+		outpoint := wire.NewOutPoint(transactionHash, item.OutputIndex, int8(item.Tree));
+		input := wire.NewTxIn(outpoint, item.Amount, nil)
+		inputs = append(inputs, input)
+
+		// todo compare capacity with length
```
barisere: Comparing capacity with length would be
```go
if len(inputs) == cap(inputs) { ...
```
If the stream has no duplicates, that is the same as comparing the lengths of both slices (as you currently do), because the capacity of `inputs` starts out as the length of `utxoKeys`, and because the `inputs` slice can be modified only if an output key is present in `utxoKeys`. So, the comparison (capacity = length) won't be necessary.

barisere (approved):
I've tested this feature using the web UI. It works as expected.

PR #47, opened 2018-12-08
add coin control feature to cli interface

itswisdomagain:
Follow up to PR #46, this PR adds a new command `send-custom` similar to the `send` command with the addition of prompting users to select from a list of unspent outputs to use in creating the new transaction.

The flow for `dcrcli send-custom` is:
- Select source account
- Enter destination address
- Enter amount to send
- Select from a list of unspent outputs in the source account (selection is done by inputting a range representing the utxos to use, using comma to select utxos that are not listed consecutively)
- Enter wallet passphrase
- After successful sending, transaction hash is displayed
Use the hash to confirm that the input used for the transaction as displayed on https://testnet.dcrdata.org is the same as the utxos selected when making the transaction

This screenshot exemplifies this flow:
screen shot 2561-12-08 at 8 40 59 pm

The inputs selected in that transaction are **254df1462f1d3d23b3f8af6ddd6e32f3e8331f162c50fef21d886d6c57587a61:1**, **036d98eca1388feb0b3f7e5757a6a8d3a34d40eab352e9ee4c60419d5b2d2290** and **50f24453c9a0b126c63a34d29108e330e7b0b603171f0b6d4dbd04933b7815c9:0**. The final transaction hash is **d57fe68a967b2156522868e41635eaac2d76b94f53d8bec442410b87ececbdba** and the transaction inputs/outputs can be viewed [here on dcrdata](https://testnet.dcrdata.org/tx/d57fe68a967b2156522868e41635eaac2d76b94f53d8bec442410b87ececbdba)


itswisdomagain:
2 other implementation paths I considered but did not end up pursing are:

1. Trigger coin control feature using an optional flag to the single existing `send` command. The flag could have a long form `coincontrol` and short form `cc`. With this approach, `dcrcli send` sends using default method while `dcrcli send --coincontrol` and/or `dcrcli send -cc` prompts for selection of utxos to use in new transaction. The current approach uses a different command altogther `dcrcli send-custom`

2. Instead of listing all the utxos in an account all at once and asking for user to enter some range of numbers to select which to use, another approach that came to mind was displaying the utxos one at a time and asking the user to enter `yes`/`y or `no`/`n` to chose to include or not include the displayed utxo in the transaction to be created. A third possible response would be something like `done`/`d` if the user is okay with the selection so far and does not want to view more utxos to select from

Anyone think any of these approaches is worth implementing?

barisere (changes requested):
This implementation currently does not handle the case where there is no unspent output on an account. The following is the output shown.
```
$ ./dcrcli -C dcrcli.conf send-custom
Destination Address: TsfDLrRkk9ciUuwfp2b8PawwnukYD7yAjGd
Amount (DCR): 20
Select unspent outputs (e.g 1-4,6)
=> 1
Error: invalid input. Selection should be numbers between 1 - 0

Select unspent outputs (e.g 1-4,6)
=> 

```
I think the program should fail early with a message explaining the reason for the failure.

itswisdomagain:
Noted. Will apply a fix.

cli/commands.go
```diff
@@ -11,7 +11,8 @@ type command struct {
 func supportedCommands() []command {
 	return []command{
 		{"balance", "show your balance", balance},
-		{"send", "send a transaction", send},
+		{"send", "send a transaction", normalSend},
+		{"send-custom", "send a transaction manually selecting inputs from unspent outputs", customSend},
```
barisere: The description for `send-custom` should be
> send a transaction, manually selecting inputs from unspent outputs

Notice the comma? I tried to ignore it, but it kept suggested another meaning (that the transaction, not the user, is manually selecting the inputs). Who gets picky about commas :unamused:

cli/helpers.go
```diff
@@ -100,3 +102,88 @@ func getWalletPassphrase() (string, error) {
...
+			if err != nil || max < minAllowed || max > maxAllowed {
+				return errWrongInput
+			}
+
+			// ensure min is actually smaller than max, swap if otherwise
+			if min > max {
+				temp := max
+				max = min
+				min = temp
+			}
```
barisere, line 151:
```go
temp := max
max = min
min = temp
```
This is the same as
```go
min, max = max, min
```

cli/helpers.go
```diff
@@ -100,3 +102,88 @@ func getWalletPassphrase() (string, error) {
...
+		}
+
+		if len(selection) == 0 {
+			return errWrongInput
+		}
+
+		var totalAmountSelected float64
+		for _, n := range selection {
+			utxo := utxos[n-1]
```
barisere:
```go

for _, n := range selection {
	utxo := utxos[n-1]
```
It is not clear that `n` is a value that starts from 1 (hence the need for the subtraction). How about doing the subtraction somewhere closer to the point where `n` is first received (maybe on line 155)? The fresh context can help make it easier to understand.

cli/handlers.go
```diff
@@ -39,12 +39,32 @@ func balance(walletrpcclient *walletrpcclient.Client, commandArgs []string) (*re
...
+	// fetch utxos from selected account and check if user can proceed with this custom send command
+	// get all utxos in account, pass 0 amount to get all
+	utxos, err := walletrpcclient.UnspentOutputs(sourceAccount, 0)
+	if err != nil {
+		return nil, err
+	}
+	if len(utxos) == 0 {
+		return nil, fmt.Errorf("Selected account does not have any unspent outputs. Cannot proceed")
+	}
+
```
barisere, line 67: This is done regardless of whether the operation uses coin control or a normal send. The comment on line 58 is misleading. The edit history of this comment shows that.

cli/commands.go
itswisdomagain: I do! I’m a grammar fanatic. Nice catch!

cli/helpers.go
itswisdomagain: Noted. Makes sense

cli/handlers.go
itswisdomagain: 🤦🏾‍♂️

barisere (approved):

PR #48, opened 2018-12-12
Transaction history on web and cli

itswisdomagain:
This takes the work done by @codemaestro64 on #42 to completion.

Web screenshot (`Transactions page`):
screen shot 2561-12-12 at 4 49 39 pm

Terminal screenshot (`dcrcli history`):
screen shot 2561-12-12 at 4 49 24 pm


walletrpcclient/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+)
+
+func (c *Client) processTransactions(transactionDetails []*pb.TransactionDetails) ([]*Transaction, error) {
+	transactions := make([]*Transaction, 0, len(transactionDetails))
+	
+	for _, txDetail := range transactionDetails {
+		// use any of the addresses in inputs/outputs to determine if this is a testnet tx
+		var isTestnet bool
+		for _, output := range txDetail.Credits {
+			isTestnet = strings.HasPrefix(output.Address, "T")
```
itswisdomagain: There's a better way to determine this. Will update.

yaowork: take a look at how [dcrutil's address library](https://github.com/decred/dcrd/blob/master/dcrutil/address.go#L203) does it

walletrpcclient/types.go
```diff
@@ -30,5 +30,16 @@ type UnspentOutputsResult struct {
...
+type Transaction struct {
+	Hash      string  `json:"hash"`
+	Type	string 	`json:"type"`
+	Amount     float64 `json:"amount"`
+	Fee			float64 `json:"fee"`
+	Direction 	string `json:"direction"`
+	IsTestnet	bool `json"is_testnet"`
+	Timestamp int64   `json:"timestamp"`
+	FormattedTime string `json:"formatted_time"`
+}
```
barisere: You can configure your editor to format code on save.

walletrpcclient/helpers.go
```diff
@@ -0,0 +1,101 @@
...
+			FormattedTime: time.Unix(txDetail.Timestamp, 0).Format("Mon Jan 2, 2006 3:04PM"),
+		}
+
+		transactions = append(transactions, tx)
+	}
+	
+	return transactions, nil
+}
+
+func transactionAmountAndDirection(txDetail *pb.TransactionDetails) (int64, string) {
```
barisere: There is more than one concern in this function. Can it be split? One function to get transaction direction, another to get amount (maybe three, for output amount, input amount, and the actual amount).
Also, using a type for the transaction direction can improve readability at the call sites.

itswisdomagain: Can't be split without repeating code, because the amount and direction is determined in almost the same way.
The code to sum the input and output amount is really brief and isn't needed elsewhere, doesn't really make much sense to create separate functions for them...

barisere: I can see a way to split it without repeating the work, but I agree it should that way for now because it is used in only that sense.

barisere (approved):

Issue #54, opened 2018-12-12
send and receive cmd output


raedah:
```
[x@rose dcrcli]$ ./dcrcli send
Destination Address:
invalid address

Destination Address:
invalid address

```
Correct msg would be 'You did not specify an address.' and then just exit. 

Otherwise, "That is not a valid address. Try again."
Do not have unneeded blank lines in the output when re-asking the question.

raedah:
Similiar for amount
```
Amount (DCR):
error parsing amount: strconv.ParseFloat: parsing "": invalid syntax
```
You did not specify an amount.' and then just exit. 
"That is not a valid amount. Try again."

raedah:
On success the output looks like this..
```
Amount (DCR): 2.123
Wallet Passphrase:                                      Result | Hash
                                            |
 The transaction was published successfully | b5f844895a4e5b4e496929be0356e8435a8c30976d6bd59a29539fddd37fc9d7
```
better would be
```
Amount (DCR): 2.123
Wallet Passphrase:
You are about to send 2.123 DCR to TsgQwgjtgG7KB68pJwyUAq54uZ6GW1V73dJ
Are you sure? (y/n): y
sent txid b5f844895a4e5b4e496929be0356e8435a8c30976d6bd59a29539fddd37fc9d7
```

raedah:
```
$ ./dcrcli receive
                             Address | QR Code
                                     |
 TsVhcKRQbTGwpVUX1TQJKMPDyTknKRBfkVW |

[Giant QR code]

```
better would be
```
$ ./dcrcli receive
TsVhcKRQbTGwpVUX1TQJKMPDyTknKRBfkVW
[QR code]

```
Lets make the qr code smaller. Smallest usable would be preferred.

Jujhar:
I am working on this.

itswisdomagain:
@Jujhar, one good way to fix this cli output message across the cli package is to do the following:

1. Currently each command handler has this signature
`handler func(walletrpcclient *walletrpcclient.Client, commandArgs []string) (*response, error)`

2. The *response returned from executing each command is printed to stdOut, if there's an error, the error is instead printed to stdErr

If we change the signature of the command handler to
`handler func(walletrpcclient *walletrpcclient.Client, commandArgs []string) error` so that each command can only return an error if there's one, and then each command prints it's own message after successful execution.

That way, each command handler can print it's success message in a way different from others and mustn't be forced to use the same pattern.

[...]

Issue #53, opened 2018-12-12
simplify output of running ./dcrcli without a command


raedah:
Current output is
```
$ ./dcrcli
Usage:
  dcrcli [OPTIONS] <command> [<args...>]

Available commands:

     balance | show your balance
        send | send a transaction
 send-custom | send a transaction, manually selecting inputs from unspent outputs
     receive | show your address to receive funds

```
This can be simplified to.
```
$ ./dcrcli
available cmds:  balance, send, receive
experimental:  sendcustom

```

Issue #49, opened 2018-12-12
balance command simple and detailed


raedah:
The balance command by default should have simple output, which is, just returning the Total. For the more detailed view, an argument would be passed.

This style would be duplicated across other UI's as well, with the default balance showing only the Total, and a detailed view being available by user selection.

raedah:
If Spendable is different then Total, then display like this.

Total 978.1184554 DCR  (Spendable 64.17666807 DCR)

ademuanthony:
I can't see the optional arg that should be passed if the user wants more information on the cli.
Same goes for the web; the option to select if more detail is required is still missing.

itswisdomagain:
This issue has only been partially resolved. The required changes have not been implemented on web interface

itswisdomagain:
Fix completed in #102

PR #63, opened 2018-12-16
[WIP] ./dcrcli help

### LINKED ISSUE #51: ./dcrcli help
Add a ./dcrcli help   command.  It will be a more verbose version of the simple help that is returned when just running  ./dcrcli  by itself. Also have  ./dcrcli help cmd  return just the extended help for that one command.

ademuanthony:
fixes [51](https://github.com/raedahgroup/dcrcli/issues/51). 
Added a './dcrcli help [command]' command that shows general help message or specific help message when called with the optional [command] param
iss-51-all

### DIFF
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -18,5 +18,6 @@ func supportedCommands() []*command {
 		{"receive", "show your address to receive funds", "dcrcli receive [account]", receive, false},
 		{"history", "show your transaction history", "dcrcli history", transactionHistory, false},
 		{"help", "show the command line help", "dcrcli help [command]", help, false},
+
 	}
 }
--- a/cli/commands.go
+++ b/cli/commands.go
@@ -4,17 +4,19 @@ package cli
 type command struct {
 	name        string
 	description string
+	usage       string
 	handler     handler
 	experimental bool
 }
 
 // supportedCommands provides the commands available from the cli
-func supportedCommands() []command {
-	return []command{
-		{"balance", "show your balance", balance, false},
-		{"send", "send a transaction", normalSend, false},
-		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", customSend, true},
-		{"receive", "show your address to receive funds", receive, false},
-    	{"history", "show your transaction history", transactionHistory, false},
+func supportedCommands() []*command {
+	return []*command{
+		{"balance", "show your balance", "dcrcli balance", balance, false},
+		{"send", "send a transaction", "dcrcli send", normalSend, false},
+		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", "dcrcli send-custom", customSend, false},
+		{"receive", "show your address to receive funds", "dcrcli receive [account]", receive, false},
+		{"history", "show your transaction history", "dcrcli history", transactionHistory, false},
+		{"help", "show the command line help", "dcrcli help [command]", help, false},
 	}
 }
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -201,22 +201,14 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
 	if len(commandArgs) == 0 {
-		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n\n"
-		usageText := "Usage:\n\n    dcrcli [OPTIONS] <command> [<args...>]\n\nAvailable commands:"
-		res = &response{
-			columns: []string{header + usageText},
-		}
-		commands := supportedCommands()
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
+		fmt.Println(header)
+		PrintHelp("")
 
-		for _, command := range commands {
-			item := []interface{}{
-				command.name,
-				command.description,
-			}
-			res.result = append(res.result, item)
-		}
-		additionalHelp := "\n\nUse \"dcrcli help <command>\" for more information about a command."
-		res.result = append(res.result, []interface{}{additionalHelp})
+		additionalHelp := "\nUse \"dcrcli help <command>\" for more information about a command."
+		fmt.Println(additionalHelp)
+
+		return &response{}, nil
 	} else {
 		cmdText := commandArgs[0]
 		commands := supportedCommands()
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -232,7 +232,7 @@ func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) 
 			return nil, fmt.Errorf("Invalid command, %s", cmdText)
 		}
 
-		text := fmt.Sprintf("%s - %s\n\nUsage:\n\n    ./dcrcli %s", command.name, command.description, command.name)
+		text := fmt.Sprintf("%s - %s\n\nUsage:\n\n    %s", command.name, command.description, command.usage)
 		res = &response{
 			columns:[]string{text},
 		}
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -8,8 +8,8 @@ import (
 	qrcode "github.com/skip2/go-qrcode"
 )
 
-func balance(walletrpcclient *rpcclient.Client, commandArgs []string) (*response, error) {
-	balances, err := walletrpcclient.Balance()
+func balance(c *cli, commandArgs []string) (*response, error) {
+	balances, err := c.walletrpcclient.Balance()
 	if err != nil {
 		return nil, err
 	}
@@ -39,16 +39,17 @@ func balance(walletrpcclient *rpcclient.Client, commandArgs []string) (*response
 	return res, nil
 }
 
-func normalSend(walletrpcclient *rpcclient.Client, _ []string) (*response, error) {
-	return send(walletrpcclient, false)
+func normalSend(c *cli, _ []string) (*response, error) {
+	return send(c, false)
 }
 
-func customSend(walletrpcclient *rpcclient.Client, _ []string) (*response, error) {
-	return send(walletrpcclient, true)
+func customSend(c *cli, _ []string) (*response, error) {
+	return send(c, true)
 }
 
-func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
+func send(c *cli, custom bool) (*response, error) {
 	var err error
+	walletrpcclient := c.walletrpcclient
 
 	sourceAccount, err := getSendSourceAccount(walletrpcclient)
 	if err != nil {
@@ -122,7 +123,8 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
 	return res, nil
 }
 
-func receive(walletrpcclient *rpcclient.Client, commandArgs []string) (*response, error) {
+func receive(c *cli, commandArgs []string) (*response, error) {
+	walletrpcclient := c.walletrpcclient
 	var recieveAddress uint32
 
 	// if no address passed in
@@ -169,8 +171,8 @@ func receive(walletrpcclient *rpcclient.Client, commandArgs []string) (*response
 	return res, nil
 }
 
-func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*response, error) {
-	transactions, err := walletrpcclient.GetTransactions()
+func transactionHistory(c *cli, _ []string) (*response, error) {
+	transactions, err := c.walletrpcclient.GetTransactions()
 	if err != nil {
 		return nil, err
 	}
@@ -199,9 +201,9 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 	return res, nil
 }
 
-func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
+func help(_ *cli, commandArgs []string) (res *response, err error) {
 	if len(commandArgs) == 0 {
-		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
+		header := "Dcrcli is a command-line utility that interfaces with the Decred wallet.\n"
 		fmt.Println(header)
 		PrintHelp("")
 
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -201,9 +201,9 @@ func transactionHistory(c *cli, _ []string) (*response, error) {
 	return res, nil
 }
 
-func help(_ *cli, commandArgs []string) (res *response, err error) {
+func help(c *cli, commandArgs []string) (res *response, err error) {
 	if len(commandArgs) == 0 {
-		header := "Dcrcli is a command-line utility that interfaces with the Decred wallet.\n"
+		header := fmt.Sprintf("%s is a command-line utility that interfaces with the Decred wallet.\n", c.appName)
 		fmt.Println(header)
 		PrintHelp("")
 
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -198,3 +198,44 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 	return res, nil
 }
+
+func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
+	if len(commandArgs) == 0 {
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n\n"
+		usageText := "Usage:\n\n    ./dcrcli [OPTIONS] <command> [<args...>]\n\nAvailable commands:"
+		res = &response{
+			columns:[]string{header + usageText},
+		}
+		commands := supportedCommands()
+
+		for _, command := range commands {
+			item := []interface{}{
+				command.name,
+				command.description,
+			}
+			res.result = append(res.result, item)
+		}
+		additionalHelp := "\n\nUse \"./dcrcli help <command>\" for more information about a command."
+		res.result = append(res.result, []interface{}{additionalHelp})
+	}else {
+		cmdText := commandArgs[0]
+		commands := supportedCommands()
+		var command command
+		var found bool
+		for _, cmd := range commands {
+			if cmd.name == cmdText {
+				command = cmd
+				found = true
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("Invalid command, %s", cmdText)
+		}
+
+		text := fmt.Sprintf("%s - %s\n\nUsage:\n\n    ./dcrcli %s", command.name, command.description, command.name)
+		res = &response{
+			columns:[]string{text},
+		}
+	}
+	return
+}
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -204,7 +204,7 @@ func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) 
 		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n\n"
 		usageText := "Usage:\n\n    dcrcli [OPTIONS] <command> [<args...>]\n\nAvailable commands:"
 		res = &response{
-			columns:[]string{header + usageText},
+			columns: []string{header + usageText},
 		}
 		commands := supportedCommands()
 
@@ -217,7 +217,7 @@ func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) 
 		}
 		additionalHelp := "\n\nUse \"dcrcli help <command>\" for more information about a command."
 		res.result = append(res.result, []interface{}{additionalHelp})
-	}else {
+	} else {
 		cmdText := commandArgs[0]
 		commands := supportedCommands()
 		var command command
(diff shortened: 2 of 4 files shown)


cli/handlers.go
```diff
@@ -198,3 +198,44 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+
+		for _, command := range commands {
+			item := []interface{}{
+				command.name,
+				command.description,
+			}
+			res.result = append(res.result, item)
+		}
+		additionalHelp := "\n\nUse \"dcrcli help <command>\" for more information about a command."
+		res.result = append(res.result, []interface{}{additionalHelp})
```
itswisdomagain: aside from `header` and `additionalHelp`, the rest of the help message content can be printed by calling `PrintHelp(c.appName)`, to avoid code repetition.

also, there's a new idea suggested in #54 (https://github.com/raedahgroup/dcrcli/issues/54#issuecomment-447550604), that proposes that command handlers no longer return a response object, thus allowing each command the freedom to display their output messages as best appropriate.

for now, this command could return nil for `res` and do all the needed printing here.

cli/commands.go
```diff
@@ -4,16 +4,18 @@ package cli
...
-		{"balance", "show your balance", balance},
-		{"send", "send a transaction", normalSend},
-		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", customSend},
-		{"receive", "show your address to receive funds", receive},
-		{"history", "show your transaction history", transactionHistory},
+		{"balance", "show your balance", "dcrcli balance", balance},
+		{"send", "send a transaction", "dcrcli send", normalSend},
+		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", "dcrcli send-custom", customSend},
+		{"receive", "show your address to receive funds", "dcrcli receive", receive},
+		{"history", "show your transaction history", "", transactionHistory},
```
itswisdomagain: I think the usage here is `dcrcli history`

itswisdomagain: something I've noticed about the `usage` for all commands is, it's basically same as concatenating the app name `dcrcli` with the command name `send` (for example)

if we can "construct" the usage from what we already have, there might be no need to have a separate property for it in the `command` struct

itswisdomagain: In line with #53, what I think might be worth introducing is a boolean `experimental` property, and also a means to track information about required and optional command arguments that may be used with each command.

An example is the receive command. That command accepts an account number argument by default. e.g `dcrcli receive 0`. If no argument is passed, the user is prompted to select an account.
I'm thinking the extended help for the receive command should indicate that this command works with an optional argument.

itswisdomagain (changes requested):

cli/commands.go
itswisdomagain: The arguments to a command could also be a flag (for boolean arguments) such as `-l/--list` or `-e/--extended` for the use case presented in #49

ademuanthony: I was thinking it may be easier to provide the usage since some commands may require addition args, options/flags and may even have sub-commands in future

itswisdomagain: You're right, it may be easier. Not sure though. But I'm inclined to go that way (for now). With this approach, command usages will need to provide information on required/optional options/flags, such as the `receive` command explained above. So maybe we can start with that, let's make the usage description for the receive command show it's optional argument.

Also, history command does not have any usage text yet...

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 	return res, nil
 }
+
+func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
+	if len(commandArgs) == 0 {
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
```
itswisdomagain: We've preferred not hardcoding the app name since it could (will likely) change in the future, there should be another way to use the "foolproof" app name stored in the `cli` struct

itswisdomagain: Also, dcrwallet rpc is just one possible means by which dcrcli interfaces with a decred wallet. Another mode of interfacing is in the works, we should probably adjust this help text accordingly, you know, not limit it to just dcrwallet rpc.

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+		fmt.Println(header)
+		PrintHelp("")
+
+		additionalHelp := "\nUse \"dcrcli help <command>\" for more information about a command."
```
itswisdomagain: Same comment about hardcoded app name applies here also

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 	return res, nil
 }
+
+func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
+	if len(commandArgs) == 0 {
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
+		fmt.Println(header)
+		PrintHelp("")
```
itswisdomagain: calling `HelpMessage()` gives you the string representation of the current help information, that might be more readable in this context, seeing as `header` and `additionalHelp` are declared then printed.

It'd aid readability to also have something like `help := HelpMessage()` after `header` and before `additionalHelp`

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+		return &response{}, nil
+	} else {
+		cmdText := commandArgs[0]
+		commands := supportedCommands()
+		var command command
+		var found bool
+		for _, cmd := range commands {
+			if cmd.name == cmdText {
+				command = cmd
+				found = true
```
itswisdomagain: `found` is not necessary, to know if the command was found later, you can check if `command == nil`
also, you should `break` once you find the command you're looking for, really no need to keep the loop going

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
+		fmt.Println(header)
+		PrintHelp("")
+
+		additionalHelp := "\nUse \"dcrcli help <command>\" for more information about a command."
+		fmt.Println(additionalHelp)
+
+		return &response{}, nil
+	} else {
+		cmdText := commandArgs[0]
```
itswisdomagain: `cmdText` might be better called `cmdName`, line 218 helps with this line of thought

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+func help(_ *rpcclient.Client, commandArgs []string) (res *response, err error) {
+	if len(commandArgs) == 0 {
+		header := "Dcrcli is a command-line utility that interfaces with Dcrwallet rpc's methods.\n"
+		fmt.Println(header)
+		PrintHelp("")
+
+		additionalHelp := "\nUse \"dcrcli help <command>\" for more information about a command."
+		fmt.Println(additionalHelp)
+
+		return &response{}, nil
```
itswisdomagain: This return is unnecessary in my opinion
if you do not return anything here, the program continues to after the if/else block and just returns, thus using the named return variables, which will be nil (and that's fine, to be expected)

the only change this would require is a change to line 83 of `cli.go` (`RunCommand` fn), there should be a check if `res != nil` before a print is attempted. Other PRs already do this check in recognition of the fact that command handlers would no longer need to return a response, since each handler would ideally (going forward) print their own successful responses and only return an error otherwise for the `RunCommand` fn to handle.

Once all handlers have been modified to print their own success messages, we'd revisit the signature of the `handler` struct to remove the first return object.

cli/handlers.go
```diff
@@ -198,3 +198,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+				command = cmd
+				found = true
+			}
+		}
+		if !found {
+			return nil, fmt.Errorf("Invalid command, %s", cmdText)
+		}
+
+		text := fmt.Sprintf("%s - %s\n\nUsage:\n\n    %s", command.name, command.description, command.usage)
+		res = &response{
```
itswisdomagain: In line with one of the comments above, let's print the "success" message here directly, and return nil for `res`

itswisdomagain: A thought.... can we do without the empty line space for usage text?
screen shot 2561-12-18 at 9 18 41 am

Put it in 1 line like `Usage: dcrcli receive` instead? I don't think it hurts readability but if you think otherwise, we should leave it as is.

cli/help.go
```diff
@@ -24,7 +24,7 @@ func PrintHelp(appName string) {
 
 func writeHelpMessage(prefix string, w *tabwriter.Writer) {
 	res := &response{
-		columns: []string{prefix + "[OPTIONS] <command> [<args...>]\n\nAvailable commands:"},
+		columns: []string{prefix + "dcrcli [OPTIONS] <command> [<args...>]\n\nAvailable commands:"},
```
itswisdomagain: `prefix` was supposed to hold the app name where necessary, to prevent hardcoding the app name. hardcoding it yet, produces this effect when you run `dcrcli -h`
screen shot 2561-12-18 at 9 21 31 am

itswisdomagain (changes requested):
Took my first real dive into this PR. Here are some thoughts to consider...

cli/handlers.go
ademuanthony: I cannot do `command == nil` as `command` is not nullable
Thanks for the `break`, was a real oversight

itswisdomagain:
The area in red looks off...
screenshot 2018-12-19 at 1 36 23 pm

itswisdomagain:
Would it hurt readability to take off the empty line space?
screenshot 2018-12-19 at 1 39 13 pm

Also, the receive usage text can benefit from the pattern established above. Instead of `Usage: dcrcli receive [<args...>]`, how about being more specific about what "args..." is/are like `Usage: dcrcli receive [account number]` or `Usage: dcrcli receive [account]`

I prefer the latter, as I also want to suggest we allow users enter account names OR account numbers. dcrwallet rpc has a method to retrieve account number using account name. And users would/may find it easier to remember account name than account number. Will create a separate issue for that with @raedah's approval.

itswisdomagain:
There's a concept of short and long description for each command. I think (hope) @barisere's research into supporting command options (amongst other things) will bring this to life.

It'd come in quite handy for this use case, showing the short description when listing all the commands in response to the `dcrcli help` command, while showing the long description when responding to `dcrcli help [command]`

Something like the `send` command would especially benefit from this. When a user requests "detailed" help for the `send` command, it would be "helpful" to mention that the command would require certain information progressively, such as the source/sending account, amount to send, address(es) to send to and the private passphrase for the wallet.

cli/handlers.go
itswisdomagain: `additionalHelp` still uses hardcoded app name

cli/handlers.go
itswisdomagain: Now that you mention it, we should make `supportedCommands()` return `[]*command`. Using a pointer really makes more sense.

Then you can make type of command `*command` and then you can perform a nil check

itswisdomagain (changes requested):

ademuanthony:
> There's a concept of short and long description for each command. I think (hope) @barisere's research into supporting command options (amongst other things) will bring this to life.
> 
> It'd come in quite handy for this use case, showing the short description when listing all the commands in response to the `dcrcli help` command, while showing the long description when responding to `dcrcli help [command]`
> 
> Something like the `send` command would especially benefit from this. When a user requests "detailed" help for the `send` command, it would be "helpful" to mention that the command would require certain information progressively, such as the source/sending account, amount to send, address(es) to send to and the private passphrase for the wallet.

Should we add this to the command struct and use it to kind add full documentation for each command? Like to try and explain all the options and args (both required and optional ones) for each command

ademuanthony:
help-screen-211912

cli/handlers.go
```diff
@@ -198,3 +201,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
 	return res, nil
 }
+
+func help(c *cli, commandArgs []string) (res *response, err error) {
+	if len(commandArgs) == 0 {
+		header := fmt.Sprintf("%s is a command-line utility that interfaces with the Decred wallet.\n", c.appName)
+
+		buf := bytes.NewBuffer([]byte{})
+		writer := tabWriter(buf)
+		writeHelpMessage(fmt.Sprintf("Usage: \n%s", c.appName), writer)
```
itswisdomagain: Could use a single space after app name, the output currently jams it all up

cli/handlers.go
```diff
@@ -198,3 +201,36 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+		commands := supportedCommands()
+		var command *command
+		for _, cmd := range commands {
+			if cmd.name == cmdName {
+				command = cmd
+				break
+			}
+		}
+		if command == nil {
+			return nil, fmt.Errorf("Invalid command, %s", cmdName)
```
itswisdomagain: comma or colon, I think colon
`invalid command: received`, instead of `invalid command, recieved`

itswisdomagain: comma or colon, I think colon
`invalid command: received`, instead of `invalid command, received`

itswisdomagain (changes requested):

ademuanthony:
072012-help

itswisdomagain:
> Should we add this to the command struct and use it to kind add full documentation for each command? Like to try and explain all the options and args (both required and optional ones) for each command

Yes please. Also, the current short descriptions might need some work. For example, @raedah pointed out how unhelpful the description for the balance command is:
`balance | show your balance`
That could probably be `shows total balance for each account in wallet`, long description could go on to say, `also shows spendable balance if different from total balance. Use -e/--extensive to view a more detailed balance breakdown`

_forgive this coming late, I wrote the comment hours ago with the other review but somehow did not hit the comment button_

ademuanthony:
Resolved conflict

cli/commands.go
```diff
@@ -4,17 +4,20 @@ package cli
...
-		{"balance", "show your balance", balance, false},
-		{"send", "send a transaction", normalSend, false},
-		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", customSend, true},
-		{"receive", "show your address to receive funds", receive, false},
-    	{"history", "show your transaction history", transactionHistory, false},
+func supportedCommands() []*command {
+	return []*command{
+		{"balance", "show your balance", "dcrcli balance", balance, false},
+		{"send", "send a transaction", "dcrcli send", normalSend, false},
+		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", "dcrcli send-custom", customSend, false},
```
itswisdomagain, line 17: This ought to be experimental. Probably made the mistake when fixing conflicts

itswisdomagain (changes requested):
1 small mixup in the latest changes. Also remember to give attention to my last comment about better short and long descriptions for each command

itswisdomagain:
We can borrow some of your recent work on #64, wrt providing a "better" way to access the `AppName` and thus stick to the principle of least privileges, passing on `walletrpcclient` to the command handlers and not the entire `cli` struct.

When that PR is merged, this change would become inevitable. Why not we do it now?

itswisdomagain:
Should we put this PR on pause? There's some interesting work going on with #71 that would alter how the commands are currently structured and probably make it easier for the command-specific help command to work as expected. @ademuanthony please take a look at that PR and share your thoughts

ademuanthony:
> Should we put this PR on pause? There's some interesting work going on with #71 that would alter how the commands are currently structured and probably make it easier for the command-specific help command to work as expected. @ademuanthony please take a look at that PR and share your thoughts

The work going on at [#71](https://github.com/raedahgroup/dcrcli/pull/71) will change a lot of things. 
I think we should pause this until that is done

itswisdomagain:
The PR (#71) that was blocking this has been completed and merged. Work on this can continue. When it's ready for review, rename the PR to remove the [WIP] label

The changes in #71 includes detailed help message printing with `dcrcli -h` and command-specific help message printing with `dcrcli <command> -h`. After conferring with @raedah, this PR remains relevant: the goal as mentioned in issue #51 is to introduce a new command `dcrcli help`. The expected behaviour of this new command would be similar to `dcrcli -h` and `dcrcli <command> -h`

**Next steps for this PR**
- **Regarding `dcrcli help`**, I think the output of `dcrcli -h` is okay as is, so `dcrcli help` can simply trigger that functionality.
- **Regarding `dcrcli help <command>`**, the similar syntax (`dcrcli <command> -h`) isn't exactly there yet. This PR _may_ not need to alter the functionality of `dcrcli <command> -h`, it probably should instead focus on achieving the concept expressed below for `dcrcli help <command>`

**Expected behaviour for `dcrcli help <command>`**
- [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448583095):
`dcrcli help <command>` should clearly show the kind of arguments `<command>` works with. For example, `dcrcli help receive` should show `Usage: dcrcli receive [account]`

- [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448585542):
`dcrcli help <command>` should include additional information that describe how the command works. For example, `dcrcli help send` should mention that `the command would require certain information progressively, such as the source/sending account, amount to send, address(es) to send to and the private passphrase for the wallet`

- [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448952265):
`dcrcli help <command>` should give information about flags/options that `<command>` works with. For example, `dcrcli help balance` should show `Usage: dcrcli balance [-d/--detailed]`. It should go on to explain what `-d/--detailed` does: `Display detailed account balance report`

PS: The behaviour depicted above are not hard and fast rules, just how I imagine/expect a verbose help message to appear. Alternative/supplementary ideas are definitely welcome

barisere:
>     * **Regarding `dcrcli help <command>`**, the similar syntax (`dcrcli <command> -h`) isn't exactly there yet. This PR _may_ not need to alter the functionality of `dcrcli <command> -h`, it probably should instead focus on achieving the concept expressed below for `dcrcli help <command>`
> 
> 
> **Expected behaviour for `dcrcli help <command>`**
> 
>     * [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448583095):
>       `dcrcli help <command>` should clearly show the kind of arguments `<command>` works with. For example, `dcrcli help receive` should show `Usage: dcrcli receive [account]`

```
$ ./dcrcli -C dcrcli.conf receive -h
Usage:
  dcrcli [OPTIONS] receive [account]
To view application options, use 'dcrcli -h'
```
[OPTIONS] is printed because the top-level application options are still relevant for running the command. The last line of the message tells the user how to view those options.
>     * [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448585542):
>       `dcrcli help <command>` should include additional information that describe how the command works. For example, `dcrcli help send` should mention that `the command would require certain information progressively, such as the source/sending account, amount to send, address(es) to send to and the private passphrase for the wallet`

The additional information you speak of can be placed in the `long-description` struct tag of the command. If you add `long-description` tag to `BalanceCommand` as follows, you get that additional information in the help output for `balance`.
```go
type CliCommands struct {
	Balance    BalanceCommand    `command:"balance" description:"show your balance" long-description:"Voluptate sunt consectetur eiusmod nulla tempor et amet non nostrud nulla consequat cillum aute laborum"`
	// other commands
}
```
```
$ ./dcrcli -C dcrcli.conf balance -h
Usage:
  dcrcli [OPTIONS] balance [balance-OPTIONS]

Voluptate sunt consectetur eiusmod nulla tempor et amet non nostrud nulla consequat cillum aute laborum

[balance command options]
      -d, --detailed  Display detailed account balance report
To view application options, use 'dcrcli -h'
```
Observe that the long description is shown for command-specific help, while the short description is shown for the top-level (overview) help.
It may be a good idea to make the short description the first sentence of the long description.
>     * [From this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448952265):
>       `dcrcli help <command>` should give information about flags/options that `<command>` works with. For example, `dcrcli help balance` should show `Usage: dcrcli balance [-d/--detailed]`. It should go on to explain what `-d/--detailed` does: `Display detailed account balance report`
> 
See the output of `$ ./dcrcli -C dcrcli.conf balance -h` above. The information about flags/options are included in the help output.

itswisdomagain:
@ademuanthony regarding this https://github.com/raedahgroup/godcr/pull/86#issuecomment-450521416:

> Though the circular dependency still continues for my current use case as I want to reference the CliRoot in the cli package in commands; whereas the cli package already references the commands package.
> 
> So I am thinking, what if the commands are moved to the cli package, seeing that that is the main component in that package, it may not be that bad to have it at the root.
> It may look like we will lose the organization that the current structure provides, but with the current convention of adding 'Command' suffix to the command names, I think it makes sense enough to have cli.SendCommand

Can you the share the code snippet where you use `CliRoot` in the `cli` package. I had similar circular dependency issue while working on #88, may be able to offer some assistance

ademuanthony:
I have continued this work at https://github.com/raedahgroup/godcr/pull/98

PR #60, opened 2018-12-14
Mobilewallet mode

itswisdomagain:
This PR introduces a second **method** for connecting to a decred wallet asides via a running `dcrwallet` deamon: `mobilewallet` library.

`mobilewallet` (will soon be renamed to `dcrlibwallet`) is a self-contained wallet library that provides decred wallet functionalities standalone, without requiring a connection to a running dcrwallet daemon.

With this PR, `mobilewallet` is the default wallet connection method used by `dcrcli`; with the option of using `dcrwallet` by setting `UseWalletRPC=1` in dcrcli.conf or passing `-u/--usewalletrpc` in command-line options.

**Notes**
There are (potentially) rare cases where a user runs `dcrwallet` without loading a wallet (i.e. `dcrwallet --noinitialload`). This PR takes care of such scenarios by providing create wallet and sync blockchain functionality for dcrwallet rpc connections.

In cli mode, command execution outputs/results with mobilewallet may be inaccurate if the self-contained mobilewallet blockchain isn't up to date. For accuracy of results, commands should be executed immediately after a sync operation. Preferably, this should be done by running `dcrcli <command> --sync`. Alternatively, you can manually trigger blockchain update by running `dcrcli --sync` and subsequently run `dcrcli <command>`.

Operation result accuracy is not a concern when running dcrcli in http mode because blockchain sync is always performed when the server is started and the web interface disables wallet-related operations until the sync is completed. After the sync is completed, and for as long as the web server stays running, block update notifications are subscribed to and the wallet database remains up to date.

**Expected behaviour**
This PR should be able to perform the following operations regardless of whether wallet connection is established via dcrwallet and mobilewallet, producing very identical outputs in both cases.

**For cli**

| Operation | Status | Notes
|---|---|---|
| `dcrcli -v` | :white_check_mark: |
| `dcrcli [command] -h` | :white_check_mark: |
| `dcrcli --createwallet` | :white_check_mark: |
| `dcrcli [command] --sync` | :white_check_mark: |
| `dcrcli balance [-d/--detailed]` | :white_check_mark: |
| `dcrcli createaccount <accountName>` _(experimental feature)_ | :white_check_mark: |
| `dcrcli receive [account-name]` | :white_check_mark: |
| `dcrcli send` | :white_check_mark: | **requires** `--sync` flag when using `mobilewallet` |
| `dcrcli send-custom` | :white_check_mark: | **requires** `--sync` flag when using `mobilewallet` |
| `dcrcli history` | :white_check_mark: |

[...]

PR #64, opened 2018-12-17
detect if dcrcli.conf is not configured

### LINKED ISSUE #50: detect if dcrcli.conf is not configured
When running ./dcrcli with no commands, currently the simple dcrcli usage cmds are shown. But these commands are not yet available if the user does not have dcrcli.conf set with the correct rpc login information. So this should be detected, and a msg displayed to let the user know that they need to specify this information. The simple usage commands would not be displayed, and instead instructions would be displayed to set edit the config file or to specify the appropriate rpc command line options (-h).

Additionally, we could autoconfigure by looking for the default dcrwallet config and auto parsing the needed settings, asking the user if they would like to use these values.

ademuanthony:
fixes [#50](https://github.com/raedahgroup/dcrcli/issues/50)
If the config file is not found and the user did not pass Dcrwallet credentials in command line options, propose to guide the user in creating a new config file and do same if the user accepts, else quit the application
detect-conf

The app will suggest default values and use them if none is provided for the RPC credential
iss-50-use-default-values


config.go
```diff
@@ -140,6 +142,17 @@ func loadConfig(appName string) (*config, []string, error) {
 				err)
 			parser.WriteHelp(os.Stderr)
 			return nil, nil, err
+		} else {
+			// The config file is not accessible
+			if preCfg.WalletRPCServer == "" || preCfg.RPCUser == "" || preCfg.RPCPassword == "" {
+				fmt.Fprintln(os.Stderr, fmt.Sprintf(`The config file, %s is missing and you did not pass the Dcrwallet RCP credetial in command line options.\n`, preCfg.ConfigFile))
```
barisere: Using backticks around the string creates an unescaped string literal. This makes the "\n" appear in the output stream literally.
Using `fmt.Fprintln` causes two lines to be printed in the output stream when the backticks are corrected to quotation marks.
Replace the backticks with quotation marks, and remove the "\n" at the end of the string as follows.
```go
fmt.Fprintln(os.Stderr, fmt.Sprintf("The config file, %s is missing and you did not pass the Dcrwallet RCP credetial in command line options.", preCfg.ConfigFile))
```
Also, the prompt message reads better as follows.
```go
fmt.Fprintln(os.Stderr, fmt.Sprintf("The config file (%s) is missing, and you did not pass dcrwallet RCP credentials in command line options.", preCfg.ConfigFile))
```

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
 }
+
+
+
+
+func loadDefaultConfig(appName string) (config *config, args []string, err error) {
+	inputErr := func(error) error {
+		return fmt.Errorf("error receiving input: %s", err.Error())
+	}
+	promtMsg := "Do you want to create the configuration file now? y|yes or n|no"
```
barisere: Options for command line input are usually shown in brackets. Line 178 would read better as follows.
```go
promtMsg := "Do you want to create the configuration file now? [y or yes / n or no]"
```
Compare with the prompt from an `apt` command.
```
Do you want to continue? [Y/n]
```

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
+	if err != nil {
+		return nil, nil, inputErr(err)
+	}
+
+	input = strings.Trim(input, "\n")
+	if input == "yes" || input == "y" {
+		// Todo load config from dcrwallet
+		rpcAddress, err := terminalprompt.RequestInput("Wallet RPC server (localhost:19111)", nil)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error receiving input: %s", err.Error())
```
barisere: You can use the helper function declared on line 175 on line 189.
```go
if err != nil {
	return nil, nil, inputErr(err)
}
```

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
+		}
+
+		if strings.Trim(rpcAddress, "\n") == "" {
```
barisere: The return value of `terminal.RequestInput` is already stripped of the ending "\n", so you can simply compare the value as follows.
```go
if rpcAddress == "" {
```
Do same for lines 200 and 208.

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
+; HTTP Server
+; ------------------------------------------------------------------------------
+
+; Server address to serve dcrcli via HTTP (when http=true or http=1)
+serveraddress=127.0.0.1:7778
+
+; Turn HTTP mode on (http=true or http=1) or off (http=false or http=0).
+; HTTP mode is off by default.
+; http=true`
+
```
barisere: Hard-coding the content of the configuration file will result in more maintenance burden than necessary.
@itswisdomagain what other approach would be cleaner and more maintainable?

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
+		_, err = f.WriteString(confText)
+		if err != nil {
+			return nil, nil, err
+		}
+
+		fmt.Println("\n\nThe config file has been set successfully\n")
+		return loadConfig(appName)
+
+	} else if input == "no" || input == "n" {
+		return config, nil, errors.New("Please set dcrcli.conf manually to use Dcrcli")
```
barisere: How about using the `appName` argument we already have access to?
```go
// return config, nil, errors.New("Please set dcrcli.conf manually to use Dcrcli")
return config, nil, fmt.Errorf("To use %s, please set config file manually, or specify appropriate command line options.", appName)
```

config.go
```diff
@@ -154,3 +167,95 @@ func loadConfig(appName string) (*config, []string, error) {
...
+	}
+
+	return config, nil, errors.New("Invalid optin args")
```
barisere: "optin" is probably a typo. Please verify that.

cli/terminalprompt/terminalprompt.go
```diff
@@ -34,6 +34,9 @@ func RequestInput(message string, validate ValidatorFunction) (string, error) {
 		if err != nil {
 			return "", err
 		}
+		if validate == nil {
```
barisere: Thanks for adding these checks :slightly_smiling_face:

barisere (changes requested):
You've done good work here. Let's tie up loose edges so we can merge this PR.

config.go
ademuanthony: I agree,
How will the app be distributed? 
If the sample-dcrcli.conf will always be available at the root, then we can use it

config.go
ademuanthony: I don't know how I missed this

config.go
```diff
@@ -154,3 +167,94 @@ func loadConfig(appName string) (*config, []string, error) {
...
+		return fmt.Errorf("error receiving input: %s", err.Error())
+	}
+	promtMsg := "Do you want to create the configuration file now? [y or yes / n or no]"
+	input, err := terminalprompt.RequestInput(promtMsg, nil)
+	if err != nil {
+		return nil, nil, inputErr(err)
+	}
+
+	if input == "yes" || input == "y" {
+		// Todo load config from dcrwallet
```
itswisdomagain: Seeing  as we're leaving this for later, it'd help to include why this "step" should be done, something like "load config from dcrwallet to auto collect the values below instead of requesting user input"...

config.go
```diff
@@ -154,3 +167,94 @@ func loadConfig(appName string) (*config, []string, error) {
...
+		rpcAddress, err := terminalprompt.RequestInput("Wallet RPC server (localhost:19111)", nil)
+		if err != nil {
+			return nil, nil, inputErr(err)
+		}
+
+		if rpcAddress == "" {
+			rpcAddress = "localhost:19111"
```
itswisdomagain: default variables should be defined where you have line 19-23

itswisdomagain (changes requested):

cli/terminalprompt/terminalprompt.go
itswisdomagain: This is nice. Let's go all the way by removing `EmptyValidator` declared in line 12, and replace all its usages with `nil`

itswisdomagain (changes requested):

config.go
itswisdomagain: I'll get this PR's code and take it for a spin later today. I agree with both of you, a better, cleaner way would definitely be preferred. I'll share more thoughts later

barisere: Can we set this at build time?
```sh
go build -ldflags "-X main.sampleConfig=`cat sample-dcrcli.conf`"
```
Instead of filling in the values for rpc address, username, and password, we can append the necessary configuration lines to the `sampleConfig` variable and write it to disk.
Setting the build flag will be done on every build. This means we will use a build tool such as Make to ensure the build steps are followed.

itswisdomagain: 2 thoughts here. Both came as a result of thinking about these 2 questions: 
- Do we really need a pre-existing sample config file?
- How would a regular user "create" his config file if there was no option to auto-generate one for the user?

Answers:
- No, the `sample-dcrcli.conf` is for development purposes (I believe), it does not get included in the release build
- Taking `dcrd` binary as an example, without building from source or having the source code (which is not necessary to run `dcrd`), how would you prepare the config file? `dcrd -h` tells you the configurable options and also tells you where the program expects to find the default config file. What I'd do is create a file in that location and set values for only the options I'm interested in as I can see from the output of the `dcrd -h` command.

These thoughts led to these options. The premise is that, in order to prepare the config file content, we don't need to have some sample file or base text stored somewhere to "edit":

Option 1:
Create the file content **_using just the information_** we've collected from the user or from parsing dcrwallet config. Right now, that would look something like this:
```
var conf []string

rpcAddress, err := terminalprompt.RequestInput(fmt.Sprintf("Wallet RPC server (%s)", defaultRpcAddress), nil)
if err != nil {
	return nil, nil, inputErr(err)
}
if rpcAddress == "" {
	conf = append(conf, fmt.Sprintf("walletrpcserver=%s", defaultRpcAddress))
} else {
	conf = append(conf, fmt.Sprintf("walletrpcserver=%s", rpcAddress))
}
```

Progressively append each option to the `conf` slice as you get it. Then join the slice using `strings.Join(conf, "\n")` and write that to the default config file location. Some/most of the comments in sample-dcrcli.conf aren't relevant anyways. `dcrcli -h` already explains what each option is for.

Option 2:
Assuming, we want to generate a more complete config file, even if we have to comment out the options that we do not want to set values for at the moment. And we also would love to include some comments about each of these options in the file, so that when a user goes to edit subsequently, he wouldn't feel lost or need to run back to `dcrcli -h` to catch his breath.

We can, at runtime, get all the available options (their long name and descriptions) and use that to create a more robust content for the config file. The code below gives the idea:
```
group := preParser.Groups()[0]
for _, option := range group.Options() {
	fmt.Println(option.LongName, option.Description)
}
```

barisere: @itswisdomagain the downside to option 1 is that the configuration file generated for the user will be missing essential information for further customization.
The downside to option 2 is that the content of the generated configuration file may not match the sample-dcrcli.config file. Users (including developers) may come across that file and be confused for a moment what the differences mean. I hope we can avoid that.

A third option is to generate the configuration instead of hard-coding it in the sample file. That way, the code is the single source of truth and will serve both the developers and the users.
The technique is as follows.
We'll use the `text/template` package to define the format of the file. Then we define a command, `configure`, that will write out the sample configuration using either user-supplied values for the fields or default values where no input was provided by the user.
Running the `configure` command becomes the first step after downloading the binary. We can even provide a helper script that does that automatically as part of the installation step.

I'll make a demo of this option tomorrow and ask for feedback.

itswisdomagain: I think this is what's holding this PR. @ademuanthony working on this?

ademuanthony:
Add a new handler; 'dcrcli configure' that will create the dcrcli.conf file for the user.
config-file-032112

the file is now generated using the 'text/template' package and the sample config is now removed leaving the string holding the template as the single source of truth for the configuration file.

cli/commands.go
```diff
@@ -15,6 +17,7 @@ func supportedCommands() []command {
 		{"send", "send a transaction", normalSend, false},
 		{"send-custom", "send a transaction, manually selecting inputs from unspent outputs", customSend, true},
 		{"receive", "show your address to receive funds", receive, false},
-    	{"history", "show your transaction history", transactionHistory, false},
+		{"history", "show your transaction history", transactionHistory, false},
+		{"configure", fmt.Sprint("help in properly setting up %s", AppName()), configure, false},
```
itswisdomagain: This should be `fmt.Sprintf`

itswisdomagain: Also, I'm thinking `config` instead of `configure` for command name, it's shorter and descriptive/informative still

cli/helpers.go
```diff
@@ -185,3 +187,9 @@ func getUtxosForNewTransaction(utxos []*walletrpcclient.UnspentOutputsResult, se
 	}
 	return selectedUtxos, nil
 }
+
+func AppName() string {
```
itswisdomagain: The introduction of this function is nice! It also means we no longer need to have `appName` property in the `cli` struct, nor pass `appName` as a parameter to the `cli.New` function.

With this current code changes, the only usage of `cli.appName` is in line 97 of `cli.go` and that can easily be updated to call this function instead of referring to the `appName` property of `cli`. Making that change would also mean that the function `invalidCommandReceived` can be simplified, no longer attached to the `cli` struct as it would be no longer needed.

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 	return res, nil
 }
+
+
+func configure(_ *rpcclient.Client, _ []string) (resp *response, err error) {
+	inputErr := func(error) error {
+		return fmt.Errorf("error receiving input: %s", err.Error())
+	}
+	promtMsg := "Do you want to create the configuration file now? [y or yes / n or no]"
```
itswisdomagain: The user issued the command to create a config file, I'm sure there really is no need to ask this.

This question would only make sense if the user were trying to perform some other action and we notice there is no config and want to _suggest_ that the user creates one

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
 
 	return res, nil
 }
+
+
+func configure(_ *rpcclient.Client, _ []string) (resp *response, err error) {
```
itswisdomagain: I want to recommend moving this function to config.go. I'll mention my reasons in some comments below

itswisdomagain: If we do this though, it'll be difficult to register this as a command in `commands.go`. We can create a config flag for it, say `--init` and that may not be totally bad seeing as this is not an action many users will perform more than once

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+	var input, rpcAddress, rpcUsername, rpcPassword string
+	input, err = terminalprompt.RequestInput(promtMsg, nil)
+	if err != nil {
+		fmt.Println(inputErr(err))
+		return
+	}
+
+	if input == "yes" || input == "y" {
+		// Todo load config from dcrwallet to auto collect the values below instead of requesting user input
+		rpcAddress, err = terminalprompt.RequestInput(fmt.Sprintf("Wallet RPC server (%s)", defaultRpcAddress), nil)
```
itswisdomagain: There's no indication to the user that
- the text in brackets are default values
- not providing any input means accepting the default value

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+			RpcAddress string
+			RpcUsername string
+			RpcPassword string
+			ServerAddress string
+		}{
+			RpcAddress:rpcAddress, RpcUsername: rpcUsername, RpcPassword: rpcPassword,
+			ServerAddress: ServerAddress,
+		}
+
+		confText :=
```
itswisdomagain: I'd suggest putting this in a `conf.tmpl` file instead, but we'd be faced with the issue of including the file with the built binary when building. To keep this file clean, I want to suggest putting this in a `dcrcli.conf.go` file, that returns this text template or that holds a function, accepting the user inputs as parameters and feeding them into a string and returning the final output (similar to what you did previously, difference is, it's going into a different file to make things cleaner and we no longer have the sample conf file to avoid confusion)

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+; ------------------------------------------------------------------------------
+
+; Server address to serve dcrcli via HTTP (when http=true or http=1)
+serveraddress={{.ServerAddress}}
+
+; Turn HTTP mode on (http=true or http=1) or off (http=false or http=0).
+; HTTP mode is off by default.
+; http=true`
+
+		f, err := os.Create(defaultConfigFile)
```
itswisdomagain: 2 issues here:
- first, the default config file is already declared in `config.go`, you can't import it here, and we should try to avoid repetition as much as possible, which is one reason I recommended moving this function to `config.go`
- second, if a user specifies a different config file location via the `-C/--configfile` option, your code will ignore that, simply because you don't have access to that info in this context. You would have access to that info if this code was in `config.go`, then instead of using the default config file, you'd use `config.ConfigFile` which would point to the user specified value or the default as the case may be

screenshot 2018-12-22 at 10 15 56 am

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+		if err != nil {
+			return nil, fmt.Errorf("error in creating config file: %v", err)
+		}
+		defer f.Close()
+
+		tmpl := template.New("main")
```
itswisdomagain: Won't a better template name be "config" instead of "main"?

cli/handlers.go
```diff
@@ -198,3 +218,114 @@ func transactionHistory(walletrpcclient *rpcclient.Client, _ []string) (*respons
...
+
+		err = tmpl.Execute(f, config)
+		if err != nil {
+			return nil, err
+		}
+
+		fmt.Println("\n\nThe config file has been set successfully\n")
+		return nil, nil
+
+	} else if input == "no" || input == "n" {
```
itswisdomagain: To make the code a little easier to read, in these circumstances, you could just check for this condition first and return, then you won't need an else block

config.go
```diff
@@ -2,6 +2,7 @@ package main
 
 import (
 	"fmt"
+	"github.com/btcsuite/go-flags"
```
itswisdomagain: usually external imports are kept separated at the bottom... any specific reason you brought this here?

config.go
```diff
@@ -140,6 +141,22 @@ func loadConfig(appName string) (*config, []string, error) {
 				err)
 			parser.WriteHelp(os.Stderr)
 			return nil, nil, err
+		} else {
```
itswisdomagain: The return statement above makes this else unnecessary, you can just write the code here, knowing it will only execute if the if statement fails validation, because if it passes validation, it'll return before getting to execute the code here

config.go
```diff
@@ -140,6 +141,22 @@ func loadConfig(appName string) (*config, []string, error) {
...
+			// The configure command should go through
+			isConfigureCommand := func(args []string) bool {
```
itswisdomagain: The logic here can be used to invoke the `configure` function (assuming we agree to bring the function to this file instead of `handlers.go`)

main.go
```diff
@@ -1,14 +1,16 @@
...
-	"strings"
-
 	"github.com/raedahgroup/dcrcli/cli"
 	"github.com/raedahgroup/dcrcli/walletrpcclient"
 	"github.com/raedahgroup/dcrcli/web"
+	"os"
+)
+
+var (
+	errConfigNotFound = errors.New("config not found")
```
itswisdomagain: This is only used once, any need to create a variable for it?

main.go
```diff
@@ -42,11 +44,15 @@ func (v *Version) String() string {
...
-	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	appName := cli.AppName()
 
 	config, args, err := loadConfig(appName)
-	if err != nil {
+	if err != nil && err != errConfigNotFound {
+		os.Exit(1)
+	}
+	if err == errConfigNotFound {
+		fmt.Println(fmt.Sprintf("The config file (%s) is missing, and you did not pass dcrwallet RCP credentials in command line options.", defaultConfigFile))
```
itswisdomagain: As per the comment on #50 
> instructions would be displayed to set edit the config file or to specify the appropriate rpc command line options (-h)

My interpretation of that information is that, if no config file is found, rpc options not set, the user should be informed of 2 alternative options he can take
- set/create the config file using the appropriate command
- pass rpc credentials by following the instructions set in `dcrcli -h`

The second option is not currently offered to the user

itswisdomagain (changes requested):
Getting this auto-config thing to work smoothly is not proving as straightforward as we'd like but we'll get it. Some more thoughts on what _could_ be done better

main.go
ademuanthony: it is referenced in config.go 158 and main.go 50 and 53

ademuanthony:
I have added an --init flag and moved the configure func to config.go.
The user can now set the config by calling 'dcrcli --init' or accepting the auto-create suggestion during startup 
init-for-config

config.go
```diff
@@ -19,6 +21,11 @@ const (
 	defaultLogDirname        = "log"
 	defaultLogFilename       = "dcrcli.log"
 	defaultHTTPServerAddress = "127.0.0.1:1234"
+
+	defaultRpcAddress  = "localhost:19111"
+	defaultRpcUsername = "rpcuser"
+	defaultRpcPassword = "rpcpass"
+	ServerAddress      = "127.0.0.1:7778"
```
itswisdomagain: Let's use `defaultHTTPServerAddress` instead, btw, `ServerAddress` doesn't need to be public. We can/should make `defaultHTTPServerAddress` = `127.0.0.1:7778`

itswisdomagain: Also, we don't need to automatically save a value for `serveraddress` to `dcrcli.conf`. The program is designed such that if there is no value set in config, the default value would be used.

If we must save a value for this option to `dcrcli.conf`, we should request it from the user. But I don't think that is relevant, we probably should just stick to having a default value to use if user doesn't manually supply one.

config.go
```diff
@@ -129,18 +137,45 @@ func loadConfig(appName string) (*config, []string, error) {
 		os.Exit(0)
 	}
 
+	//config
+	if preCfg.Init {
```
itswisdomagain: What if the file already exists and user passes the `--init` flag? I think a warning is in order before the existing config file is overwritten.

config.go
```diff
@@ -129,18 +137,45 @@ func loadConfig(appName string) (*config, []string, error) {
...
 	if err != nil {
 		if _, ok := err.(*os.PathError); !ok {
-			fmt.Fprintf(os.Stderr, "Error parsing config file: %v\n",
-				err)
+			fmt.Fprintf(os.Stderr, "Error parsing config file: %v\n", err)
 			parser.WriteHelp(os.Stderr)
 			return nil, nil, err
 		}
+
+		if preCfg.WalletRPCServer == "" || preCfg.RPCUser == "" || preCfg.RPCPassword == "" {
```
itswisdomagain: When I run ` dcrcli -c="localhost:19111" -u="rpcuser" -p="rpcpass"`, I still get the error about missing config fie (true) and not passing dcrwallet RPC credentials in command line options (false)

config.go
```diff
@@ -129,18 +137,45 @@ func loadConfig(appName string) (*config, []string, error) {
...
+			yes, err := terminalprompt.RequestInput("Do you want to create the config file now? (y/n)", nil)
+			if err != nil {
+				return nil, nil, err
+			}
+			if strings.ToLower(yes) == "y" {
+				err = configure(preCfg)
+				if err == nil {
+					return loadConfig(appName)
+				}
+				fmt.Println(err)
```
itswisdomagain: This should behave as though `--init` flag was passed, **exit** after running `configure`.

Currently, if I get to this point and say yes, it creates the config, prints out `The config file has been set successfully` but also proceeds to print out the following message which is a little confusing:
```
available cmds:  balance, send, receive, history
experimental:  send-custom
```

config.go
```diff
@@ -154,3 +189,68 @@ func loadConfig(appName string) (*config, []string, error) {
...
+	}
+
+	var rpcAddress, rpcUsername, rpcPassword string
+
+	// Todo load config from dcrwallet to auto collect the values below instead of requesting user input
+	rpcAddress, err = terminalprompt.RequestInput(fmt.Sprintf("Wallet RPC server (default = %s)", defaultRpcAddress), nil)
+	if err != nil {
+		return inputErr(err)
+	}
+
```
itswisdomagain: Can we maintain the pattern below and remove this empty line space?

config.go
```diff
@@ -154,3 +189,68 @@ func loadConfig(appName string) (*config, []string, error) {
...
+	}
+	if rpcPassword == "" {
+		rpcPassword = defaultRpcPassword
+	}
+
+	config := struct {
+		RpcAddress    string
+		RpcUsername   string
+		RpcPassword   string
+		ServerAddress string
```
itswisdomagain: As mentioned in an earlier comment, saving this to `dcrcli.conf` is unnecessary - the app will not break if this value is not set in config.

There are a couple of other config options that are needed but have default values like `RPCCert` and `NoDaemonTLS`. Their absence in `dcrcli.conf` or in the command-line options won't cause the app to break. Don't see a need to ask the user to supply such information when initialising a conf file.

If we want to ask the user to supply them, then let's make them behave like `WalletRPCServer`, `RPCUser` and `RPCPassword` config options (no default values). They must be provided either in command-line options or saved in `dcrcli.conf`. if none of the above is done, then the user should be prompted to create a config file or pass those values.

If we're using default values, let's keep using default values until a user on his own goes to set a different value in `dcrcli.conf` or passes a different value in command-line options

main.go
```diff
@@ -42,8 +39,7 @@ func (v *Version) String() string {
 }
 
 func main() {
-	appName := filepath.Base(os.Args[0])
-	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
+	appName := cli.AppName()
 
 	config, args, err := loadConfig(appName)
```
itswisdomagain: This is the only use of `appName` in this function. We can do without it. `loadConfig` is able to access `cli.AppName()` directly where needed. It currently does so in line 167 of config.go

itswisdomagain (changes requested):
We're close! Minor changes left...

itswisdomagain:
We'll need to update readme, the part about configuring dcrcli using the sample config file.

barisere (changes requested):
If the configuration directory (~/.dcrcli) does not exist, the program should create it before writing the configuration file. Currently, it does not check whether the directory exists.
```
$ ./dcrcli --init
Wallet RPC server (default = localhost:19111): 
RPC username (default = rpcuser): 
RPC password (default = rpcpass): error in creating config file: open /home/barisere/.dcrcli/dcrcli.conf: no such file or directory
```

barisere (commented):
Currently, the program does not offer a default when asking the user whether to create the configuration file immediately.
```
$ ./dcrcli
The config file (/home/barisere/.dcrcli/dcrcli.conf) is missing, and you did not pass dcrwallet RPC credentials in command line options.
Please run 'dcrcli --init' to create the config file or pass RPC credentials by following the instructions set in 'dcrcli -h'
Do you want to create the config file now? (y/n): 
```
The output of the last line of the prompt should be
```
Do you want to create the config file now? (Y/n): 
```
That indicates that "Y" is the default, so the user can simply hit enter and have the process continue. With the current implementation, the process aborts if the user does not provide an input.
Also check for wrong input. Validate that the input is either "Y" or "N", and it should be case-insensitive.

config.go
barisere: @ademuanthony That happens because the command line options have not been parsed at this point. Parse the command line arguments first (`parser.Parse`), before parsing the configuration file.

barisere: The absence of those properties does not mean that the configuration file is missing. As @itswisdomagain noted, those properties can be provided via command line options.
I have an idea what the solution may be. First, parse the command line options as I suggested in [my previous comment](https://github.com/raedahgroup/dcrcli/pull/64/files#r243846517). Then add an `else` branch to the `if` block on line 155. The `else` branch should run only if there was an `*os.PathError` and the properties you check for on line 161 are empty; i.e., the configuration file is missing, and the associated options were not provided on the command line.

barisere (changes requested):

itswisdomagain:
I just came into a shocking revelation. RPC username (`config.RPCUser`) and RPC password (`config.RPCPassword`) are currently not used in connecting to dcrwallet daemon. The `New` function of `walletrpcclient.go` only uses Wallet RPC server address (`config.WalletRPCServer`) in connecting to dcrwallet daemon.

The only exception to this is when the user specifies `config.NoDaemonTLS` to be false and a connection to dcrwallet daemon is attempted over tls. Then RPC server certificate (`config.RPCCert`) is used together with the server address in establishing a connection to dcrwallet daemon.

This means that the only info we need to run `dcrcli` effectively is the Wallet RPC server address. If we're creating a config file out of necessity, then the only necessary data to request from the user is  Wallet RPC server address. Pretty much makes the whole conf creation process pointless. Why run through so much code just for a single value that we can ask the user to supply via command-line options?

It gets even exciting. The behaviour for dcrwallet is such that if a user does not set a gRPC address to use for incoming connections, `localhost:9111` or `localhost:19111` is automatically used for mainnet and testnet respectively. If we were to borrow such similar behaviour, we may not even need to worry ourselves whether or not there is a conf file or if the user provided the wallet rpc address via command-line options. If either of those possibilities occur, we use the default address for testnet or mainnet.

The only potential problem with using this default value is - what if the value isn't correct? What if the user set a different dcrwallet listen address? That problem wouldn't really solved by asking the user to specify an address instead of using the default address. The user may still specify an incorrect address. This is a whole other problem that gave birth to issue #77. This PR does not confirm if an rpc server address is valid or not. That's for a different PR in response to #77.

This PR is slowly becoming unnecessary work. I'll create a different PR very soon to remove the unused rpc username and password options from config.go; and also use a default rpc address if none is provided. When that PR gets merged, one is left asking, is there ever a scenario where we'll need to ask a user to create a `dcrcli.conf` file to continue using the app? Maybe. Maybe not.

My recommendation at this point is, we change this PR's direction. The goal should now be to create a conf file whenever necessary by running `dcrcli --init`. When a user runs that command, he should be prompted to provide values for all the available dcrcli options, or accept the default values. Then the conf file is created.

README.md
```diff
@@ -34,13 +34,14 @@ $ go install
...
+Begin by running the following command to create the configuration file:
 
 ```bash 
-cp sample-dcrcli.conf dcrcli.conf 
+dcrcli --init 
 ``` 
 
-Then edit dcrcli.conf and input your RPC settings. After you are finished, move dcrcli.conf to the `appdata` folder (default is `~/.dcrcli` on Linux, `%localappdata%\Dcrcli` on Windows). See the output of `dcrcli -h` for a list of all options
+Follow the command prompt and input your RPC settings. 
+The configuration file, dcrcli.conf will be created in your `appdata` folder (default is `~/.dcrcli` on Linux, `%localappdata%\Dcrcli` on Windows). See the output of `dcrcli -h` for a list of all options
```
barisere: What is the default `appdata` folder on Mac?

README.md
```diff
@@ -34,13 +34,14 @@ $ go install
 
 ### Create configuration file 
 
-Begin with the sample configuration file:
+Begin by running the following command to create the configuration file:
 
 ```bash 
-cp sample-dcrcli.conf dcrcli.conf 
+dcrcli --init
```
barisere: You can prepend this with a "$" to indicate that it's a command to run at the terminal.
```
$ dcrcli --init
```

README.md
itswisdomagain: Great question. That's one of the changes that the readme needs.
For Mac: `~/Library/Application Support/Dcrcli/dcrcli.conf`

config.go
```diff
@@ -39,6 +46,7 @@ type config struct {
 	HTTPServerAddress string `short:"s" long:"serveraddress" description:"Address and port of the HTTP server."`
 	HTTPMode          bool   `long:"http" description:"Run in HTTP mode."`
 	NoDaemonTLS       bool   `long:"nodaemontls" description:"Disable TLS"`
+	Init              bool   `long:"init" description:"Create the config file"`
```
itswisdomagain: PR #60 introduces a new flag (`--createwallet`) for creating wallet (if it doesn't exist already) before using dcrcli. I initially used `--init` for that use case but decided `init` was a little ambiguous and not very descriptive.

I want to suggest using a more descriptive flag here, say `--createconfig` or `--config`. `--config` might not work since it could be confused with `--configfile` used for specifying a config file path.

itswisdomagain (changes requested):

barisere:
@itswisdomagain `--init` works, for me. It follows a familiar pattern in CLIs such as `git` and `npm`. I would even prefer it as a command.
To create the configuration for running dcrcli:
```
$ dcrcli init [app_directory]
```
But that's just me being familiar with other tools that use this style.

itswisdomagain:
@barisere I think in those cases, `init` is required to use the app, so it makes sense. In our case, `init` isn't required and does not perform the function of getting the app ready for use. It's just a utility function, as the user can create the conf file without using `init`

barisere:
@ademuanthony Please address https://github.com/raedahgroup/dcrcli/pull/64#pullrequestreview-187716940.
The configuration directory should be created if it does not already exist.
```
$ ./dcrcli --init
Wallet RPC server (default = localhost:19111):
RPC username (default = rpcuser):
RPC password (default = rpcpass): error in creating config file: open /home/barisere/.dcrcli/dcrcli.conf: no such file or directory
```
Also move the error message to its own line.

barisere (changes requested):
When the creation of the configuration file succeeds, the output is as follows.
```
$ ./dcrcli --init -C ./dcrcli.conf.new
Wallet RPC server (default = localhost:19111):
RPC username (default = rpcuser):
RPC password (default = rpcpass):

The config file has been set successfully

```
The blank lines are not necessary, and the absolute path to the file written is displayed to the user.
An output like the following looks fine to me. What do you think about it?
```
$ ./dcrcli --init -C ./dcrcli.conf.new
Wallet RPC server (default = localhost:19111):
RPC username (default = rpcuser):
RPC password (default = rpcpass):
The config file has been written to /home/barisere/code/go/src/github.com/raedahgroup/dcrcli/dcrcli.conf.new
```

barisere:
I notice we're simply logging errors to the console. I suggest we log them to a file as well, and let the users know about the file. That way, the history of errors does not get lost, and we can get easily introduce telemetry if we want to. The main benefit is being able to go look at detailed error output (including stack traces if necessary) in the log file, and knowing where to start looking for problems. If a user reports an error, we can request the log file for richer context.
@ademuanthony @itswisdomagain

barisere (changes requested):
When the configuration file is created, the default `rpccert` field should be an actual path on the user's filesystem. Currently it is hard-coded.
Compare
```
 ; Specify dcrwallet's RPC certificate, or disable TLS for the
 ; rpccert=/home/me/.dcrwallet/rpc.cert
```
vs
```
; Specify dcrwallet's RPC certificate, or disable TLS for the
; rpccert=/home/barisere/.dcrwallet/rpc.cert
```

itswisdomagain:
> I notice we're simply logging errors to the console. I suggest we log them to a file as well, and let the users know about the file. That way, the history of errors does not get lost, and we can get easily introduce telemetry if we want to. The main benefit is being able to go look at detailed error output (including stack traces if necessary) in the log file, and knowing where to start looking for problems. If a user reports an error, we can request the log file for richer context.
> @ademuanthony @itswisdomagain

Yes! Yes! A new issue should be created for this and addressed in a separate PR. Will need permission from @raedah first though

README.md
itswisdomagain: Instead of saying 
> See the output of `dcrcli -h` for a list of all options"

I think a more relevant information to display here is 

> To use a different app data folder, use `dcrcli --init -C="path/to/desired/app/data/folder"`

itswisdomagain: The **Using dcrcli** section below should come above this **Configuration** section. That section (**Using dcrcli**) should mention using
- `dcrcli [options] <command> [args...]` to run a command
- `dcrcli -h` to view general program help, application options and available commands
- `dcrcli <command> -h` to view help information for a particular command

config/config.go
```diff
@@ -15,6 +17,11 @@ import (
 const (
 	defaultConfigFilename    = "dcrcli.conf"
 	defaultHTTPServerAddress = "127.0.0.1:1234"
+
+	defaultRpcAddress  = "localhost:19111"
+	defaultRpcUsername = "rpcuser"
+	defaultRpcPassword = "rpcpass"
+	ServerAddress      = "127.0.0.1:7778"
```
itswisdomagain, line 22: This isn't relevant, we already have `defaultHTTPServerAddress`

config/config.go
```diff
@@ -36,6 +44,7 @@ type Config struct {
 	HTTPServerAddress string `short:"s" long:"serveraddress" description:"Address and port of the HTTP server."`
 	HTTPMode          bool   `long:"http" description:"Run in HTTP mode."`
 	NoDaemonTLS       bool   `long:"nodaemontls" description:"Disable TLS"`
+	Init              bool   `long:"init" description:"Create the config file"`
```
itswisdomagain, line 45: In line with [this comment](https://github.com/raedahgroup/dcrcli/pull/64#issuecomment-449858448) and [this reply](https://github.com/raedahgroup/dcrcli/pull/64#issuecomment-450151078), I want to re-recommend this flag be renamed.

It's okay if git and npm uses `init`, you can't use those programs without first issuing that command, "init" makes sense.

In our case, "init" isn't a pre-requirement to use the program, so it's quite misleading... What this does _isn't_ setting up the program for use, but instead creating a config file on-demand - at any time the user says he wants to create one. I think `--config` is appropriate

config/config.go
```diff
@@ -93,3 +110,69 @@ func LoadConfig() (*Config, *flags.Parser, error) {
...
 	return &config, parser, nil
 }
+
+
+func configure(cfg Config) (err error) {
+	inputErr := func(error) error {
+		return fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	var rpcAddress, rpcUsername, rpcPassword string
```
itswisdomagain, line 135: The `rpcUsername` and `rpcPassword` values are not used in the dcrcli program. They should be completely removed from the list of config options and this function should not request values for them from the user.

itswisdomagain: Since those 2 options are no longer required, it doesn't make sense for a user to run `dcrcli --init` or `dcrcli --config` just to set `rpcAddress`. I think it'd be appropriate to prompt user to provide values for all or most of the application options. One exception would be the `ConfigFile` option (only makes sense in command-line options).

Then use the values provided to create the config file.

config/config.go
```diff
@@ -93,3 +110,69 @@ func LoadConfig() (*Config, *flags.Parser, error) {
...
+	}
+
+	config := struct {
+		RpcAddress    string
+		RpcUsername   string
+		RpcPassword   string
+		ServerAddress string
+	}{
+		RpcAddress: rpcAddress, RpcUsername: rpcUsername, RpcPassword: rpcPassword,
+		ServerAddress: ServerAddress,
```
itswisdomagain, line 170: We should not directly save a default/static value to config file. Before doing so, the user should be prompted for a value to use with the option of retaining the default value

config/dcrcli.config.go
```diff
@@ -20,8 +23,10 @@
...
 
 ; Server address to serve dcrcli via HTTP (when http=true or http=1)
-; serveraddress=127.0.0.1:7778
+serveraddress={{.ServerAddress}}
 
 ; Turn HTTP mode on (http=true or http=1) or off (http=false or http=0).
 ; HTTP mode is off by default.
-; http=true
+; http=true`
+
```
itswisdomagain, line 31: We need to remember to update this file as more application options/flags are added. For example, we now have UseTestNet (`; testnet=1`)

itswisdomagain (changes requested):

config/config.go
ademuanthony: should we remove the config flag entirely and just request the address from the user at the point of loading the config options in the program?

itswisdomagain: Remove the config/init flag? No. We've done too much work here already to throw it away.

We also should not request the address from the user. To connect to dcrwallet, the dcrwallet cert file path is also needed but we don't request that from the user either. If the user doesn't set a value in config or in command-line options, we use the default. That's currently what we're doing for the wallet rpc server address as well.

itswisdomagain:
I'm gonna close this PR @ademuanthony and @barisere. The issue this PR took to solve is outdated and irrelevant. But the work done so far, isn't completely unusable. We'll channel it into a more relevant issue.

I'll create the new issue and reference this PR. However, this PR should not be re-opened for the new issue. A new PR should be created. Most of the code changes here can and should be retained though.

I've updated the issue being fixed by this PR (#50) with a reason for closing both the issue and this PR. You can find the reason here https://github.com/raedahgroup/godcr/issues/50#issuecomment-450585613

PR #66, opened 2018-12-18
Simplify output of running ./dcrcli without a command

### LINKED ISSUE #53: simplify output of running ./dcrcli without a command

Quadriyanney:
Simplified the output of ./dcrcli based on specifications in issue #53 
closes #53


Quadriyanney:
screenshot from 2018-12-18 07-54-16

Quadriyanney:
screenshot from 2018-12-18 07-57-18

itswisdomagain:
Hi @Quadriyanney, this is the desired output according to the issue #53 

screen shot 2561-12-18 at 8 05 54 am

Yours is a little different, yeah?

Quadriyanney:
Thought the "usage" part was supposed to show first, I could remove it tho

cli/writer.go
```diff
@@ -24,18 +24,34 @@ func printResult(w *tabwriter.Writer, res *response) {
 		header += res.columns[i] + tab
 		spaceRow += " " + tab
 	}
-
 	fmt.Fprintln(w, header)
 	fmt.Fprintln(w, spaceRow)
+
+
+	var availableCommands []interface{}
```
itswisdomagain: You really should not modify this `printResult` function. It is used by other commands to display their individual outputs after successful execution of each command.

You should instead, just go ahead to print what you want, directly in the `writeHelpMessage` function and that function would also no longer need to call this `printResult` function. 

This `printResult` function formats output in a certain uniform way, which we are trying to do without, allowing different commands and actions to display output in different ways to best fit the command output, rather than following a general format. This function may go soon when all commands are able to present their own message internally, as part of their execution. Refer to [this comment on issue 54](https://github.com/raedahgroup/dcrcli/issues/54#issuecomment-447550604) for additional context.

itswisdomagain (changes requested):

cli/writer.go
Quadriyanney: Okay, I'll work on that

Quadriyanney:
screenshot from 2018-12-18 08-49-45

itswisdomagain:
Notice a discrepancy with `dcrcli -h` with these changes:
screen shot 2561-12-18 at 8 51 56 am

Previously, it looked like:
screen shot 2561-12-18 at 8 49 35 am

I'm guessing you can immediately figure out where the discrepancy is coming from and apply a fix.

While at it, also rebase your branch with `raedahgroup/dcrcli master`, and fix conflicts

Quadriyanney:
screenshot from 2018-12-19 07-15-22

cli/help.go
```diff
@@ -17,9 +18,37 @@ func HelpMessage() string {
 
 // PrintHelp outputs help message to os.Stderr
 func PrintHelp(appName string) {
-	usagePrefix := fmt.Sprintf("Usage:\n  %s ", appName)
 	stderrTabWriter := tabWriter(os.Stderr)
-	writeHelpMessage(usagePrefix, stderrTabWriter)
+	writeSimpleHelpMessage(stderrTabWriter)
```
itswisdomagain: the `appName` parameter to this function is no longer needed

cli/help.go
```diff
@@ -17,9 +18,37 @@ func HelpMessage() string {
...
+}
+
+func writeSimpleHelpMessage(w *tabwriter.Writer) {
+	var availableCommands []interface{}
+	var experimentalCommands []interface{}
```
itswisdomagain: We know what kind of data we're storing in these slices (`string`), why not use that instead of `interface{}`?

cli/help.go
```diff
@@ -17,9 +18,37 @@ func HelpMessage() string {
...
+	}
+	availableRowStr = strings.TrimSuffix(availableRowStr, ", ")
+	fmt.Fprintln(w, fmt.Sprintf(availableRowStr, availableCommands...))
+
+	experimentalRowStr := "experimental: "
+	for range experimentalCommands {
+		experimentalRowStr += "%v, "
+	}
+	experimentalRowStr = strings.TrimSuffix(experimentalRowStr, ", ")
+	fmt.Fprintln(w, fmt.Sprintf(experimentalRowStr, experimentalCommands...))
```
itswisdomagain: If you use `[]string` above, you can avoid having to loop over both slices again and just print out the desired output like so:

```
fmt.Fprintln(w, "available cmds: ", strings.Join(availableCommands, ", "))
fmt.Fprintln(w, "experimental: ", strings.Join(experimentalCommands, ", "))
```

itswisdomagain (changes requested):
This looks much more like it. Way to go!
Take a look at these few suggestions and we should be good.

cli/help.go
itswisdomagain: Just confirmed, this function (as it was) is used in other places. Don't think it's a good idea to alter it, may be better to create a different function for this use case instead.

A good example of an affected use case is when you pass an invalid option/flag to dcrcli such as `dcrcli -l`. The output for such cases is different now from what used to be.

Quadriyanney: I don't really get this

cli/help.go
Quadriyanney: Okay

Quadriyanney:
screenshot from 2018-12-19 17-13-46

itswisdomagain (approved):

PR #62, opened 2018-12-16
balance command cleanup

itswisdomagain:
fixes #49, display only total by default, show spendable if different from total

**before (web)**
screen shot 2561-12-16 at 1 59 22 am
**before (cli)**
screen shot 2561-12-16 at 1 58 45 am

**after (balance == spendable)**
**web**
screen shot 2561-12-16 at 1 04 43 am
**cli**
screen shot 2561-12-16 at 2 30 18 am

**after (balance != spendable)**
**web**
screen shot 2561-12-16 at 2 31 06 am

**cli**
screen shot 2561-12-16 at 2 31 26 am

If there are more than 1 account in wallet, cli will list balance for each account along with account name in the following fashion:
screen shot 2561-12-16 at 2 10 36 am

Issue #55, opened 2018-12-12
sendcustom more features


raedah:
```
Destination Address 2 (or blank to continue):
[repeat until done]
How many change outputs would you like to use? (default: 1): 2
[if more then one was specified]
Use random amounts for the change outputs? (Y/n): y
Enter your spending passphrase: 
You are about to spend the input
9.423 DCR from TsgUAq54uZ6GW1VQwgjtgG7KB68pJwy73dJ (default)
and send it to
2.123 DCR to TsgQwgjtgG7KB68pJwyUAq54uZ6GW1V73dJ
1.123 DCR to TsgJwyUAq54uZ6GWQwgjtgG7KB68p1V83dJ
3.123 DCR to TsgJwyWQwgjtgG7KB68pUAq54uZ6G1V83dJ (change)
3.123 DCR to TsgJwyuZ6G1V8WQwgjtgG7KB68pUAq543dJ (change)
Do you want to broadcast it? (y/n): y
sent txid b5f844895a4e5b4e496929be0356e8435a8c30976d6bd59a29539fddd37fc9d7

```

raedah:
When listing the spent outputs, show account, date, confirms.
Instead of showing the utxo id, show the Received with Address.
Sort by amount. Do not show the list be default. Have a (l)ist command.
By default auto pick the first best sized input, show it, and provide it as a default available selection.

itswisdomagain:
1 thought @raedah, the feature to send to multiple addresses can also be used in regular send and also on the web ui for custom send. Should this issue extend to those as well? For clarity,

- Replicate the flow above in the web interface version of send custom
- For the cli and web interface for the _regular_ send feature, add support for specifying multiple destination addresses (and amount for each address)

itswisdomagain:
Some observations from your initial comment:

- Supporting multiple destination addresses would require being able to specify the amount to send to each address
- If user choses to use more than 1 change output, what happens if they (1) chose to use random amounts for the change outputs (2) if they don't?
- Does (1) above spread the change equally across the change addresses, while (2) asks them to enter the amount they'd want sent to each change address?
- The change address(es) are all generated from the account specified at the start of the command, this is fine, yes?

raedah:
Yes, it is essentially one feature for control of the inputs, and another feature for control of the outputs. They can be used where appropriate.

[...]

PR #71, opened 2018-12-19
Refactor commands to use the go-flags package properly.

### LINKED ISSUE #70: Error parsing config file output
I got the following error when trying to run dcrcli with a no-longer-supported option in my config file:
screenshot 2018-12-19 at 1 01 18 pm

Extremely long. Took several tries (switching between branches) to determine what the error was. The actual error was hidden in the long message.

I'm thinking we should instead show just the error message like so:
screenshot 2018-12-19 at 1 03 29 pm
and then add "instructions"/"tip" on how to view a list of the supported options, or something along that line.

The current approach will only get longer as options are added to config

barisere:
Commands and subcommands can now be used in a simpler way.

EDIT: This also fixes #70


itswisdomagain:
Some observations from preliminary testing:
- Running `dcrcli` without a command shows a message that doesn't appear customizable. Compare #53 
- There's a funny behaviour when using an unknown option/flag. Some examples to illustrate:
- `dcrcli -n` (unknown flag) doesn't show any message or perform any action. The program just exits
- `dcrcli -n=abc` (unknown option) shows a message to specify a command, ignoring the invalid option passed
- `dcrcli send -n` does nothing, while `dcrcli send -n=abc` finally says something expected: "unknown flag `n'"

I think 1 primary trigger for this refactor is to enable us use command options, can you provide an example to that effect? Use the receive command, let's assume it accepts an "AccountNumber" option, type uint32, short "n", long "account". This option is only useful when executing the receive command. I'm eager to see how this plays out

Once we have those issues ironed out, I'll take a deeper dive into the code

barisere:
> Some observations from preliminary testing:
> 
>     * Running `dcrcli` without a command shows a message that doesn't appear customizable. Compare #53
> 
>     * There's a funny behaviour when using an unknown option/flag. Some examples to illustrate:
> 
>     ...

Can you check those again? The behaviour has changed.

> I think 1 primary trigger for this refactor is to enable us use command options, can you provide an example to that effect? Use the receive command, let's assume it accepts an "AccountNumber" option, type uint32, short "n", long "account". This option is only useful when executing the receive command. I'm eager to see how this plays out

Oops, I used the `send` command instead. Assume the `send` command has a sub command, `custom`, that only prints "Custom send called" and the address passed as a flag. The flag has a short form "-a" and a long form "--address". We create this command as follows.
```go
// file: cli/commands.go

type SendCommand struct {
	Custom CustomSend `command:"custom"`
}

type CustomSend struct {
	Address string `short:"a" long:"address"`
}

func (c CustomSend) Execute(args []string) error {
	fmt.Println("Custom send called")
	fmt.Println("address value:", c.Address)
	return nil
}
```
We register the command as follows.
```go
// file: cli/app.go

type AppCommands struct {
	Send       SendCommand       `command:"send" description:"send a transaction" subcommands-optional:"optional"`
	// other fields
}

var DcrcliCommands AppCommands
```
And we parse it as follows.
```go
// file: main.go

cli.Setup(client)
parser := flags.NewParser(&cli.DcrcliCommands, flags.Default)
if _, err := parser.Parse(); err != nil {
	// handle error
}
```
Below is help output when running the compiled code.
```
$ ./dcrcli -C ./dcrcli.conf send custom --help
Usage:
  dcrcli [OPTIONS] send custom [custom-OPTIONS]

Application Options:
  -v, --version          Display version information and exit
  -C, --configfile=      Path to configuration file
  -u, --rpcuser=         RPC username
  -p, --rpcpass=         RPC password
  -w, --walletrpcserver= Wallet RPC server to connect to
  -c, --rpccert=         RPC server certificate chain for validation
  -s, --serveraddress=   Address and port of the HTTP server.
      --http             Run in HTTP mode.
      --nodaemontls      Disable TLS

Help Options:
  -h, --help             Show this help message

[custom command options]
      -a, --address=
```
When we run the `custom` subcommand without the address flag, the output is as follows.
```
$ ./dcrcli -C ./dcrcli.conf send custom 
Custom send called
address value:
```
When we run the `custom` subcommand with the address flag set, the output is as follows.
```
$ ./dcrcli -C ./dcrcli.conf send custom --address=TsfDLrRkk9ciUuwfp2b8PawwnukYD7yAjGd
Custom send called
address value: TsfDLrRkk9ciUuwfp2b8PawwnukYD7yAjGd
```
If we use the flag without providing an argument, the following error is shown.
```
$ ./dcrcli -C ./dcrcli.conf send custom --address
expected argument for flag `-a, --address'

$ ./dcrcli -C ./dcrcli.conf send custom -a
expected argument for flag `-a, --address'
```
That's mostly it.

cli/commands.go
```diff
@@ -1,19 +1,56 @@
 package cli
 
-// command is an action that can be requested at the cli
-type command struct {
-	name        string
-	description string
-	handler     handler
+type BalanceCommand struct{}
+
+func (b BalanceCommand) Execute(args []string) error {
```
itswisdomagain: I see a lot of repetition with these Execute methods. I expect there should be a way to prevent that. I have something in mind, but want to test it out first. Might commit my changes to this branch if it pays off

main.go
```diff
@@ -69,14 +81,57 @@ func enterHttpMode(config *config) {
...
+}
+
+func loadConfig(appName string) (*cli.Config, []string, error) {
+	// load defaults first
+	cfg := cli.DefaultConfig()
+
+	// Load additional config from file
+	parser := flags.NewParser(&cfg, flags.IgnoreUnknown)
+
+	// Errors here will be owing to unknown flags and commands, which are not of interest at this point.
```
itswisdomagain: Why are they not of interest?

itswisdomagain (changes requested):
This looks like great improvement. 1 other immediate cause for concern is this:
- `config.go` is not restricted to just cli-related usages, it shouldn't belong in the `cli` package

I have an idea, want to try it out on your branch and see how it goes

main.go
itswisdomagain: I think I get why now

main.go
```diff
@@ -41,42 +43,111 @@ func (v *Version) String() string {
...
 
-func enterCliMode(appName string, config *config, args []string) {
-	client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
-	if err != nil {
-		fmt.Fprintln(os.Stderr, "Error connecting to RPC server")
-		fmt.Fprintln(os.Stderr, err.Error())
+func enterCliMode(config *core.Config, client *walletrpcclient.Client) {
+	cli.Setup(client)
+	parser := flags.NewParser(&cli.DcrcliCommands, flags.Default)
+	if _, err := parser.Parse(); err != nil {
```
itswisdomagain, line 47: When no command is passed, this produces the following output:
`Please specify one command of: balance, history, receive, send or send-custom`

The current output (based on #53 and #66) is:
```
available cmds:  balance, send, receive, history
experimental:  send-custom
```

The latter should be preferred.

main.go
```diff
@@ -41,42 +43,111 @@ func (v *Version) String() string {
...
+	return false
+}
+
+func handleParseError(err error, parser *flags.Parser) {
+	if (parser.Options & flags.PrintErrors) != flags.None {
+		// error printing is already handled by go-flags.
+		return
+	}
+	if e, ok := err.(*flags.Error); ok && e.Type == flags.ErrHelp {
+		parser.WriteHelp(os.Stderr)
```
itswisdomagain, line 81: Not critical but if this can be improved, we should. Compare dcrcli current help output to that of this PR:

dcrcli currently:
screenshot 2018-12-23 at 2 05 20 pm

with this PR:
screenshot 2018-12-23 at 2 05 31 pm

Notice the presense of `[<args...>]` in the usage text that's no longer included with this PR.

Also notice that commands are currently displayed before options, I think it's more user-friendly that way - showing commands before options, but I may be wrong. When I run the help command of any app, it's rarely because I want to see the app options that I can configure but because I want to see which of the supported commands I need at the moment to achieve something. More often than not, I'm looking for a command, not an option

itswisdomagain: Still on help message displays, this PR does a neat job of introducing command-specific help with the `dcrcli <command> -h` syntax. But the message displayed is too verbose, I believe. Can we make it more concise? It should actually provide "details" about how the command functions and the optional arguments and flags it works with.

screenshot 2018-12-23 at 2 24 03 pm

In a bid to achieve similar objectives set forth in issue #51, PR #63 has begun heading in the desired direction I believe. [This comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448585542) and [this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448952265) mention possible ways to achieve the objective of displaying relevant messages for command-specific help requests.

By the way, that PR is on pause because it expects this PR will somewhat make it easier to achieve the goal. Can it? Will it?

main.go
```diff
@@ -41,42 +43,111 @@ func (v *Version) String() string {
...
+
+func loadConfig() (*core.Config, *flags.Parser, error) {
+	// load defaults first
+	cfg := core.DefaultConfig()
+	commands := cli.DcrcliCommands
+	commands.Config = cfg
 
-	c := cli.New(client, appName)
-	c.RunCommand(args)
+	parser := flags.NewParser(&commands, flags.HelpFlag)
```
itswisdomagain: This is potentially problematic. Regardless of whether the user wants to run in web or cli mode, this code parses command-line options assuming we're running on cli. The following scenario highlights how this could be problematic:

Say the cli balance command uses a flag (`-e/--extended` | compare #62), and a user runs the following:
Scenario 1: `dcrcli send -e` -> error, unknown flag `e`
Scenario 2: `dcrcli balance -e` -> valid, `e` is a valid flag for the `balance` **cli** command
Scenario 3: `dcrcli balance -e --http` -> valid, because `e` is a valid flag for the `balance` command but this should ideally fail because `e` is an invalid flag for any command in http mode. http mode does not even support cli commands by the way.

The ideal behaviour for the Scenario 3 would have been to return an error that the flag `e` is unsupported in http mode. If this parser was created with just `&cfg` and not `&commands`, this flag would have been rejected

itswisdomagain (changes requested):
Feedback from testing, haven't really paid another attention to the code. I believe this PR comes with changes to the behaviour of running `dcrcli` (i.e. without commands), `dcrcli -h` and `dcrcli <command> -h`. Those were scenarios I tested for in getting the feedback below.

A fourth use case I simulated in my head but I know this PR also introduces changes regarding, is command-specific options. According to my simulation, this should work well (for the most part). The exception, a pointed below, is when using a command-specific option when running in cli mode. Since commands are limited to cli mode, command-specific options should throw an error when parsing if the user is running on http mode.

I believe the the identified issues below are fixable. An adjusted version of this PR demonstrates potential ways of fixing some of the identified issues. Reference links are provided in such cases.

main.go
itswisdomagain: Did something here that might help. Might.
https://github.com/itswisdomagain/dcrcli/blob/pr71-update/main.go#L32

main.go
itswisdomagain: Parsing twice is one work-around. First parse is as basic as possible, does not include cli commands and command-specific options. This parse too would not throw error if it encounters an invalid option because such option may actually be a valid command-specific option if we're running in cli.
Compare https://github.com/itswisdomagain/dcrcli/blob/pr71-update/main.go#L14

First parse helps us know what mode we're running on, so second parse can be more thorough, including commands and command-specific options if running in cli and the reverse if not.
Compare https://github.com/itswisdomagain/dcrcli/blob/pr71-update/main.go#L39 and https://github.com/itswisdomagain/dcrcli/blob/pr71-update/main.go#L50

itswisdomagain:
We'll need to update the **Using dcrcli** section of the readme

main.go
barisere: The closest I could get is this.
```
$ ./dcrcli -C dcrcli.conf 
Available Commands:  balance, beta, history, receive, send, send-custom
```
I could not access the command data struct, so the experimental flag could not be used.

main.go
barisere: > Not critical but if this can be improved, we should. Compare dcrcli current help output to that of this PR:
> 
> dcrcli currently:
> screenshot 2018-12-23 at 2 05 20 pm
> 
> with this PR:
> screenshot 2018-12-23 at 2 05 31 pm
> 
> Notice the presense of `[<args...>]` in the usage text that's no longer included with this PR.

If the command has positional arguments, then [args] or <args> is displayed, depending on whether the arguments are required or not.
```
$ ./dcrcli -C dcrcli.conf receive -h
Usage:
  dcrcli [OPTIONS] receive [account]

Application Options:
  -v, --version          Display version information and exit
  -C, --configfile=      Path to configuration file
  -u, --rpcuser=         RPC username
  -p, --rpcpass=         RPC password
  -w, --walletrpcserver= Wallet RPC server to connect to
  -c, --rpccert=         RPC server certificate chain for validation
  -s, --serveraddress=   Address and port of the HTTP server.
      --http             Run in HTTP mode.
      --nodaemontls      Disable TLS

Help Options:
  -h, --help             Show this help message
```

> Also notice that commands are currently displayed before options, I think it's more user-friendly that way - showing commands before options, but I may be wrong. When I run the help command of any app, it's rarely because I want to see the app options that I can configure but because I want to see which of the supported commands I need at the moment to achieve something. More often than not, I'm looking for a command, not an option

I couldn't find a way around this. It's not quite configurable with `go-flags`.

main.go
barisere: `--http` is a switch. Because it is also a top-level option, we encounter it immediately after the first parse. So, the program switches to web mode if the `--http` flag is encountered; every other command is ignored.
The current order of priority is as follows.
- -v displays the version information and exits. Every other flag or command is ignored.
- -h/--help
- --http

main.go
itswisdomagain: This is close...

I just remembered a suggestion/request from @raedah that might help. Making reference to the output of `dcrcli -h`, here's @raedah's comment:

> send-custom cmd can be separated from the basic cmds in a more advanced sections, ..so we are creating categories

To expatiate, the idea is to separate commands into category, such that when displaying `dcrcli -h`, commands such as `send-custom` would not appear under the **Available commands** section but under a new section **Advanced commands**

If we're able to achieve separating the commands into categories/sections, I think this PR will be able to achieve the desired output for running `dcrcli` without any command

main.go
itswisdomagain: All good. Getting the first part to work is priceless

itswisdomagain:
For reference purposes, here are the objectives of (or issues affected by) this PR:

- command-specific options
**Reference issue**: #49, also [this comment](https://github.com/raedahgroup/dcrcli/pull/63#discussion_r242304719)
**Status**: Green (feature fully supported by this PR)
**More info**: Once this PR is merged, work on #62 can continue

- experimental commands distinction when running dcrcli without commands
**Reference issue**: #53
**Status**: Yellow (partly supported)
**More info**: Requirement is fully implemented in #66,  but works quite differently in this PR
**Way forward**: Leave as is. A new issue ([described here](https://github.com/raedahgroup/dcrcli/pull/71#discussion_r)) should be created and fixed in a different PR. Fixing that issue would make it possible for this requirement to be fully implemented.

- command-specific help
**Reference issue**: #51, also [this comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448585542) and [this other comment](https://github.com/raedahgroup/dcrcli/pull/63#issuecomment-448952265)
**Status**: Red (support for this feature/requirement is not solid enough)
**More info**: This PR does a neat job of introducing command-specific help with the `dcrcli <command> -h` syntax. But the message displayed is too verbose. It should be more concise, providing "details" about how the command functions and the optional arguments and flags it works with. Similar work being done in #63 cannot continue with what we have at the moment
**Way forward**: Make all commands have long description tags in `app.go`. #63 then uses a parser to get the long description for each command and displays that when `dcrcli help <command>` is issued. `dcrcli help` in #63 should produce the same output as `dcrcli -h` in this PR. Remove support for command-specific help via `dcrcli <command> -h` in this PR or have `dcrcli <command> -h` display what it currently displays AND include instruction to use `dcrcli help <command>` to get detailed help for the command. For this PR, `dcrcli help <command>` may just print a dummy message awaiting complete implementation in #63

_PS: I haven't reviewed the code yet, this "overview" may take a different outlook after I review the code_

EDIT (after code review)
Regarding the way forward for **command-specific help**, let's leave the code as is, and allow #63 take care of that. I think there's somewhat enough base work to build on, following the idea of adding long description tags to the commands or by implementing Usage interface on each command. A number of options we can explore, let's get this PR merged first and we'll deal with that

cli/app.go
```diff
@@ -0,0 +1,16 @@
+package cli
+
+import (
+	"github.com/raedahgroup/dcrcli/cli/core"
+)
+
+type AppCommands struct {
+	core.Config
+	Balance    BalanceCommand    `command:"balance" description:"show your balance"`
+	Send       SendCommand       `command:"send" description:"send a transaction" subcommands-optional:"optional"`
```
itswisdomagain: `subcommands-optional` tag not necessary, must be left-over code from an experiment with subcommands...

cli/app.go
```diff
@@ -0,0 +1,16 @@
...
+	SendCustom SendCustomCommand `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
+	Receive    ReceiveCommand    `command:"receive" description:"show your address to receive funds"`
+	History    HistoryCommand    `command:"history" description:"show your transaction history"`
+}
+
+var DcrcliCommands AppCommands
```
itswisdomagain: I see this is only used in line 87 of `main.go`, you can use `&cli.AppCommands{}` in lieu of this variable, rendering this unnecessary.

Also, I think the contents of this file would make more sense in `commands.go`, preferably at the top, and then keep the individual command structs  at the bottom

I also recommend moving the Execute methods to `handlers.go`. So that instead of having different handler functions for each command and calling these functions in Execute, we can perform all command execution in the `Execute` method of the command struct. This also has the added advantage of making any command-specific options readily accessible in the handler function (aka Execute)

cli/cli.go
```diff
@@ -1,96 +1,22 @@
...
-		funcMap:         make(map[string]handler),
-		walletrpcclient: walletrpcclient,
-		appName:         appName,
-	}
-
-	// register handlers
-	client.registerHandlers()
-
-	return client
+type response struct {
```
itswisdomagain: Permit me to saddle you with this requirement. This type should go. All commands should handle displaying appropriate responses after successful execution. Compare [this comment](https://github.com/raedahgroup/dcrcli/issues/54#issuecomment-447550604). You may not want to stress too much about how to display the successful output for different commands, just print what is currently printed. Other PRs building off related issues will revise the successful message printing for each command. Examples are #49, #54 and #55

As provided for by the signature of `Execute` in `goflags.Commander`, commands should only return an error which is printed to the terminal.

cli/commands.go
```diff
@@ -1,19 +1,60 @@
...
+		return err
+	}
+	printResult(stdoutWriter, res)
+	return nil
+}
+
+type SendCommand struct{}
+
+func (s SendCommand) Execute(args []string) error {
+	res, err := normalSend(walletClient, args)
```
itswisdomagain: As mentioned in an earlier comment, the code in `normalSend` should come here. No need to call another handler function, this particular pattern is repeated across all Execute methods in this file, causing avoidable repetition.

Also, assuming `SendCommand` supports some command-specific options, we can have access to such option here seamlessly (`s.option`). Passing such options to the `normalSend` handler will be another stress that can be avoided.

itswisdomagain: This comment applies to all `Execute` methods in this file. Also, it might make sense to put these in `handlers.go`, leaving `commands.go` as a file that describes the supported commands and `handler.go` as the file that actually perform the actions for each command.

cli/core/config.go
```diff
@@ -0,0 +1,50 @@
...
+	defaultLogFilename       = "dcrcli.log"
+	defaultHTTPServerAddress = "127.0.0.1:1234"
+)
+
+var (
+	defaultAppDataDir          = dcrutil.AppDataDir("dcrcli", false)
+	defaultDcrwalletAppDataDir = dcrutil.AppDataDir("dcrwallet", false)
+	defaultRPCCertFile         = filepath.Join(defaultDcrwalletAppDataDir, "rpc.cert")
+	defaultConfigFile          = filepath.Join(defaultAppDataDir, defaultConfigFilename)
+	defaultLogDir              = filepath.Join(defaultAppDataDir, defaultLogDirname)
```
itswisdomagain: `defaultLogDir`, `defaultLogFilename` and `defaultLogDirname` have no use and can be deleted

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
+	os.Exit(1)
+}
+
+func supportedCommands(parser *flags.Parser) []string {
+	registeredCommands := parser.Commands()
+	commands := make([]string, 0, len(registeredCommands))
+	for _, command := range registeredCommands {
+		commands = append(commands, command.Name)
+	}
+	sort.Strings(commands)
```
itswisdomagain: Don't see a need for sort. The commands are listed in the order in which they're defined in the `AppCommands` struct. If there's a particular order of listing commands we prefer, it should be reflected there

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
+	}
 
-	if config.HTTPMode {
-		fmt.Println("Running in http mode")
-		enterHttpMode(config)
-	} else {
-		enterCliMode(appName, config, args)
+	// Show version and exit if the version flag was specified
+	if config.ShowVersion {
+		fmt.Println(appVersion)
```
itswisdomagain: When you call `loadConfig` above, a check is made for the version flag and handled accordingly, this is never true and this code never runs. If you agree this code can go safely, then `appVersion` will have just one usage (in the `loadConfig` function), it might no longer make sense to keep the value in a global variable

cli/core/config.go
```diff
@@ -0,0 +1,50 @@
+package core
```
itswisdomagain: This file does not belong in `cli`, it applies to the entire app, not just `cli` interface. We can move the `core` package to the root of the project without causing issues

cli/core/config.go
```diff
@@ -0,0 +1,50 @@
...
+
+func DefaultConfig() Config {
+	return Config{
+		ConfigFile:        defaultConfigFile,
+		RPCCert:           defaultRPCCertFile,
+		HTTPServerAddress: defaultHTTPServerAddress,
+	}
+}
+
+func AppName() string {
```
itswisdomagain: Love having this here. `AppName()` is not a cli-only thing, it should be accessible to any interface package and should not reside in the `cli` package

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
+	}
+}
+
+func enterHTTPMode(config *core.Config, client *walletrpcclient.Client) {
 	web.StartHttpServer(config.HTTPServerAddress, client)
 }
 
-func enterCliMode(appName string, config *config, args []string) {
-	client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
+func enterCliMode(config *core.Config, client *walletrpcclient.Client) {
```
itswisdomagain: `config` is not used here

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
+	cli.Setup(client)
+	parser := flags.NewParser(&cli.DcrcliCommands, flags.Default&(^flags.PrintErrors))
```
itswisdomagain: The flag options here is equivalent to `flags.HelpFlag | flags.PassDoubleDash` (which I think is more readable). Compare https://godoc.org/github.com/jessevdk/go-flags#Default

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
-func enterHttpMode(config *config) {
 	client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "Error connecting to RPC server")
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
 	}
 
+	if config.HTTPMode {
+		fmt.Println("Running in http mode")
```
itswisdomagain: Maybe move this to the `enterHTTPMode` function? That function currently has only 1 line of code, makes you _almost_ wonder why that single line of code isn't just use directly here

main.go
```diff
@@ -41,42 +45,120 @@ func (v *Version) String() string {
...
+	registeredCommands := parser.Commands()
+	commands := make([]string, 0, len(registeredCommands))
+	for _, command := range registeredCommands {
+		commands = append(commands, command.Name)
+	}
+	sort.Strings(commands)
+	return commands
+}
+
+func loadConfig() (*core.Config, *flags.Parser, error) {
```
itswisdomagain: Can we put these and the related functions in `config.go` instead? And invoke the needed functions using `core.Fn`?

itswisdomagain (changes requested):
Got around to reviewing the code. Some thoughts...

cli/app.go
barisere: Yes, a leftover. Removed.

cli/cli.go
barisere: I think this should be done in a different PR. Take the `balance` handler: #49 will change the way the output is displayed. Same for the others. Since they're working on the output messages also, we can let them do it, or do it ourselves in a separate PR. Doing it now will require that I change the output myself (even if I do not handle formatting, I get to unnecessarily mess with what they're doing).

itswisdomagain: Okay

cli/commands/app.go
```diff
@@ -0,0 +1,12 @@
+package commands
```
itswisdomagain: I love this file/folder structure! I think `commands.go` will be a better name for this file.

cli/commands/history.go
```diff
@@ -0,0 +1,49 @@
...
+func (h HistoryCommand) Execute(_ []string) error {
+	res, err := transactionHistory(cli.WalletClient)
+	if err != nil {
+		return err
+	}
+	cli.PrintResult(cli.StdoutWriter, res)
+	return nil
+}
+
+func transactionHistory(walletrpcclient *walletrpcclient.Client) (*cli.Response, error) {
```
itswisdomagain: Do we need this function? It's pretty much straight forward like in `balance.go`. Usually the heavyweight lifting is done in the `walletrpcclient` package and these `Execute` functions just call the appropriate functions in `walletrpcclient` and display the result. By the time we do away with the `cli.Response` struct, the code here will even be much more simpler, just call `transactions, err := walletrpcclient.GetTransactions()` and then print the result to terminal without creating a `response` object

cli/commands/balance.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"github.com/raedahgroup/dcrcli/cli"
+)
+
+// BalanceCommand displays the user's account balance.
+type BalanceCommand struct{}
+
+// Execute runs the `balance` command, displaying the user's account balance.
+func (b BalanceCommand) Execute(args []string) error {
+	walletrpcclient := cli.WalletClient
```
itswisdomagain: Why create a variable for this? It's only used in the next line

cli/commands/history.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"github.com/raedahgroup/dcrcli/cli"
+	"github.com/raedahgroup/dcrcli/walletrpcclient"
+)
+
+// HistoryCommand enables the user view their transaction history.
+type HistoryCommand struct{}
+
+// Execute runs the `history` command.
+func (h HistoryCommand) Execute(_ []string) error {
```
itswisdomagain: In `Execute()` of `balance.go`, `args` is not used but the variable name was maintained. I think I like that as it gives some meaning to the function. If we however want to use `_` for cases where `args` isn't used, we should make it uniform for all command `Execute` methods. I prefer keeping `args` though, but purely a personal preference

cli/commands/receive.go
```diff
@@ -0,0 +1,65 @@
...
+
+// ReceiveCommand generates and address for a user to receive DCR.
+type ReceiveCommand struct {
+	Args struct {
+		Account string `positional-arg-name:"account"`
+	} `positional-args:"yes"`
+}
+
+// Execute runs the `receive` command.
+func (r ReceiveCommand) Execute(_ []string) error {
```
itswisdomagain: Seeing as we can define command arguments (both optional and required) in the command struct, will we ever need the `args` parameter of the `Execute` function?

cli/commands/receive.go
```diff
@@ -0,0 +1,65 @@
...
+	var accountNumber uint32
+	walletrpcclient := cli.WalletClient
```
itswisdomagain: Also do not see a need for this variable creation. We can surely just use `cli.WalletClient` directly

config/config.go
```diff
@@ -0,0 +1,89 @@
+package config
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/jessevdk/go-flags"
+
```
itswisdomagain: are these spaces necessary?

main.go
```diff
@@ -3,80 +3,83 @@ package main
 import (
 	"fmt"
 	"os"
-	"path/filepath"
+	"sort"
 	"strings"
 
+	"github.com/raedahgroup/dcrcli/cli/commands"
+
```
itswisdomagain: The spaces here, needed?

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
-	return fmt.Sprintf("%d.%d.%d%s",
-		v.Major, v.Minor, v.Patch, hashStr)
-}
-
 func main() {
-	appName := filepath.Base(os.Args[0])
-	appName = strings.TrimSuffix(appName, filepath.Ext(appName))
-
-	config, args, err := loadConfig(appName)
+	config, parser, err := config.LoadConfig()
```
itswisdomagain: Variable 'config' collides with imported package name

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
+		handleParseError(err, parser)
 		os.Exit(1)
 	}
-
-	if config.HTTPMode {
-		fmt.Println("Running in http mode")
-		enterHttpMode(config)
-	} else {
-		enterCliMode(appName, config, args)
+	if config == nil {
```
itswisdomagain: There's never a scenario where `err == nil` and `config == nil`. If `config == nil`, then `err != nil`. If `err != nil`, then this code will never be executed

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
-func enterHttpMode(config *config) {
 	client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
 	if err != nil {
 		fmt.Fprintln(os.Stderr, "Error connecting to RPC server")
 		fmt.Fprintln(os.Stderr, err.Error())
 		os.Exit(1)
 	}
 
+	if config.HTTPMode {
+		enterHTTPMode(config, client)
```
itswisdomagain: What's actually required by `enterHTTPMode` is the string value of `config.HTTPServerAddress`.  Maybe pass that only instead of the entire config?

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
-	if err != nil {
-		fmt.Fprintln(os.Stderr, "Error connecting to RPC server")
-		fmt.Fprintln(os.Stderr, err.Error())
-		os.Exit(1)
+func enterCliMode(client *walletrpcclient.Client) {
+	cli.WalletClient = client
+	parser := flags.NewParser(&commands.AppCommands{}, flags.HelpFlag|flags.PassDoubleDash)
+	_, err := parser.Parse()
+	if cli.IsFlagErrorType(err, flags.ErrCommandRequired) {
+		commands := supportedCommands(parser)
```
itswisdomagain: Variable 'commands' collides with imported package name

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
 	web.StartHttpServer(config.HTTPServerAddress, client)
 }
 
-func enterCliMode(appName string, config *config, args []string) {
-	client, err := walletrpcclient.New(config.WalletRPCServer, config.RPCCert, config.NoDaemonTLS)
-	if err != nil {
-		fmt.Fprintln(os.Stderr, "Error connecting to RPC server")
-		fmt.Fprintln(os.Stderr, err.Error())
-		os.Exit(1)
+func enterCliMode(client *walletrpcclient.Client) {
```
itswisdomagain: This function could use some spacing and comments (especially as it is not immediately obvious how the cli interface processes and handles commands)

main.go
```diff
@@ -3,80 +3,83 @@ package main
...
+	cli.WalletClient = client
+	parser := flags.NewParser(&commands.AppCommands{}, flags.HelpFlag|flags.PassDoubleDash)
+	_, err := parser.Parse()
+	if cli.IsFlagErrorType(err, flags.ErrCommandRequired) {
+		commands := supportedCommands(parser)
+		fmt.Fprintln(os.Stderr, "Available Commands: ", strings.Join(commands, ", "))
```
itswisdomagain: Can we make `supportedCommands` return the joined string instead? The function isn't really used anywhere else so we can modify it to return what we actually need rather than it returning something and then we perform some other operation on the returned value.

Alternatively, we can just modify the `supportedCommands` function to print out the message and return nothing. The command can also be renamed accordingly.

cli/cli.go
```diff
@@ -1,96 +1,31 @@
...
+	StdoutWriter = tabWriter(os.Stdout)
+)
 
-	command := commandArgs[0]
-	if !c.isCommandSupported(command) {
-		c.invalidCommandReceived(command)
-		os.Exit(1)
+// IsFlagErrorType determines whether a given error is of a given flags.ErrorType.
+// It is safe to call IsFlagErrorType with err = nil.
+func IsFlagErrorType(err error, errorType flags.ErrorType) bool {
```
itswisdomagain: I think this function belongs in `config.go`. It is not only called by cli-related operations

walletrpcclient/walletrpcclient.go
```diff
@@ -314,6 +314,19 @@ func (c *Client) NextAccount(accountName string, passphrase string) (uint32, err
 	return r.AccountNumber, nil
 }
 
+func (c *Client) AccountNumber (accountName string) (uint32, error) {
```
itswisdomagain, line 317: I'm wondering why I'm seeing this as a change seeing as it is already merged into master. Did you rebase?

itswisdomagain (changes requested):
We're making progress. Just some code cleanups

walletrpcclient/walletrpcclient.go
barisere: Yes.

cli/commands/receive.go
barisere: We may never need it, but I think keeping it is fine.

main.go
barisere: Supported commands are a list of commands, so a `[]string` is appropriate.

itswisdomagain: The motivation for this requirement is that `supportedCommands` isn't used or needed anywhere else. However, the function would likely find use soon (in #63) so let's leave it as is

barisere:
@itswisdomagain The help message for command-specific help has been shortened. Here's what it looks like now.
```
$ ./dcrcli -C dcrcli.conf send -h
Usage:
  dcrcli [OPTIONS] send
```
```
$ ./dcrcli -C dcrcli.conf receive -h
Usage:
  dcrcli [OPTIONS] receive [account]
```
If that is acceptable, then command-specific help has been covered as required, leaving us with experimental commands to grok.

itswisdomagain:
> @itswisdomagain The help message for command-specific help has been shortened. Here's what it looks like now.
> 
> ```
> $ ./dcrcli -C dcrcli.conf send -h
> Usage:
>   dcrcli [OPTIONS] send
> ```
> ```
> $ ./dcrcli -C dcrcli.conf receive -h
> Usage:
>   dcrcli [OPTIONS] receive [account]
> ```

To a very large extent, it does. The associated issue #51 recommended creating a new command `dcrcli help` to do what `dcrcli -h` does and `dcrcli help <command>` to do what `dcrcli <command> -h` does now.

Let's retain this, it's a welcome change. We might need to clean it up (I'm thinking, add a message below the command usage text that says how to see the options, something like `use dcrcli -h to view command options`), but we'll have to discuss this with @raedah and then PR #63 can continue from there.

main.go
```diff
@@ -74,16 +74,23 @@ func handleParseError(err error, parser *flags.Parser) {
...
-			helpParser.Name = parser.Name
-			helpParser.Active = parser.Active
-			helpParser.WriteHelp(os.Stderr)
-		}
-	} else {
+	if !config.IsFlagErrorType(err, flags.ErrHelp) {
 		fmt.Println(err)
+		return
+	}
+	if parser.Active == nil {
```
itswisdomagain: If you use `else if` here, the return in the preceding `if` block would be unnecessary

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #72, opened 2018-12-20
Updates send and receive outputs

### LINKED ISSUE #54: send and receive cmd output

Jujhar:
Regarding the smaller QR size unless you want to display PNG in the command line via importing another PNG rendering library the size cannot be changeable pretty much any way as if you select a pixel on the output you will see it as being one unit of black or white. API does not support resizing this type of textual character output. (practically it only supports PNG and Image size change which is not what we are outputting) (Ref https://godoc.org/github.com/skip2/go-qrcode)

Fixes #54

### DIFF
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -66,7 +66,11 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 		}
 
 		if !isValid {
-			return errors.New("invalid address")
+			if (address == "") {
+				return errors.New("You did not specify an address.")
+			}
+
+			return errors.New("That is not a valid address. Try again.")
 		}
 		return nil
 	}
@@ -88,6 +92,10 @@ func getSendAmount() (float64, error) {
 	validateAmount := func(input string) error {
 		amount, err = strconv.ParseFloat(input, 64)
 		if err != nil {
+			if (input == "") {
+				return errors.New("That is did a valid amount. Try again.")
+			}
+
 			return fmt.Errorf("error parsing amount: %s", err.Error())
 		}
 		return nil
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -60,16 +60,16 @@ func selectAccount(wallet walletcore.Wallet) (uint32, error) {
 // getSendDestinationAddress fetches the destination address to send DCRs to from the user.
 func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 	validateAddressInput := func(address string) error {
-		isValid, err := wallet.ValidateAddress(address)
+		if (address == "") {
+			return errors.New("You did not specify an address. Try again.")
+		}
+		
+		isValid, err := c.IsAddressValid(address)
 		if err != nil {
 			return fmt.Errorf("error validating address: %s", err.Error())
 		}
 
 		if !isValid {
-			if (address == "") {
-				return errors.New("You did not specify an address. Try again.")
-			}
-
 			return errors.New("That is not a valid address. Try again.")
 		}
 		return nil
@@ -90,12 +90,12 @@ func getSendAmount() (float64, error) {
 	var err error
 
 	validateAmount := func(input string) error {
+		if (input == "") {
+			return errors.New("That is did a valid amount. Try again.")
+		}
+
 		amount, err = strconv.ParseFloat(input, 64)
 		if err != nil {
-			if (input == "") {
-				return errors.New("That is did a valid amount. Try again.")
-			}
-
 			return fmt.Errorf("error parsing amount: %s", err.Error())
 		}
 		return nil
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -67,7 +67,7 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 
 		if !isValid {
 			if (address == "") {
-				return errors.New("You did not specify an address.")
+				return errors.New("You did not specify an address. Try again.")
 			}
 
 			return errors.New("That is not a valid address. Try again.")
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -60,11 +60,11 @@ func selectAccount(wallet walletcore.Wallet) (uint32, error) {
 // getSendDestinationAddress fetches the destination address to send DCRs to from the user.
 func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 	validateAddressInput := func(address string) error {
-		if (address == "") {
+		if address == "" {
 			return errors.New("You did not specify an address. Try again.")
 		}
 
-		isValid, err := c.IsAddressValid(address)
+		isValid, err := wallet.ValidateAddress(address)
 		if err != nil {
 			return fmt.Errorf("error validating address: %s", err.Error())
 		}
@@ -90,8 +90,8 @@ func getSendAmount() (float64, error) {
 	var err error
 
 	validateAmount := func(input string) error {
-		if (input == "") {
-			return errors.New("That is did a valid amount. Try again.")
+		if input == "" {
+			return errors.New("You did not specify an amount. Try again.")
 		}
 
 		amount, err = strconv.ParseFloat(input, 64)
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -63,7 +63,7 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 		if (address == "") {
 			return errors.New("You did not specify an address. Try again.")
 		}
-		
+
 		isValid, err := c.IsAddressValid(address)
 		if err != nil {
 			return fmt.Errorf("error validating address: %s", err.Error())
--- a/cli/commands/helpers.go
+++ b/cli/commands/helpers.go
@@ -96,7 +96,7 @@ func getSendAmount() (float64, error) {
 
 		amount, err = strconv.ParseFloat(input, 64)
 		if err != nil {
-			return fmt.Errorf("error parsing amount: %s", err.Error())
+			return fmt.Errorf("Invalid amount. Try again")
 		}
 		return nil
 	}
--- a/cli/commands/receive.go
+++ b/cli/commands/receive.go
@@ -61,7 +61,7 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
 		}
 	}
 
-	confirm, err := terminalprompt.RequestInput("Would you like to a generate QR code? (y/N) ", validateConfirm)
+	confirm, err := terminalprompt.RequestInput("Would you like to generate a QR code? (y/N) ", validateConfirm)
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error reading your response: %s", err.Error())
 		return err
--- a/cli/commands/receive.go
+++ b/cli/commands/receive.go
@@ -5,7 +5,8 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/raedahgroup/godcr/walletrpcclient"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/cli/runner"
 	"github.com/raedahgroup/godcr/cli/termio/terminalprompt"
 	"github.com/mdp/qrterminal"
 )
@@ -46,7 +47,7 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
 	}
 
 	// Print out address as string
-	fmt.Println(receiveResult.Address)
+	fmt.Println(receiveAddress)
 
 	// Print out QR code
 	validateConfirm := func(address string) error {
@@ -60,6 +61,7 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
 
 	if confirm == "y" {
 		qrterminal.GenerateHalfBlock("https://github.com/mdp/qrterminal", qrterminal.L, os.Stdout)
+
 	}
 
 	return nil
--- a/cli/commands/receive.go
+++ b/cli/commands/receive.go
@@ -51,10 +51,21 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
 	fmt.Println(receiveAddress)
 
 	// Print out QR code
-	validateConfirm := func(address string) error {
-		return nil
+	validateConfirm := func(userResponse string) error {
+		userResponse = strings.TrimSpace(userResponse)
+		userResponse = strings.Trim(userResponse, `"`)
+		if userResponse == "" || strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
+			return nil
+		} else {
+			return fmt.Errorf("invalid option, try again")
+		}
 	}
-	confirm, _ := terminalprompt.RequestInput("Would you like to a generate QR code? (y/N) ", validateConfirm)
+
+	confirm, err := terminalprompt.RequestInput("Would you like to a generate QR code? (y/N) ", validateConfirm)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading your response: %s", err.Error())
+		return err
+	}
 
 	if strings.EqualFold(confirm, "yes") || strings.EqualFold(confirm, "y") {
 		qrterminal.GenerateHalfBlock("https://github.com/mdp/qrterminal", qrterminal.L, os.Stdout)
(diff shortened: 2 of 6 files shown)


itswisdomagain:
Is there another qr code library that does what we want - print smaller images of qr code to terminal?

cli/handlers.go
```diff
@@ -94,6 +97,19 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
...
 	}
 
+	if err != nil {
+		return nil, err
+	}
+
+	fmt.Println("You are about to send %f DCR to %s", sendAmount, destinationAddress)
+	reader := bufio.NewReader(os.Stdin)
+  fmt.Print("Are you sure? (y/n) ")
+  text, _ := reader.ReadString('\n')
```
itswisdomagain: there's a utility pacakge `terminalprompt` that handles getting user input and performing validation on the input. For example, in this case, anything that's not y, yes, n or no should be regarded as invalid

Your current code doesn't do so, it treats anything not "n" as "yes", not safe

cli/handlers.go
```diff
@@ -94,6 +97,19 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
...
+
+	if (strings.TrimRight(text, "\n") == "n") {
+		return &response{}, nil
```
itswisdomagain: what happens if you return nil instead of &response{}?

cli/handlers.go
```diff
@@ -102,24 +118,8 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
...
-		result: [][]interface{}{
-			[]interface{}{
-				"The transaction was published successfully",
-				result.TransactionHash,
-			},
-		},
-	}
-
-	return res, nil
+	fmt.Println("sent txid %s", result.TransactionHash)
```
itswisdomagain: I'm thinking this may read better:
`sent. txid: %s`

cli/handlers.go
```diff
@@ -154,19 +154,9 @@ func receive(walletrpcclient *rpcclient.Client, commandArgs []string) (*response
...
-		result: [][]interface{}{
-			[]interface{}{
-				r.Address,
-				qr.ToString(true),
-			},
-		},
-	}
-	return res, nil
+	res := r.Address + "\n" + qr.ToString(true)
+	fmt.Println(res)
```
itswisdomagain: You could as well do `fmt.Printf("%s\n%s\n", r.Address, qr.ToString(true))`. It's one-liner but can't say if it's more readable. Leaving that to your discretion...

itswisdomagain (changes requested):
2 other observations aside from those pointed underneath:
- there are still "blank lines in the output when re-asking the question"
- if an information is not specified (say an address), it may be better to exit the command; but if re-requesting for the input, it would make sense to say "try again" at the end of the invalid input message

screenshot 2018-12-20 at 12 20 00 pm

cli/handlers.go
Jujhar: Compiles but runtime error

panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x8a57da]

Jujhar:
Changed to terminalprompt and found a marginlly smaller QR output as this one has no outside padding.

Not sure how to remove those blank lines any ideas?
 A part of me feels like those are fine and Radeah was talking about double extra lines rather than just one.

cli/handlers.go
itswisdomagain: > Changed to terminalprompt and found a marginlly smaller QR output as this one has no outside padding.

Not small enough. You had another idea of using a graphics library, discuss that with @raedah 

> Not sure how to remove those blank lines any ideas?

Look into the `terminalprompt` package

> A part of me feels like those are fine and Radeah was talking about double extra lines rather than just one.

Nah. The blank lines have no use

itswisdomagain:
> Changed to terminalprompt and found a marginlly smaller QR output as this one has no outside padding.

Not small enough. You had another idea of using a graphics library, discuss that with @raedah 

> Not sure how to remove those blank lines any ideas?

Look into the `terminalprompt` package
EDIT: Found it. Lines 38, 56 and 80 of `terminalprompt.go`

> A part of me feels like those are fine and Radeah was talking about double extra lines rather than just one.

Nah. The blank lines have no use

cli/handlers.go
itswisdomagain: That's because `RunCommand` in `cli.go` in line 83 uses the return value as is. You should return nil and in main.go, check if res is nil before attempting to print. Your main.go should look like this:

```
if res != nil {
     printResult(tabWriter(os.Stdout), res)
}
```

itswisdomagain:
I'm putting this PR as WIP. After discussing with @raedah on how to deal with the mighty qrcode image on terminal, and fixing other issues highlighted, remove the [WIP] label

Jujhar:
Fixed and almost ready to go.

qr-code-72

Jujhar:
Updating to add optional QR display via additional prompt.

itswisdomagain:
@Jujhar rebase your PR branch with master, fix conflicts and then push. Ensure you give attention to ALL previous review comments (there's a comment about using `nil` instead of `&response{}` that you haven't attended to)

When you've done that, rename this PR to remove the [WIP] flag, we don't want to spend time reviewing the code yet until those are done.

cli/handlers.go
Jujhar: No longer an issue

cli/commands/helpers.go
```diff
@@ -63,7 +63,11 @@ func getSendDestinationAddress(c *walletrpcclient.Client) (string, error) {
 		}
 
 		if !isValid {
-			return errors.New("invalid address")
+			if (address == "") {
```
itswisdomagain: This check should be done before line 60. There's no need to attempt to validate an empty address

cli/commands/helpers.go
```diff
@@ -85,6 +89,10 @@ func getSendAmount() (float64, error) {
 	validateAmount := func(input string) error {
 		amount, err = strconv.ParseFloat(input, 64)
 		if err != nil {
+			if (input == "") {
```
itswisdomagain: Always check for empty input before trying to "process" the input. This check should be done before line 90

cli/commands/receive.go
```diff
@@ -42,21 +43,22 @@ func (r ReceiveCommand) Run(client *walletrpcclient.Client, args []string) error
...
-	if err != nil {
-		return fmt.Errorf("Error generating QR Code: %s", err.Error())
+	// Print out address as string
+	fmt.Printf("%s\n", receiveResult.Address)
+
+	// Print out QR code
+	validateConfirm := func(address string) error {
+		return  nil
 	}
+	confirm, _ := terminalprompt.RequestInput("Would you like to a generate QR code? (y/n) ", validateConfirm)
```
itswisdomagain: If `validateConfirm` doesn't do anything, you can simply pass nil here instead.

cli/commands/receive.go
```diff
@@ -42,21 +43,22 @@ func (r ReceiveCommand) Run(client *walletrpcclient.Client, args []string) error
 		return err
 	}
 
-	qr, err := qrcode.New(receiveResult.Address, qrcode.Medium)
-	if err != nil {
-		return fmt.Errorf("Error generating QR Code: %s", err.Error())
+	// Print out address as string
+	fmt.Printf("%s\n", receiveResult.Address)
```
itswisdomagain: `fmt.Println(receiveResult.Address)`

cli/commands/receive.go
```diff
@@ -42,21 +43,22 @@ func (r ReceiveCommand) Run(client *walletrpcclient.Client, args []string) error
...
+	// Print out QR code
+	validateConfirm := func(address string) error {
+		return  nil
 	}
+	confirm, _ := terminalprompt.RequestInput("Would you like to a generate QR code? (y/n) ", validateConfirm)
 
-	columns := []string{
-		"Address",
-		"QR Code",
+	if confirm == "Yes" || confirm == "yes" {
```
itswisdomagain: Do case-insensitive comparison using `strings.EqualFold(confirm, "yes")`
Also, no need to assign `confirm = "y"` only to check the value of confirm again in the next line. Unnecesary.
Do all your comparison here:
`if strings.EqualFold(confirm, "yes") || strings.EqualFold(confirm, "y")`, then display the qrcode

cli/commands/send.go
```diff
@@ -74,6 +74,22 @@ func send(rpcclient *walletrpcclient.Client, custom bool) error {
...
+	confirm, _ := terminalprompt.RequestInput("Are you sure? (y/n) ", validateConfirm)
+
+	if confirm == "Yes" || confirm == "yes" {
+		confirm = "y"
+	}
+
+	if confirm != "y" {
+		fmt.Printf("Operation cancelled\n")
+		return  nil
+	}
```
itswisdomagain, line 99: The comments above apply to this file as well

itswisdomagain (changes requested):

Jujhar:
Done

cli/commands/helpers.go
```diff
@@ -83,6 +87,10 @@ func getSendAmount() (float64, error) {
 	var err error
 
 	validateAmount := func(input string) error {
+		if (input == "") {
+			return errors.New("That is did a valid amount. Try again.")
```
itswisdomagain: Typo here

itswisdomagain: Also, the message here should not be "invalid amount", but "Amount not specified" like was done in the address validate function

cli/commands/helpers.go
```diff
@@ -83,6 +87,10 @@ func getSendAmount() (float64, error) {
...
+		}
+
 		amount, err = strconv.ParseFloat(input, 64)
 		if err != nil {
 			return fmt.Errorf("error parsing amount: %s", err.Error())
```
itswisdomagain: This should be "Invalid amount. Try again"

cli/commands/send.go
```diff
@@ -86,17 +96,6 @@ func send(rpcclient *walletrpcclient.Client, custom bool) error {
...
-	}
-	rows := [][]interface{}{
-		[]interface{}{
-			"The transaction was published successfully",
-			result.TransactionHash,
-		},
-	}
-
-	termio.PrintTabularResult(termio.StdoutWriter, columns, rows)
+	fmt.Println("sent. txid %s", result.TransactionHash)
```
itswisdomagain: This should be `fmt.Printf("Sent. Txid %s", result.TransactionHash)` or `fmt.Println("Send. Txid", result.TransactionHash)`

itswisdomagain (changes requested):
In the `receive.go` and `send.go` files, you're not validating the requested "y/n" input properly. Compare how [similar validation is done here](https://github.com/raedahgroup/godcr/blob/master/cli/runner/walletloader.go#L77-L96)

Take note of the other comments below, rebase your PR branch with master and `go fmt ./...` your code when done before pushing.

cli/commands/helpers.go
Jujhar: That was not part of my code but updated

Jujhar:
Alright updated again.

cli/commands/send.go
```diff
@@ -72,7 +76,34 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading your response: %s", err.Error())
+		return err
+	}
+
+	if strings.EqualFold(confirm, "yes") || strings.EqualFold(confirm, "y") {
+		confirm = "y"
+	}
+
+	if confirm != "y" {
```
itswisdomagain: Instead of checking `if confirm != "y"` here, you can check `if strings.EqualFold(confirm, "n")` in line 96 and cancel the operation there instead.

Also note that the `validateConfirm` function will not accept "yes" or "no", only "y" or "n"

cli/commands/send.go
```diff
@@ -72,7 +76,34 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+		fmt.Printf("Operation cancelled\n")
```
itswisdomagain: `fmt.Println("Operation cancelled")`

cli/commands/helpers.go
itswisdomagain: This isn't resolved. This should be "Invalid amount. Try again"

itswisdomagain (changes requested):

Jujhar:
Fixed

cli/commands/send.go
```diff
@@ -72,7 +76,30 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	validateConfirm := func(userResponse string) error {
+		userResponse = strings.TrimSpace(userResponse)
+		userResponse = strings.Trim(userResponse, `"`)
+		if userResponse == "" || strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
+			return nil
+		} else {
+			return fmt.Errorf("invalid option, try again")
+		}
+	}
+	confirm, err := terminalprompt.RequestInput("Are you sure? (y/N) ", validateConfirm)
```
itswisdomagain: As seen here https://github.com/raedahgroup/godcr/issues/54#issuecomment-446785257, there is no default option for this prompt. User must enter "y" or "n" to proceed. This prompt should not have a capital option (capital options imply that if the user does not enter any value, it's assumed they accept the option in capital)

This should be "Are you sure? (y/n)". The space at the end of that message isn't needed as well

cli/commands/send.go
```diff
@@ -72,7 +76,30 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
 		return err
 	}
 
+	fmt.Println("You are about to send %f DCR to %s", sendAmount, destinationAddress)
+
+	validateConfirm := func(userResponse string) error {
+		userResponse = strings.TrimSpace(userResponse)
+		userResponse = strings.Trim(userResponse, `"`)
+		if userResponse == "" || strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
```
itswisdomagain: This validate function should not accept empty string values ("") since there is no default option. User must enter "y" or "n". See comment below for more

cli/commands/receive.go
```diff
@@ -44,21 +47,30 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
...
+			return nil
+		} else {
+			return fmt.Errorf("invalid option, try again")
+		}
 	}
 
-	columns := []string{
-		"Address",
-		"QR Code",
+	confirm, err := terminalprompt.RequestInput("Would you like to generate a QR code? (y/N) ", validateConfirm)
```
itswisdomagain: The trailing space at the end of this prompt message is not needed

cli/commands/send.go
```diff
@@ -72,7 +76,30 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
 		return err
 	}
 
+	fmt.Println("You are about to send %f DCR to %s", sendAmount, destinationAddress)
```
itswisdomagain: `fmt.Printf("You are about to send %f DCR to %s\n", sendAmount, destinationAddress)`

itswisdomagain (changes requested):
Last changes. While at it, run `go mod tidy` to remove references to unused imports in `go.mod` and `go.sum`

itswisdomagain (approved):

cli/commands/send.go
```diff
@@ -72,7 +76,30 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+		}
+	}
+	confirm, err := terminalprompt.RequestInput("Are you sure? (y/n) ", validateConfirm)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error reading your response: %s", err.Error())
+		return err
+	}
+
+	if strings.EqualFold(confirm, "n") {
+		fmt.Println("Operation cancelled")
```
itswisdomagain: Just `Canceled` would do. And I think it's safe to stick with American English (one L in Canceled, instead of 2)

cli/commands/send.go
```diff
@@ -72,7 +76,30 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	validateConfirm := func(userResponse string) error {
+		userResponse = strings.TrimSpace(userResponse)
+		userResponse = strings.Trim(userResponse, `"`)
+		if strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
+			return nil
+		} else {
+			return fmt.Errorf("invalid option, try again")
+		}
+	}
+	confirm, err := terminalprompt.RequestInput("Are you sure? (y/n) ", validateConfirm)
```
itswisdomagain: Remove the trailing space at the end of this prompt message

itswisdomagain (changes requested):

itswisdomagain (approved):

Issue #73, opened 2018-12-20
cli receive command enhancement


itswisdomagain:
Currently, dcrcli receive command works with an optional argument to specify the account **number** for which to generate a receive address.

We should instead accept account **name**. dcrwallet rpc has a method to retrieve account number using account name. And users would/may find it easier to remember account names rather than account numbers.

PR #76, opened 2018-12-23
Receive command now accept account name as argument

Quadriyanney:
Fixes #73

screenshot from 2018-12-23 15-44-46

### DIFF
--- a/cli/handlers.go
+++ b/cli/handlers.go
@@ -121,26 +121,27 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
 }
 
 func receive(walletrpcclient *rpcclient.Client, commandArgs []string) (*response, error) {
-	var receiveAddress uint32
+	var accountNumber uint32
 
 	// if no address passed in
 	if len(commandArgs) == 0 {
 		// display menu options to select account
 		var err error
-		receiveAddress, err = getSendSourceAccount(walletrpcclient)
+		accountNumber, err = getSendSourceAccount(walletrpcclient)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		// if an address was passed in eg. ./dcrcli receive 0 use that address
-		accountNumber, err := walletrpcclient.AccountNumber(commandArgs[0])
+		// if an account name was passed in e.g. ./dcrcli receive default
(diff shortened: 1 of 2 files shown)


cli/handlers.go
```diff
@@ -123,28 +121,26 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
...
-
 		// display menu options to select account
 		var err error
-		recieveAddress, err = getSendSourceAccount(walletrpcclient)
+		receiveAddress, err = getSendSourceAccount(walletrpcclient)
 		if err != nil {
 			return nil, err
 		}
 	} else {
 		// if an address was passed in eg. ./dcrcli receive 0 use that address
```
itswisdomagain: the example here in `e.g ./dcrcli ...` needs to be updated, the argument is no longer a number but text (account name)

cli/handlers.go
```diff
@@ -123,28 +121,26 @@ func send(walletrpcclient *rpcclient.Client, custom bool) (*response, error) {
...
-		x, err := strconv.ParseUint(commandArgs[0], 10, 32)
+		accountNumber, err := walletrpcclient.AccountNumber(commandArgs[0])
 		if err != nil {
-			return nil, fmt.Errorf("Error parsing account number: %s", err.Error())
+			return nil, fmt.Errorf("Error fetching account number: %s", err.Error())
 		}
-
-		recieveAddress = uint32(x)
+		receiveAddress = accountNumber
```
itswisdomagain: Why not just assign `recieveAddress ` to the result of `walletrpcclient.AccountNumber` in line 136? No need to introduce a new variable `accountNumber`

Talking about which, a better variable name would be `accountNumber` instead of `recieveAddress `. You should probably make this correction in line 124

itswisdomagain (changes requested):
This does the job. Just some minor corrections...

cli/handlers.go
Quadriyanney: Okay, will update it

cli/handlers.go
Quadriyanney: Okay, no problem

itswisdomagain (approved):

PR #79, opened 2018-12-24
 fixed dcrcli window build error

Sirmorrison:
fixed undefined syscall error while building on windows
capture

### DIFF
--- a/cli/terminalprompt/getinput.go
+++ b/cli/terminalprompt/getinput.go
@@ -60,15 +60,15 @@ func getPasswordInput(prompt string) (string, error) {
 
 func setTerminalEcho(on bool) error {
 
-	var cmdToExecute string
+	var sttyEchoArg string
 	if on{
 		fmt.Println()
-		cmdToExecute = "echo"
+		sttyEchoArg = "echo"
 	}else {
-		cmdToExecute = "-echo"
+		sttyEchoArg = "-echo"
 	}
 
-	cmd := exec.Command("stty", cmdToExecute)
+	cmd := exec.Command("stty", sttyEchoArg)
 	cmd.Stdout = os.Stdout
 	cmd.Stderr = os.Stderr
 	cmd.Stdin = os.Stdin
--- a/cli/terminalprompt/getinput.go
+++ b/cli/terminalprompt/getinput.go
@@ -9,10 +9,9 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	//"os/signal"
+	"os/signal"
 	"strings"
-	"syscall"
-	"golang.org/x/crypto/ssh/terminal"
+	"os/exec"
 )
 
 // getTextInput - Prompt for text input.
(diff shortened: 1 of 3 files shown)


barisere (changes requested):
This PR does not fix the issue in a way that preserves the current functionality of the program.
The lines @Sirmorrison  commented out served to allow users type in sensitive information such as passphrases without revealing them on the command line. This is similar to input masking, where a particular character, say "*", is displayed instead of each character of the input text. It's a basic security measure for requesting sensitive information at the terminal: someone else looking at the screen cannot easily tell what the user is typing.
An acceptable solution to this issue is to find a way to either disable terminal echo, or mask terminal input, on Windows. This PR cannot be merged, with this approach.

cli/terminalprompt/getinput.go
```diff
@@ -9,9 +9,10 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	"os/signal"
+	//"os/signal"
```
itswisdomagain: Soo many commented code on this file! If a code isn't needed, delete the code rather than comment it out. We certainly don't want unused code on the master branch, which means even if this PR accomplished it's objective, we can't merge it to master.

cli/terminalprompt/getinput.go
```diff
@@ -9,9 +9,10 @@ import (
...
 	"strings"
 	"syscall"
+	"golang.org/x/crypto/ssh/terminal"
```
itswisdomagain: There was a time we used this and other similar libraries to collect sensitive info (e.g. password) from terminal. But the use of an external library for this requirement was discouraged. Before we can consider using an external library here, we have to be sure there's no other way to solve the issue at hand

itswisdomagain (changes requested):
I haven't tested the PR to see if it'll build or run correctly. Looking at the code changes, it doesn't matter if it builds or runs correctly. There are unacceptable codes in this PR that have to be removed first before further consideration can be granted.

itswisdomagain:
I'm gonna make this a [WIP] PR. Once you have something that solves this problem in an acceptable manner, remove the [WIP] "flag"

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/golang/protobuf v1.2.0 // indirect
 	github.com/raedahgroup/dcrcli/walletrpcclient v0.0.1
 	github.com/skip2/go-qrcode v0.0.0-20171229120447-cf5f9fa2f0d8
+	golang.org/x/crypto v0.0.0-20180808211826-de0752318171
```
itswisdomagain: Remove this from your go.mod. Maybe run `go mod tidy` to cleanup your go.mod and go.sum files

cli/terminalprompt/getinput.go
```diff
@@ -54,35 +54,25 @@ func getPasswordInput(prompt string) (string, error) {
...
-		Dir:   "",
-		Env:   []string{},
-		Files: []uintptr{os.Stdin.Fd(), os.Stdout.Fd(), os.Stderr.Fd()},
-		Sys:   nil}
-	var ws syscall.WaitStatus
-	cmd := "echo"
-	if on == false {
-		cmd = "-echo"
-	}
+func setTerminalEcho(on bool) error {
```
itswisdomagain: This empty space can be done without

cli/terminalprompt/getinput.go
```diff
@@ -54,35 +54,25 @@ func getPasswordInput(prompt string) (string, error) {
...
 
-	// Enable/disable echoing.
-	pid, err := syscall.ForkExec(
-		"/bin/stty",
-		[]string{"stty", cmd},
-		&attrs)
-	if err != nil {
-		panic(err)
+	var cmdToExecute string
+	if on{
```
itswisdomagain: go fmt this file: `go fmt ./cli/terminalprompt/getinput.go`

cli/terminalprompt/getinput.go
```diff
@@ -54,35 +54,25 @@ func getPasswordInput(prompt string) (string, error) {
...
+func setTerminalEcho(on bool) error {
 
-	// Enable/disable echoing.
-	pid, err := syscall.ForkExec(
-		"/bin/stty",
-		[]string{"stty", cmd},
-		&attrs)
-	if err != nil {
-		panic(err)
+	var cmdToExecute string
```
itswisdomagain: This is actually an argument, not a command. In line 71, the actual command is `stty` while the second parameter is the argument to the `stty` command. Rename this variable to `sttyCmdArg` or `sttyEchoArg`.

itswisdomagain (changes requested):

itswisdomagain (approved):

itswisdomagain:
@barisere can you confirm that the changes made in this PR produce the expected behaviour on linux? As mentioned here (#92), this PR on windows command prompt doesn't produce the desired password masking behaviour but @raedah agrees it can be left that way for now to allow us focus on more important tasks

barisere:
> @barisere can you confirm that the changes made in this PR produce the expected behaviour on linux? As mentioned here (#92), this PR on windows command prompt doesn't produce the desired password masking behaviour but @raedah agrees it can be left that way for now to allow us focus on more important tasks

The behaviour on Linux did not change, as expected.

barisere (approved):

PR #82, opened 2018-12-26
Desktop interface

yaowork:
This creates the Desktop interface. Currently the following functionalities have been implemented 

- Wallet Balance 
- Transaction List 
- Generate Address and corresponding QR image 

The send page and it's functionalities are in a WIP state.
screenshot from 2018-12-26 18-06-15
screenshot from 2018-12-26 18-06-19
screenshot from 2018-12-26 18-06-22
screenshot from 2018-12-26 18-09-28
screenshot from 2018-12-26 18-09-33
screenshot from 2018-12-26 18-09-44


main.go
```diff
@@ -6,6 +6,8 @@ import (
 	"sort"
 	"strings"
 
+	"github.com/raedahgroup/dcrcli/desktop"
+
```
itswisdomagain: Unnecessary space?

main.go
```diff
@@ -39,6 +43,11 @@ func enterHTTPMode(serverAddress string, client *walletrpcclient.Client) {
 	web.StartHttpServer(serverAddress, client)
 }
 
+func enterDesktopMode(client *walletrpcclient.Client) {
+	d := desktop.New(client, config.AppName())
+	d.Render()
```
itswisdomagain: Since all we do with `d` is `Render()`, why not make `New` call render after creating the desktop object and not return anything? That function could be renamed to `StartDesktopApp`, following the pattern for http mode. Also, it would be nice to log to terminal that dcrcli is running in desktop mode, same as is done in http mode

desktop/desktop.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/aarzilli/nucular/label"
+	"github.com/aarzilli/nucular/rect"
+	"github.com/raedahgroup/dcrcli/walletrpcclient"
+)
+
+type pageHandler func(*nucular.Window)
+
+type Desktop struct {
+	window       nucular.MasterWindow
+	appName      string
```
itswisdomagain: This isn't required. Can always call `config.AppName()` from anywhere in this package to get the app name. Recent PRs have made it a pattern to get the app name strictly by calling `config.AppName()` and doing away with all appName variables

itswisdomagain (changes requested):

itswisdomagain:
I did a very preliminary run-through of the code. I'd like to keep the focus on getting this PR to a state where the functionalities and usability are "okay enough" before taking a deeper dive into the code base. We can always optimze the code later; and we will. For now, let's build something that works and works well.

I'll share observations and recommendations as I test the desktop app more.

desktop/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+
+type Desktop struct {
+	window       nucular.MasterWindow
+	currentPage  string
+	walletClient *walletrpcclient.Client
+	pageHandlers map[string]pageHandler
+}
+
+const (
+	scaling      = 1.8
```
itswisdomagain: This constant is only used in `desktop/helpers.go`, best declared there

desktop/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+	navWidth     = 260
+	contentWidth = 1000
```
itswisdomagain: This constant is never used, can go

desktop/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+	d.pageHandlers["balance"] = d.BalanceHandler
+	d.pageHandlers["receive"] = d.ReceiveHandler
+	d.pageHandlers["send"] = d.SendHandler
+	d.pageHandlers["transactions"] = d.TransactionsHandler
+
+	d.pageHandlers["selectutxos"] = d.selectUTXOSHandler
+	d.pageHandlers["generateaddress"] = d.generateAddressHandler
+}
+
+func resetVars() {
```
itswisdomagain: These variables are declared and used in `handlers.go` alone. This function should go there, prolly just below where they are declared

desktop/handlers.go
```diff
@@ -0,0 +1,348 @@
...
+	err error
+
+	// walletrpcclient responses
+	accountBalanceResponse  []*walletrpcclient.AccountBalanceResult
+	accountsResponse        []*walletrpcclient.AccountBalanceResult
+	generateAddressResponse *walletrpcclient.ReceiveResult
+	transactionsResponse    []*walletrpcclient.Transaction
+	utxosResponse           []*walletrpcclient.UnspentOutputsResult
+
+	// form values
```
itswisdomagain: these form values are not used, they can go

desktop/handlers.go
```diff
@@ -0,0 +1,348 @@
...
+				content.Label("Total", "LC")
+				content.Label("Spendable", "LC")
+				content.Label("Locked", "LC")
+				content.Label("Voting Authority", "LC")
+				content.Label("Unconfirmed", "LC")
+
+				// rows
+				for _, v := range accountBalanceResponse {
+					content.Label(v.AccountName, "LC")
+					content.Label(strconv.Itoa(int(v.Total/100000000)), "LC")
```
itswisdomagain: When converting a floating-point number to an int type, the fraction is discarded. You should round first.
Create a function for reusability:
```
func amountToString(amount float64) string {
	amount = math.Round(amount)
	return fmt.Sprintf("%d DCR", int(amount))
}
```

Then here and in `TransactionsHandler` (actually anywhere you need to display amount), call `amountToString(v.Total.ToCoin())`

desktop/handlers.go
```diff
@@ -0,0 +1,348 @@
...
+				content.Label("Type", "LC")
+				content.Label("Hash", "LC")
+
+				for _, tx := range transactionsResponse {
+					content.Row(20).Ratio(0.18, 0.12, 0.1, 0.15, 0.15, 0.3)
+					amount := strconv.Itoa(int(tx.Amount))
+					fee := strconv.Itoa(int(tx.Fee))
+
+					content.Label(tx.FormattedTime, "LC")
+					content.Label(amount, "LC")
```
itswisdomagain: Based on the previous comment about amounts... you can use `amountToString(tx.Amount)` here directly and `amountToString(tx.Fee)` in the next line

desktop/handlers.go
```diff
@@ -0,0 +1,348 @@
...
+	// draw page
+	if page := newWindow("ReceivePage", w, 0); page != nil {
+		page.header("Receive")
+
+		// content area
+		if content := page.contentWindow("Receive Content"); content != nil {
+			if err != nil {
+				content.setErrorMessage(err.Error())
+			} else {
+				accounts := make([]string, len(accountsResponse))
```
itswisdomagain: `accountNames` is a more descriptive variable name

desktop/desktop.go
```diff
@@ -0,0 +1,141 @@
...
+
+	w.LayoutSpacePushScaled(rect)
+	// style navigation pane
+	setNavStyle(d.window)
+	if sw := w.GroupBegin("Navigation Group", 0); sw != nil {
+		sw.Row(40).Dynamic(1)
+		if sw.Button(label.TA("Balance", "LC"), false) {
+			d.gotoPage("balance")
+		}
+		if sw.Button(label.TA("Send", "LC"), false) {
```
itswisdomagain: Rename this label to "Send (WIP)" so anyone running the app knows the features on that page are a work in progress

itswisdomagain (changes requested):

desktop/desktop.go
itswisdomagain: I noticed that once you follow the Send process to the point where the passphrase modal appears, you can't go back. You can't access other sections of the app. For now, remove the code that opens the modal. That code can only be merged if the modal doesn't prevent a user from accessing other parts of the app

itswisdomagain (approved):

Issue #69, opened 2018-12-19
transaction details view


raedah:
webui gets a transaction details page

cli can get a  showtransaction  cmd

UPDATE:
Cli command implemented in #83

itswisdomagain:
Currently, when listing transactions, the webui links each transaction hash to decred explorer (testnet.dcrdata.org or mainnet.dcrdata.org). It should instead direct to the transaction details page

barisere:
- What information will be displayed in the transaction details?
- What will the input to the command be? A transaction hash?

itswisdomagain:
> * What information will be displayed in the transaction details?

Compare what's displayed on decred explorer (ex. https://testnet.dcrdata.org/tx/2d574a40dbc31d6accfcd0d94a04a5267dc481f5c7bc2b0db6adc2bd17b1f029) and what's displayed on the dcrandroid app (https://play.google.com/store/apps/details?id=com.decred.dcrandroid.testnet). I'll share a screenshot of the transaction page on the app in a bit

> * What will the input to the command be? A transaction hash?

Yes. Unless there's something else we can use to uniquely identify a transaction

barisere:
> Compare what's displayed on decred explorer (ex. https://testnet.dcrdata.org/tx/2d574a40dbc31d6accfcd0d94a04a5267dc481f5c7bc2b0db6adc2bd17b1f029) and what's displayed on the dcrandroid app (https://play.google.com/store/apps/details?id=com.decred.dcrandroid.testnet). I'll share a screenshot of the transaction page on the app in a bit

Is every piece of information there needed in the CLI output?

barisere:
Here is what I suggest for the command line output of transaction details. It includes most of the information shown on the decred explorer.
How does it look? Is there too much information in the output? What should be removed?

![Suggested transaction details output](https://user-images.githubusercontent.com/20553476/50450530-fd0acd80-092e-11e9-8fb3-674982b07e8d.png)

itswisdomagain:
Looks good. But maybe too detailed? I'm thinking we could remove the inputs and outputs section and only show them when the command is issued with a flag say `-d/--detailed`

It'd help to compare what we have on dcrandroid

itswisdomagain:
For context, this is how the transaction page looks on dcrandroid:
![screenshot 2018-12-27 at 9 21 37 pm](https://user-images.githubusercontent.com/18400051/50551762-6881d180-0c86-11e9-9206-2009a526cc27.png)

[...]

PR #83, opened 2018-12-27
Transaction Details View for CLI.

barisere:
Partial solution to #69


barisere:
Here's the current output of the `get-transaction` command.
```
$ ./dcrcli -t -C dcrcli.conf get-transaction c3b6b67284ab25208008838e6cbb965c0da1fd33f57e14257b24d00579487ff2
Transaction       c3b6b67284ab25208008838e6cbb965c0da1fd33f57e14257b24d00579487ff2 
Confirmations     17194                                                            
Included in block f27f487905d0247b25147ef533fda10d5c96bb6c8e8308802025ab8472b6b6c3 
Type              REGULAR                                                          
Total sent        200 DCR                                                          
Time              Wed Nov 28, 2018 9:13PM                                          
Size              3.4 kB                                                           
Fee               0.003407 DCR                                                     
Rate              0.00100948 DCR/kB
```

[Compare this output to that on decred explorer](https://testnet.dcrdata.org/tx/c3b6b67284ab25208008838e6cbb965c0da1fd33f57e14257b24d00579487ff2). The block has displayed here is incorrect (it looks like a byte-reversed copy of the transaction hash). Is it possible that dcrwallet is returning incorrect data?
The time shown in this output is approximate. That may be because `time.Unix` accepts two arguments, but the second argument (the nanoseconds) was not provided.

barisere:
Here's how the current output of `get-transaction --detailed` looks.
```
$ ./dcrcli -t -C dcrcli.conf get-transaction --detailed dac27eb8b0238fb3a69b19dcb9254f865ae6df6d7ef7ecbd1bdc828ed0c6ed45
Transaction       dac27eb8b0238fb3a69b19dcb9254f865ae6df6d7ef7ecbd1bdc828ed0c6ed45 
Confirmations     35                                                               
Included in block 45edc6d08e82dc1bbdecf77e6ddfe65a864f25b9dc199ba6b38f23b0b87ec2da 
Type              REGULAR                                                          
Total sent        0.000253 DCR                                                     
Time              Fri Dec 28, 2018 7:54AM                                          
Size              0.3 kB                                                           
Fee               0.000253 DCR                                                     
Rate              0.00100396 DCR/kB                                                

Inputs                                                             
74448f25f85bb8a49c5b84695f3f95e9f8deaf96fe8edc5069f5cce2379400d2:0 0.998494 DCR 

Outputs                             
TsdCKdRnJ354Y7sHhRdbbGs3SrBP4qfXCbR pubkeyhash 0.997241 DCR (internal) 
TsfMhvoGbn7RVq3zKUJ9obE8ut7JFrnGwgT pubkeyhash 0.001 DCR 
```

cli/commands/commands.go
```diff
@@ -2,9 +2,10 @@ package commands
...
-	Send       SendCommand       `command:"send" description:"send a transaction"`
-	SendCustom SendCustomCommand `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
-	Receive    ReceiveCommand    `command:"receive" description:"show your address to receive funds"`
-	History    HistoryCommand    `command:"history" description:"show your transaction history"`
+	Balance        BalanceCommand        `command:"balance" description:"show your balance"`
+	Send           SendCommand           `command:"send" description:"send a transaction"`
+	SendCustom     SendCustomCommand     `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
+	Receive        ReceiveCommand        `command:"receive" description:"show your address to receive funds"`
+	History        HistoryCommand        `command:"history" description:"show your transaction history"`
+	GetTransaction GetTransactionCommand `command:"get-transaction" description:"show details of a transaction"`
```
itswisdomagain: I think `show-transaction` makes more sense for the command name. It's also what @raedah suggested in #69

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+	"strings"
+
+	"github.com/raedahgroup/dcrcli/cli"
+)
+
+// GetTransactionCommand requests for transaction details with a transaction hash.
+type GetTransactionCommand struct {
+	Detailed bool `short:"d" long:"detailed" description:"Display detailed transaction information"`
+	Args     struct {
+		Hash string `positional-arg-name:"transaction hash" required:"yes"`
```
itswisdomagain: `TxHash`, to differentiate from other hashes like `BlockHash`

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+	} `positional-args:"yes"`
+}
+
+// Execute runs the get-transaction command, displaying the transaction details to the client.
+func (g GetTransactionCommand) Execute(args []string) error {
```
itswisdomagain: Maybe use `getTransactionCommand` instead of `g` so lines 20 and 39 become more readable...

walletrpcclient/helpers.go
```diff
@@ -63,13 +63,13 @@ func (c *Client) processTransactions(transactionDetails []*pb.TransactionDetails
...
-			Amount:        dcrutil.Amount(amount).ToCoin(),
-			Fee:           dcrutil.Amount(txDetail.Fee).ToCoin(),
+			Amount:        dcrutil.Amount(amount),
+			Fee:           dcrutil.Amount(txDetail.Fee),
 			Type:          txDetail.TransactionType.String(),
 			Direction:     direction,
 			Testnet:       isTestnet,
 			Timestamp:     txDetail.Timestamp,
-			FormattedTime: time.Unix(txDetail.Timestamp, 0).Format("Mon Jan 2, 2006 3:04PM"),
+			FormattedTime: time.Unix(txDetail.Timestamp, 0).UTC().Format("Mon Jan 2, 2006 3:04PM"),
```
itswisdomagain: Why UTC? Could be misleading to a user especially since "UTC" isn't appended to this formatted time. The previous code shows the time as it is on the running user's PC. But I see how this can be a problem when running an http server in utc+2 and users access the page from a different timezone.

For cli mode, we should use local pc timezone in formatting the time. In web mode, we use utc but append "utc" to the formatted time so user doesn't get confused. I think this should be done in a separate PR, will create an issue for it. `FormattedTime` will need to be removed from this core package, allowing the `cli` and `web` packages do their own formatting.

For now, revert to what we had

walletrpcclient/types.go
```diff
@@ -36,10 +39,32 @@ type UnspentOutputsResult struct {
...
+}
+
+type TxOutput struct {
+	Address     string         `json:"address"`
+	Internal    bool           `json:"internal"`
+	Value       dcrutil.Amount `json:"value"`
+	ScriptClass string         `json:"scriptClass"`
+}
+
+type GetTransactionResponse struct {
```
itswisdomagain: How about `TransactionDetail` or `DetailedTransaction`, seeing as this is an extension of the basic transaction info. Or, we could rename `Transaction` to `BasicTransactionInfo` call this `TransactionInfo`. Something descriptive of the info this type holds but without the using the `<command>Response` format.

walletrpcclient/types.go
```diff
@@ -36,10 +39,32 @@ type UnspentOutputsResult struct {
...
 	Direction     TransactionDirection `json:"direction"`
-	Testnet       bool                 `json"testnet"`
+	Testnet       bool                 `json:"testnet"`
 	Timestamp     int64                `json:"timestamp"`
 	FormattedTime string               `json:"formatted_time"`
+	Size          int                  `json:"size"`
+}
+
+type TxInput struct {
+	Value            dcrutil.Amount `json:"value"`
```
itswisdomagain: `Value` or `Amount`, which is better?

walletrpcclient/walletrpcclient.go
```diff
@@ -8,9 +8,13 @@ import (
 	"net"
 	"sort"
 
+	"github.com/decred/dcrd/chaincfg"
+	"github.com/decred/dcrd/txscript"
+
```
itswisdomagain: can do without the space

itswisdomagain (changes requested):
Preliminary review. Will do a deep dive shortly and provide more feedback

walletrpcclient/walletrpcclient.go
```diff
@@ -21,14 +25,15 @@ import (
 
 type Client struct {
 	walletServiceClient pb.WalletServiceClient
+	isTestNet           bool
```
itswisdomagain: This variable is only used in `chainCfgParams` function to determine the `*chaincfg.Params` object to use. Replace this with `activeNet *chaincfg.Params` instead. Assign the correct value in `New`. Remove the `chainCfgParams` function.

If at any other time, we need to know if we're running testnet, we can do `client.activeNet.Name != "mainnet"`. We do a reverse check for mainnet, because there are 2 possible testnets - testnet and testnet3.

walletrpcclient/walletrpcclient.go
```diff
@@ -428,3 +432,73 @@ func (c *Client) GetTransactions() ([]*Transaction, error) {
 
 	return transactions, nil
 }
+
+func (c *Client) GetTransaction(transactionHash string) (*GetTransactionResponse, error) {
+	hash, err := chainhash.NewHashFromStr(transactionHash)
+	if err != nil {
+		return nil, err
```
itswisdomagain: When an invalid hash is used with this command, the error isn't very helpful. Could instead `return nil, fmt.Errorf("invalid hash: %s\n%s", transactionHash, err.Error())` since `chainhash.NewHashFromStr` only produces an error if the hash string is invalid, but the error it returns in itself is not always helpful.

walletrpcclient/walletrpcclient.go
```diff
@@ -428,3 +432,73 @@ func (c *Client) GetTransactions() ([]*Transaction, error) {
...
+	}
+	getTxRequest := &pb.GetTransactionRequest{TransactionHash: hash[:]}
+	getTxResponse, err := c.walletServiceClient.GetTransaction(context.Background(), getTxRequest)
```
itswisdomagain, line 445: Not all invalid transaction hashes are caught by `chainhash.NewHashFromStr`, from what I notice, any hash string starting with "d" but not longer than 64bytes is seen as valid by that function. But when you send the hash byte to dcrwallet rpc, it causes the running dcrwallet daemon to close and you get the error `rpc error: code = Unavailable desc = transport is closing`

This is potentially very problematic because it would require that the user restart dcrwallet to continue using godcr. Perhaps dcrwallet should have done a better job of handling such errors without causing the program to panic we should see if there's something we can do to prevent this, perhaps revalidate the hash before sending to dcrwallet rpc.

Here's a screenshot of the error trace printed on dcrwallet terminal window, it might help to look into dcrwallet code to find the cause of the problem and figure a way to prevent it:
screenshot 2018-12-30 at 6 24 54 pm

walletrpcclient/walletrpcclient.go
```diff
@@ -428,3 +432,73 @@ func (c *Client) GetTransactions() ([]*Transaction, error) {
...
+		return nil, err
+	}
+
+	transactionHex := fmt.Sprintf("%x", getTxResponse.GetTransaction().GetTransaction())
+	msgTx, err := txhelpers.MsgTxFromHex(transactionHex)
+	if err != nil {
+		return nil, err
+	}
+
+	txInfos, err := c.processTransactions([]*pb.TransactionDetails{getTxResponse.GetTransaction()})
```
itswisdomagain: In `helpers.go`, create a new function `processTransaction` to deal with 1 transaction and modify `processTransactions` to call the new function

walletrpcclient/walletrpcclient.go
```diff
@@ -428,3 +432,73 @@ func (c *Client) GetTransactions() ([]*Transaction, error) {
...
+}
+
+func chainCfgParams(c Client) *chaincfg.Params {
+	if c.isTestNet {
+		return &chaincfg.TestNet3Params
+	}
+	return &chaincfg.MainNetParams
+}
+
+func inputsFromMsgTxIn(txIn []*wire.TxIn) []TxInput {
```
itswisdomagain: This function and the next should go into `helpers.go`

walletrpcclient/walletrpcclient.go
```diff
@@ -428,3 +432,73 @@ func (c *Client) GetTransactions() ([]*Transaction, error) {
...
+	for i, input := range txIn {
+		txInputs[i] = TxInput{Value: dcrutil.Amount(input.ValueIn), PreviousOutpoint: input.PreviousOutPoint}
+	}
+	return txInputs
+}
+
+func outputsFromMsgTxOut(txOut []*wire.TxOut, walletCredits []*pb.TransactionDetails_Output, chainParams *chaincfg.Params) ([]TxOutput, error) {
+	txOutputs := make([]TxOutput, len(txOut))
+	for i, output := range txOut {
+		scriptClass, addrs, _, err := txscript.ExtractPkScriptAddrs(output.Version, output.PkScript, chainParams)
```
itswisdomagain: Don't think we need `scriptClass`. Not really necessary to show as much detail as is displayed on explorer

walletrpcclient/types.go
```diff
@@ -36,10 +39,32 @@ type UnspentOutputsResult struct {
...
+	PreviousOutpoint wire.OutPoint  `json:"previousOutpoint"`
```
itswisdomagain: Examine usages of this variable and simplify its type. I don't think we need the `wire.OutPoint` type

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+	transaction, err := cli.WalletClient.GetTransaction(g.Args.Hash)
+	if err != nil {
+		return err
+	}
+	basicOutput := "Transaction\t%s\t\n" +
```
itswisdomagain: `Hash` instead

itswisdomagain: The extra tab (`\t`) at the end of each line before `\n` isn't needed. It causes anomaly in small-width terminal windows:
screenshot 2018-12-30 at 6 18 27 pm

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		"Confirmations\t%d\t\n" +
+		"Included in block\t%s\t\n" +
+		"Type\t%s\t\n" +
+		"Total sent\t%s\t\n" +
```
itswisdomagain: Not all transactions are sent transactions. It would be nice to use `Amount sent`, `Amount received` or `Amount transferred` for the different transaction directions.

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		"Time\t%s\t\n" +
```
itswisdomagain: dcrandroid uses `Date`, am not sure which will send the message across faster to the user

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		"Size\t%s\t\n" +
+		"Fee\t%s\t\n" +
+		"Rate\t%s/kB\t"
+
+	txSize := fmt.Sprintf("%.1f kB", float64(transaction.Size)/1000)
+	basicOutput = fmt.Sprintf(basicOutput, transaction.Hash, transaction.Confirmations, transaction.BlockHash,
```
itswisdomagain: break each value into a line of its own so we can easily tally it with the field name above

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		transaction.Type, transaction.Amount, transaction.FormattedTime, txSize, transaction.Fee, transaction.Rate)
+
+	detailedOutput := strings.Builder{}
+	if g.Detailed {
+		detailedOutput.WriteString("\nInputs\t\n")
```
itswisdomagain: same comment about trailing tab space

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		for _, input := range transaction.Inputs {
+			detailedOutput.WriteString(fmt.Sprintf("%s\t%s\t\n", input.PreviousOutpoint, input.Value))
```
itswisdomagain: same comment about trailing tab space

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		}
+		detailedOutput.WriteString("\nOutputs\t\n")
+		for _, out := range transaction.Outputs {
+			detailedOutput.WriteString(fmt.Sprintf("%s\t%s\t%s", out.Address, out.ScriptClass, out.Value.String()))
```
itswisdomagain: same comment about trailing tab space

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+	basicOutput = fmt.Sprintf(basicOutput, transaction.Hash, transaction.Confirmations, transaction.BlockHash,
+		transaction.Type, transaction.Amount, transaction.FormattedTime, txSize, transaction.Fee, transaction.Rate)
+
+	detailedOutput := strings.Builder{}
+	if g.Detailed {
+		detailedOutput.WriteString("\nInputs\t\n")
+		for _, input := range transaction.Inputs {
+			detailedOutput.WriteString(fmt.Sprintf("%s\t%s\t\n", input.PreviousOutpoint, input.Value))
+		}
+		detailedOutput.WriteString("\nOutputs\t\n")
```
itswisdomagain: same comment about trailing tab space

cli/commands/get-transaction.go
```diff
@@ -0,0 +1,57 @@
...
+		for _, out := range transaction.Outputs {
+			detailedOutput.WriteString(fmt.Sprintf("%s\t%s\t%s", out.Address, out.ScriptClass, out.Value.String()))
+			if out.Internal {
+				detailedOutput.WriteString(" (internal)")
+			}
+			detailedOutput.WriteString("\t\n")
```
itswisdomagain: same comment about trailing tab space

itswisdomagain (changes requested):

itswisdomagain:
Regarding the output of the detailed command, I particularly like the simplicity of dcrandroid:
screenshot 2018-12-27 at 9 45 39 pm

For Inputs, we should probably display amounts first. It's more friendly to the eyes.
For outputs, `Amount  Address (internal/external)` does it as well.

Also, for `detailed` view, title the first section:
```
General Info
Hash        dac27eb8b0238fb3a69b19dcb9254f865ae6df6d7ef7ecbd1bdc828ed0c6ed45 
Confirmations     35                                                               
Included in block 45edc6d08e82dc1bbdecf77e6ddfe65a864f25b9dc199ba6b38f23b0b87ec2da 
Type              REGULAR                                                          
Total sent        0.000253 DCR                                                     
Time              Fri Dec 28, 2018 7:54AM                                          
Size              0.3 kB                                                           
Fee               0.000253 DCR                                                     
Rate              0.00100396 DCR/kB                                                

Inputs                                                             
74448f25f85bb8a49c5b84695f3f95e9f8deaf96fe8edc5069f5cce2379400d2:0 0.998494 DCR 

Outputs                             
TsdCKdRnJ354Y7sHhRdbbGs3SrBP4qfXCbR pubkeyhash 0.997241 DCR (internal) 
TsfMhvoGbn7RVq3zKUJ9obE8ut7JFrnGwgT pubkeyhash 0.001 DCR 
```

walletrpcclient/walletrpcclient.go
barisere: I cannot reproduce this error. Looking at the code for `NewHashFromString`, I understand that it allows strings shorter than 64 characters. But the special case of a string starting with "d" crashing dcrwallet is weird.
```
$ ./godcr -t -C ../dcrcli.conf show-transaction d970a25cbe7421f9371180baa1
rpc error: code = NotFound desc = wallet.TransactionSummary(00000000000000000000000000000000000000d970a25cbe7421f9371180baa1): item does not exist
```
The above does not crash dcrwallet. dcrwallet logs the following error.
```
2018-12-30 20:44:15.418 [ERR] GRPC: Unary method /walletrpc.WalletService/GetTransaction invoked by 127.0.0.1:55700 errored: rpc error: code = NotFound desc = wallet.TransactionSummary(00000000000000000000000000000000000000d970a25cbe7421f9371180baa1): item does not exist
```

itswisdomagain: Ah. Maybe it's been fixed and I'm using an older version of dcrwallet

itswisdomagain (approved):

PR #86, opened 2018-12-28
Make packages commands and config disjoint

barisere:
@ademuanthony discovered that a circular dependency occurred between cli/commands and config packages, while he was working on another PR. I noticed that before, but I managed to sidestep it for a while.
This PR makes those packages disjoint. config package now handles only the top-level application options and configuration. Both the `config.Config` type and the `commands.CliCommands` type are mediated by a third type, `AppRoot`, in the cli package.

I also removed a dependency on a global variable that provided the wallet rpc client to the commands. That dependency is now injected at runtime. A consequence of this change is that the `Execute` methods on the commands is now a noop in any command that requires a wallet rpc client, but because it is required for `go-flags` to identify the type as a `Commander`, it cannot be removed yet. The alternative is to provide our own routing of arguments to commands, which is still some more code. I'm fine with a three-line, noop method copied around, because it does nothing more than satisfying `go-flags`. If we ever replace `go-flags`, all those methods will disappear easily.


cli/cli.go
```diff
@@ -1,17 +1,36 @@
...
+	Commands commands.CliCommands
+	Config   config.Config
 }
 
-var (
-	WalletClient *walletrpcclient.Client
-	StdoutWriter = tabWriter(os.Stdout)
-)
+// CommandHandler provides a type name for the command handler to register on flags.Parser
+type CommandHandler func(flags.Commander, []string) error
```
itswisdomagain: This can be private

cli/cli.go
```diff
@@ -1,17 +1,36 @@
...
+
+// DefaultCommandHandler provides a command handler that provides walletrpcclient.Client
+// to commands.WalletCommandRunner types. Other command that satisfy flags.Commander and do not
+// depend on walletrpcclient.Client will be run as well.
+// If the command does not satisfy any of these types, ErrNotSupported will be returned.
+func DefaultCommandHandler(client *walletrpcclient.Client) CommandHandler {
+	return func(command flags.Commander, args []string) error {
+		if command == nil {
+			return fmt.Errorf("unsupported command")
```
itswisdomagain: According to the go-flags doc, `command == nil` isn't a case of unsupported command
> The command passed into CommandHandler may be nil in case there is no command to be executed when parsing has finished.

cli/commands/balance.go
```diff
@@ -12,9 +12,16 @@ type BalanceCommand struct {
 	Detailed bool `short:"d" long:"detailed" description:"Display detailed account balance report"`
 }
 
-// Execute runs the `balance` command, displaying the user's account balance.
+// Execute is a stub method to satisfy the commander interface, so that
```
itswisdomagain: Here's a thought. To avoid having this method stub on all commands implementing `WalletCommandRunner`, you can create a `CommanderStub` struct in `commands.go` and make such commands embed `CommanderStub` in their definition.

In commands.go,
```
// CommanderStub implements `flags.Commander` using a noop Execute method to satisfy the commander interface
// Commands embedding this struct would ideally implement `WalletCommandRunner` so that their `Run` method can
// be invoked by the custom command handler which will inject the necessary dependencies to run the command.
type CommanderStub struct {}
func (c CommanderStub) Execute(args []string) error {
	return nil
}
```

Then in files like this (balance.go), the command struct becomes
```
type BalanceCommand struct {
	*CommanderStub
        ....
}
```

cli/cli.go
```diff
@@ -1,17 +1,36 @@
...
-	StdoutWriter = tabWriter(os.Stdout)
-)
+// CommandHandler provides a type name for the command handler to register on flags.Parser
+type CommandHandler func(flags.Commander, []string) error
+
+// DefaultCommandHandler provides a command handler that provides walletrpcclient.Client
+// to commands.WalletCommandRunner types. Other command that satisfy flags.Commander and do not
+// depend on walletrpcclient.Client will be run as well.
+// If the command does not satisfy any of these types, ErrNotSupported will be returned.
+func DefaultCommandHandler(client *walletrpcclient.Client) CommandHandler {
```
itswisdomagain: This could better be called `CommandHandlerWrapper` seeing as it is a wrapper around the actual handler for each command.

cli/commands/commands.go
```diff
@@ -8,3 +13,10 @@ type CliCommands struct {
 	Receive    ReceiveCommand    `command:"receive" description:"show your address to receive funds"`
 	History    HistoryCommand    `command:"history" description:"show your transaction history"`
 }
+
+// WalletCommandRunner defines an optional interface that application commands dependent on
```
itswisdomagain: "optional" isn't really necessary, interfaces by default are optional
also, application commands dependent on walletrpcclient.Client do need to implement this interface, so it isn't optional for such commands, it is "optional" for other commands, but then aren't all interfaces optional?

cli/commands/receive.go
```diff
@@ -3,8 +3,10 @@ package commands
 import (
 	"fmt"
 
-	"github.com/raedahgroup/dcrcli/cli"
-	qrcode "github.com/skip2/go-qrcode"
+	"github.com/raedahgroup/dcrcli/walletrpcclient"
+
```
itswisdomagain: The space can be done without

cli/termio/writer.go
```diff
@@ -44,9 +48,15 @@ func PrintResult(w *tabwriter.Writer, res *Response) {
...
 func PrintStringResult(output ...string) {
-	writer := tabWriter(os.Stdout)
+	writer := TabWriter(os.Stdout)
 	for _, str := range output {
 		fmt.Fprintln(writer, str)
 	}
 	writer.Flush()
 }
+
+type Response struct {
```
itswisdomagain: Every other thing in this file and (ideally in this package), strictly has to do with terminal input/output. This? Not really. It's sort of a bridge between command handlers and terminal output.

2 thoughts:
- Put this in the commands package
- Or do away with it all together, let the `PrintResult` function take `columns []string` and `result [][]interface{}`

I prefer the second approach, it's more readable. We could even change the function to `PrintTabularResult` and the parameters to `column` and `rows`. Now that's more descriptive of what this function actually does.

config/config.go
```diff
@@ -56,40 +53,35 @@ func AppName() string {
...
 	// Parse command line options again to ensure they take precedence.
 	_, err = parser.Parse()
-	if err != nil && !IsFlagErrorType(err, flags.ErrCommandRequired) {
-		return nil, parser, err
+	if err != nil && !IsFlagErrorType(err, flags.ErrHelp) {
+		return config, parser, err
 	}
 
-	return &config, parser, nil
+	return config, parser, nil
```
itswisdomagain: All return paths return `config`, `parser` and `err`. Maybe you could use named return parameters. _Could_

main.go
```diff
@@ -2,12 +2,12 @@ package main
 
 import (
 	"fmt"
+	"github.com/raedahgroup/dcrcli/cli"
```
itswisdomagain: IDE did this no doubt. This import can return to where it was...

cli/cli.go
```diff
@@ -1,17 +1,36 @@
...
+	"github.com/raedahgroup/dcrcli/config"
 	"github.com/raedahgroup/dcrcli/walletrpcclient"
 )
 
-type Response struct {
-	Columns []string
-	Result  [][]interface{}
+// AppRoot is the entrypoint to the cli application.
+// It defines both the commands and the options available.
+type AppRoot struct {
```
itswisdomagain: `App` signifies the entire dcrcli program, including the web and other interfaces. This struct defines cli-only properties, a better name can be used

itswisdomagain (changes requested):
Very neat work. Some changes here and there

itswisdomagain:
This just came to mind:
In main.go, `config.LoadConfig()` will reject cli command-specific flags/options as unknown. Example is `dcrcli balance -d`

2 possible solutions:
- `config.LoadConfig()` should use `AppRoot` as the first argument to `flags.NewParser`. The function would then need to register a noop command handler on `parser` to prevent commands from being executed during the initial config parsing. One problem with this approach is that, you'll likely run into the circular dependency issue once again. Another caveat of this approach is that, if http mode is enabled, cli-only flags/options will be allowed to roam free if used "aimlessly" with the associated cli command. An example is `dcrcli balance -d --http`. Don't know about you, but I think a user running such command is confused and needs to be reminded that there is a difference between cli mode and http mode.

- Use `flags.HelpFlag|flags.IgnoreUnknown` as second argument to `flags.NewParser`. This has the advantage of deferring unknown flag errors to be handled by the cli command parser in `enterCliMode` of main.go. But it also means that if http mode is enabled, unknown options will be permitted to roam free. To circumvent that caveat, http mode should trigger another parsing that does not ignore unknown flags/options and does not include cli commands. Doing this second parsing for http mode has the added advantage of reminding a "confused" user of the cli/http mode difference if a user passes a cli-only flag. For example, `dcrcli balance -d --http` will give an error `-d` is an unknown option. Of course, that message can be better, we should also tell the user that commands are not accepted/allowed in http mode. We can do that if we modify `config.LoadConfig()` to also return `args`. In http mode, if `len(args) > 0`, a command was passed, we should remind the user that http mode don't work with commands.

The following is how I expect `config.LoadConfig()` to appear using the second possibility:
```go
func LoadConfig(ignoreUnknownOptions bool) (Config, *flags.Parser, []string, error) {
	// load defaults first
	config := defaultConfig()

	var parserOptions flags.Options
	if ignoreUnknownOptions {
		parserOptions = flags.HelpFlag|flags.IgnoreUnknown
	} else {
		parserOptions = flags.HelpFlag
	}
	parser := flags.NewParser(&config, parserOptions)

	args, err := parser.Parse()
	if err != nil && !IsFlagErrorType(err, flags.ErrHelp) {
		return config, parser, args, err
	}

	if config.ShowVersion {
		return config, parser, args, fmt.Errorf(AppVersion())
	}

	// Load additional config from file
	err = flags.NewIniParser(parser).ParseFile(config.ConfigFile)
	if err != nil {
			return config, parser, args, fmt.Errorf("Error parsing configuration file: %v", err.Error())
		}
		return config, parser, args, err
	}

	// Parse command line options again to ensure they take precedence.
	args, err = parser.Parse()
	if err != nil && !IsFlagErrorType(err, flags.ErrHelp) {
		return config, parser, args, err
	}

	return config, parser, args, nil
}
```

cli/cli.go
barisere: If the command passed into `CommandHandler` is nil, it means we registered a command that does not satisfy the `CommandHandler` interface. it's something we should catch early. For now, what error message can be used?

cli/commands/balance.go
barisere: How did I miss that? Thank you :slightly_smiling_face:

main.go
barisere: IDE :man_facepalming:

config/config.go
barisere: That would be convenient, but it tends to produce an inconsistent style and obscure the return values of the function at the return site.
```go
func LoadConfig() (config Config, parser *flags.Parser, err error) {
	// load defaults first
	config = defaultConfig()

	parser = flags.NewParser(&config, flags.HelpFlag)

	_, err = parser.Parse()
	if err != nil && !IsFlagErrorType(err, flags.ErrHelp) {
		return
	}

	if config.ShowVersion {
		return config, parser, fmt.Errorf(AppVersion())
	}

	// Load additional config from file
	err = flags.NewIniParser(parser).ParseFile(config.ConfigFile)
	if err != nil {
			return config, parser, fmt.Errorf("Error parsing configuration file: %v", err.Error())
		}
		return
	}

	// Parse command line options again to ensure they take precedence.
	_, err = parser.Parse()
	if err != nil && !IsFlagErrorType(err, flags.ErrHelp) {
		return
	}

	return
}
```
At some points we simply return, at some points we say what we're returning, and at the final return it's not clear what the return values should be, especially that err should be nil. Thus, you have to keep the code trace in your head at every point to know what each return value will be (debugger mode). I think named returns can make readability worse.

barisere:
https://github.com/raedahgroup/dcrcli/pull/86#issuecomment-450446124
@itswisdomagain I've addressed this in my latest commits.

cli/cli.go
itswisdomagain: "Invalid command" is good. From your explanation, the command is actually supported but not properly defined. If the command is not supported, the program won't even get to this point. Would it?

cli/cli.go
itswisdomagain: Looks like you missed this... `CommandHandler` should be private, it isn't used outside this file and I don't suppose we'll need to create an instance of this type outside this package, would we?

cli/cli.go
itswisdomagain: Since this only occurs for commands not properly defined, "command is not execution-ready" sounds more appropriate. Unsupported commands would not even get to this point. So the user should know that the command is actually supported (we registered it on the parser afterall and it shows in `dcrcli -h`) but for some reason the command isn't ready for execution. A command that is actually ready for execution would implement `Commander`.

cli/termio/helpers.go
```diff
@@ -1,4 +1,4 @@
-package cli
+package termio
```
itswisdomagain: Forgot to mention this previously. I think this file belongs in the `commands` package. The functions here exist for the sole purpose of assisting the process of executing one or more commands. `termio` is just 1 of the packages that this file uses in performing its functions.

ademuanthony:
Great effort.
Though the circular dependency still continues for my current use case as I want to reference the `CliRoot` in the `cli` package in `commands`; whereas the `cli` package already references the `commands` package.

So I am thinking, what if the commands are moved to the `cli` package, seeing that that is the main component in that package, it may not be that bad to have it at the root.
It may look like we will lose the organization that the current structure provides, but with the current convention of adding 'Command' suffix to the command names, I think it makes sense enough to have `cli.SendCommand`

cli/cli.go
```diff
@@ -1,17 +1,36 @@
...
+	"github.com/raedahgroup/dcrcli/config"
 	"github.com/raedahgroup/dcrcli/walletrpcclient"
 )
 
-type Response struct {
-	Columns []string
-	Result  [][]interface{}
+// CliRoot is the entrypoint to the cli application.
+// It defines both the commands and the options available.
+type CliRoot struct {
```
ademuanthony: Since we are already in the `cli` package, I am thinking `Root` maybe a sufficient name for the struct

config/config.go
```diff
@@ -56,40 +53,51 @@ func AppName() string {
...
-	}
-
-	_, err := parser.Parse()
-	if err != nil && !IsFlagErrorType(err, flags.ErrCommandRequired) {
-		return nil, parser, err
+	if ignoreUnknownOptions {
+		parser.Options = parser.Options | flags.IgnoreUnknown
 	}
 
-	if config.ShowVersion {
```
itswisdomagain: We should retain the behaviour of handling `-v` cases in this function. Not in `main.go`. Moving this handler to `main.go` does not provide any advantage. Rather, main.go refers back to this file to get the version info to be displayed.

My estimation is, you moved this handler from here to ensure that errors arising from invalid flags/options are handled before attempting to print help information. If so, that goal is not achieved because before handling `-v` cases in main.go, the previous line explicitly instructs `config.ParseConfig` to ignore unknown flags/options.

I recommend that we can continue to simply check for `-v` presence after first command-line options parsing, display version info and exit without even parsing the config file. If we want to bother ourselves with catching invalid flags/options, we would run into trouble with command-specific flags/options that will be seen as invalid when running `config.LoadConfig()`

main.go
```diff
@@ -28,6 +32,10 @@ func main() {
 	}
 
 	if appConfig.HTTPMode {
+		if len(args) > 0 {
+			fmt.Println("cannot use --http with a command")
```
itswisdomagain: `args` returned from `config.LoadConfig(true)` contains all elements encountered that are not valid config options. This would include commands, yes, since commands are not part of the config struct used in `config.LoadConfig` but it could also contain command-specific flags/options, invalid flags/options and any other random unexpected text encountered while parsing.

For example, `args` returned for
- `dcrcli balance --http` would contain "balance", a valid command but not expected in http mode
- `dcrcli -b --http` would contain "-b", an entirely unknown option
- `dcrcli balance -d --http` would contain "balance and -d", both valid command and valid command-flag respectively but unexpected in http mode
- `dcrcli hey --http` would contain "hey", an entirely unexpected, random text
- `dcrcli -t balance -d hey -http` would contain "balance, -d and hey" as unknown "options"

To accurately deal with all possibilities above and other similar possibilities, this message displayed to the user should be more informative. I thought of `fmt.Printf("unexpected command, argument, flag or option: '%s'\n", strings.Join(args, ", "))`. For the scenarios above, the message would be
- unexpected command, argument, flag or option: 'balance'
- unexpected command, argument, flag or option: '-b'
- unexpected command, argument, flag or option: 'balance, -d'
- unexpected command, argument, flag or option: 'hey'
- unexpected command, argument, flag or option: 'balance, -d, hey'

itswisdomagain (changes requested):

cli/cli.go
itswisdomagain: Agreed. I also think `Root` doesn't exactly sound like a good name. But then...

cli/cli.go
barisere: We shouldn't let commands that are unavailable show up in the help message. How about letting the user know we've messed something up, and asking them to report it to us.
> the command <command name\> was not properly setup. Please report this bug at https://github.com/raedahgroup/godcr/issues.

dcrcli.conf
```diff
@@ -0,0 +1,27 @@
+[Application Options]
```
itswisdomagain: Remove this file

config/config.go
```diff
@@ -26,17 +24,16 @@ var (
 
 // Config holds the top-level options for the CLI program.
 type Config struct {
-	ShowVersion       bool   `short:"v" long:"version" description:"Display version information and exit"`
+	ShowVersion       bool   `short:"v" long:"version" description:"Display version information and exit. Every other flag or command is ignored."`
```
itswisdomagain: "Every other" should be "Any other'

itswisdomagain (changes requested):

dcrcli.conf
barisere: [PR 64](https://github.com/raedahgroup/godcr/pull/64) should take care of this. From when PR 64 gets merged, that file will not be needed again. But this PR makes no assumption about that file.

itswisdomagain (approved):

PR #98, opened 2018-12-31
./godcr help

### LINKED ISSUE #51: ./dcrcli help

ademuanthony:
fixes [51](https://github.com/raedahgroup/dcrcli/issues/51).
Added a './dcrcli help [command]' command that shows general help message or specific help message when called with the optional [command] param
This is a replacement for [PR 63](https://github.com/raedahgroup/godcr/pull/63) in which I had some issue and have to take the work to another branch
help-reworked

Individual command
help-rewored-single


cli/commands/help.go
```diff
@@ -0,0 +1,43 @@
...
+
+	return nil
+}
+
+func PrintCommandHelp(appName string, command *flags.Command) {
+	helpParser := flags.NewParser(nil, flags.HelpFlag)
+	helpParser.Name = appName
+	helpParser.Active = command
+	helpParser.WriteHelp(os.Stderr)
+	fmt.Printf("To view application options, use '%s help'\n", config.AppName())
```
barisere: Use `appName` instead of calling `config.AppName`.
Also, this function is equivalent to that in main. Delete the one in main, and use this one there.

barisere (changes requested):
Nice work @ademuanthony, it's almost done. Let's see...

cli/commands/help.go
```diff
@@ -29,7 +29,7 @@ func (h HelpCommand) Run(parser *flags.Parser) error {
 		return fmt.Errorf("unknown command %q", h.Args.CommandName)
 	}
 
-	PrintCommandHelp(targetCommand.Name, targetCommand)
+	PrintCommandHelp(config.AppName(), targetCommand)
```
barisere: `package commands` should not depend on `package config`. That was the purpose of the work we did splitting them. The current app name is registered on the parser as `parser.Name`, use that instead.

cli/commands/commands.go
```diff
@@ -2,10 +2,11 @@ package commands
...
 // Commands defines the commands and options available on the cli
 type Commands struct {
-	Balance         BalanceCommand         `command:"balance" description:"show your balance"`
+	Balance         BalanceCommand         `command:"balance" description:"shows total balance for each account in wallet" long-description:"Shows total balance for each account in wallet. Also shows spendable balance if different from total balance. Use -d/--detailed to view a more detailed balance breakdown"`
 	Send            SendCommand            `command:"send" description:"send a transaction"`
 	SendCustom      SendCustomCommand      `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
 	Receive         ReceiveCommand         `command:"receive" description:"show your address to receive funds"`
 	History         HistoryCommand         `command:"history" description:"show your transaction history"`
 	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"show details of a transaction"`
+	Help            HelpCommand            `command:"help" description:"show help for a command"`
```
itswisdomagain: `godcr help` doesn't show help for any command, `godcr help <command>` does. This description should be updated. Also, the description should let users know they can run `godcr help <command-name>` for command-specific help. They shouldn't need to run `godcr help -h` before the see the optional command name argument.

cli/commands/help.go
```diff
@@ -0,0 +1,40 @@
...
+		parser.WriteHelp(termio.StdoutWriter)
+		return nil
+	}
+
+	var targetCommand = parser.Find(h.Args.CommandName)
+	if targetCommand == nil {
+		return fmt.Errorf("unknown command %q", h.Args.CommandName)
+	}
+
+	helpParser := flags.NewParser(nil, flags.HelpFlag)
```
itswisdomagain: There's a function performing practically the same thing as this in `cli.go`. Bring that function to this file and have `cli.go` invoke the function moved here

cli/commands/help.go
```diff
@@ -0,0 +1,40 @@
...
+	helpParser.Name = parser.Name
+	helpParser.Active = targetCommand
+	helpParser.WriteHelp(os.Stderr)
```
itswisdomagain: The output of this command shouldn't go to `Stderr`, it's a valid command. Display the output as you would for any other command - to `Stdout`. That's what's done in the version of this code in `cli.go`

itswisdomagain (changes requested):
Below are a few ways the code could be improved.

Also, the help message displayed for specific commands is not detailed enough yet. I think only balance command has help sufficient info.

Compare https://github.com/raedahgroup/godcr/pull/63#issuecomment-449841140 and https://github.com/raedahgroup/godcr/pull/63#issuecomment-449847773 for what's expected.

cli/runner/parsercommand.go
```diff
@@ -0,0 +1,23 @@
+package runner
```
barisere: Why is this package necessary?

cli/runner/parsercommand.go
```diff
@@ -0,0 +1,23 @@
...
+	flags.Commander
+}
+
+// ParserCommand implements `flags.Commander`, using a noop Execute method to satisfy `flags.Commander` interface
+// Commands embedding this struct should ideally implement `ParserCommandRunner` so that their `Run` method can
+// be invoked by `CommandRunner.Run` which will inject the necessary dependencies to run the command
+type ParserCommand struct{}
+
+// Noop Execute method added to satisfy `flags.Commander` interface
+func (w ParserCommand) Execute(args []string) error {
```
barisere: This stub is not different from `CommanderStub` in cli/commands/commands.go. Why do we have it here?

cli/runner/parsercommand.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"context"
+	"github.com/jessevdk/go-flags"
+)
+
+// ParserCommandRunner defines the Run method that cli commands that depends on
+// flags.Parser can implement to have it injected at run time
+type ParserCommandRunner interface {
+	Run(ctx context.Context, parser *flags.Parser, args []string) error
```
barisere: What would the use of `ctx` be? There's no need for a context here.

barisere (changes requested):
@ademuanthony you force-pushed your last commit, erasing the entire history of the previous commits published on this PR. This erases the context of the previous reviews, making it hard to track the progress of this PR. The merge conflict you had was trivial, by inspection. You should avoid force-pushing and erasing history as much as possible. It's usually problematic.

cli/runner/parsercommand.go
ademuanthony: this was already in the master. I was only making my branch to work with what was already on the ground.

cli/runner/parsercommand.go
ademuanthony: CommanderStub in no longer in the master. We now have WalletCommand

cli/runner/parsercommand.go
ademuanthony: I can't see it being useful
Will remove it

cli/runner/parsercommand.go
barisere: I wouldn't know, because of the force-push. We can make changes anyway.

cli/runner/parsercommand.go
barisere: Wow, and that slipped by without a review. :unamused:

cli/commands/help.go
```diff
@@ -30,11 +30,20 @@ func (h HelpCommand) Run(ctx context.Context, parser *flags.Parser, args []strin
...
+	helpParser.Name = appName
+	helpParser.Active = command
+	helpParser.WriteHelp(os.Stdout)
+	fmt.Printf("To view application options, use '%s help'\n", appName)
+}
+
+/*
+I have looked for how to get command-specific options but did
+find a way around it and so went back to this. I will be happy to have a pointer to this
+*/
```
barisere: I guess the problem this comment speaks of has been resolved. Can it be removed?

cli/commands/receive.go
```diff
@@ -62,3 +62,7 @@ func (receiveCommand ReceiveCommand) Run(ctx context.Context, wallet walletcore.
 	termio.PrintTabularResult(termio.StdoutWriter, columns, rows)
 	return nil
 }
+/*
+func (ReceiveCommand) Usage() string {
+	return "Usage: dcrcli receive [account]"
+}*/
```
barisere: Can this comment be removed?

cli/commands/commands.go
```diff
@@ -8,5 +8,5 @@ type Commands struct {
 	Receive         ReceiveCommand         `command:"receive" description:"show your address to receive funds"`
 	History         HistoryCommand         `command:"history" description:"show your transaction history"`
 	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"show details of a transaction"`
-	Help            HelpCommand            `command:"help" description:"show help for a command"`
+	Help            HelpCommand            `command:"help" description:"application help. You can run 'help <command-name>' for command-specific help."`
```
barisere: Let's use the following message.
> run 'help' for general help, or 'help <command-name>' for command-specific help.

barisere (changes requested):

barisere (approved):

barisere:
To take advantage of the work done in this PR, the descriptions (both long and short) on the commands should be updated, but this should be done in a separate pull request.
@itswisdomagain what do you think? Can this be merged?

cli/commands/help.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+type HelpCommand struct {
+	runner.ParserCommand
+	Args struct {
+		CommandName string `positional-arg-name:"command-name"`
+	} `positional-args:"yes"`
+}
+
+func (h HelpCommand) Run(ctx context.Context, parser *flags.Parser, args []string) error {
```
itswisdomagain: We don't need context, you can adjust the definition of the `Run` method in `ParserCommandRunner` interface to remove this parameter

cli/commands/commands.go
```diff
@@ -2,10 +2,11 @@ package commands
...
 type Commands struct {
-	Balance         BalanceCommand         `command:"balance" description:"show your balance"`
+	Balance         BalanceCommand         `command:"balance" description:"shows total balance for each account in wallet" long-description:"Shows total balance for each account in wallet. Also shows spendable balance if different from total balance. Use -d/--detailed to view a more detailed balance breakdown"`
 	Send            SendCommand            `command:"send" description:"send a transaction"`
 	SendCustom      SendCustomCommand      `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
 	Receive         ReceiveCommand         `command:"receive" description:"show your address to receive funds"`
 	History         HistoryCommand         `command:"history" description:"show your transaction history"`
 	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"show details of a transaction"`
+	Help            HelpCommand            `command:"help" description:"shows application help, run 'help' for general help 'help <command-name>' for command-specific help."`
```
itswisdomagain: You already explained that this command "shows application help" (maybe that can be "shows general application help")
No need to repeat that info by saying "run 'help' for general application help", the next info should be "run `help <command-name> for command-specific help

This can be simplified to:

> Shows general application help. Run `help <command-name>` to get help message for a specific command

cli/commands/help.go
```diff
@@ -0,0 +1,44 @@
...
+	if targetCommand == nil {
+		return fmt.Errorf("unknown command %q", h.Args.CommandName)
+	}
+
+	PrintCommandHelp(parser.Name, targetCommand)
+
+	return nil
+}
+
+func PrintCommandHelp(appName string, command *flags.Command) {
```
itswisdomagain, line 35: We can improve this function's behaviour.
Currently,
screenshot 2019-01-01 at 12 01 26 am

The long description should be shown first, before usage. I think we can remove the reference to app options in the usage text for simplicity, having 2 [options] in the usage text can be confusing.

**This is a desired output:**
```
$ godcr balance -h
Show total balance for each account in wallet. Also shows spendable balance if different from total balance
Usage: godcr balance [args] [options]

Arguments:
account-name (required)       Name of account to generate balance for
account-number                Account number of account to generate balance for

Options
-d, --detailed                Display detailed account balance report

Use godcr -h to view application options
```

Achieving this would mean not using the parser.WriteHelp method but rather using `command. ShortDescription ` and `command. LongDescription` for short and long descriptions (no need to include short description in long description anymore), `command.Options` to get the [command options](https://godoc.org/github.com/jessevdk/go-flags#Option) (might be nil), `command.Args()` to get the [command arguments](https://godoc.org/github.com/jessevdk/go-flags#Arg) (might be nil also)

itswisdomagain (changes requested):
We're close, but the critical objective of this PR hasn't been satisfactorily met.

ademuanthony:
help-detailed

cli/commands/help.go
```diff
@@ -0,0 +1,123 @@
...
+	if len(command.LongDescription) > 0 {
```
itswisdomagain: We can avoid repeating the short description text in the long description and instead combine both when printing the description message for the command. That way, commands without long description will show just the short description and repetition is avoided.

`fmt.Printf("%s. %s\n")` does the trick

With this approach, we have to ensure now and going forward, that the both short and long descriptions for commands do not end with a period. Also, both descriptions should start with the first letter of the first word capitalized.

cli/commands/help.go
```diff
@@ -0,0 +1,123 @@
...
+	fmt.Println()
+
+	args := command.Args()
+	options := command.Options()
+	usageText := fmt.Sprintf("Usage: %s %s", appName, command.Name)
+	if args != nil && len(args) > 0 {
+		usageText += " [args]"
+	}
+	if options != nil && len(options) > 0 {
+		usageText += " [options]"
```
itswisdomagain: Always append the `[options]` text because even commands without command-specific options can still make use of general app options. That's why every command help message ends with instruction to view general app options. If there are command-specific options, then they are listed before the instruction to view general app options

cli/commands/commands.go
```diff
@@ -2,10 +2,11 @@ package commands
 
 // Commands defines the commands and options available on the cli
 type Commands struct {
-	Balance         BalanceCommand         `command:"balance" description:"show your balance"`
-	Send            SendCommand            `command:"send" description:"send a transaction"`
-	SendCustom      SendCustomCommand      `command:"send-custom" description:"send a transaction, manually selecting inputs from unspent outputs"`
-	Receive         ReceiveCommand         `command:"receive" description:"show your address to receive funds"`
-	History         HistoryCommand         `command:"history" description:"show your transaction history"`
-	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"show details of a transaction"`
+	Balance         BalanceCommand         `command:"balance" description:"Shows total balance for each account in wallet" long-description:"Shows total balance for each account in wallet. Also shows spendable balance if different from total balance. Use -d/--detailed to view a more detailed balance breakdown"`
```
itswisdomagain: The long description should no longer include text from the short description (see comment below). Also information about command args and options can be omitted as these are well defined in the "Arguments" and "Options" section of the command's help message

cli/commands/commands.go
```diff
@@ -2,10 +2,11 @@ package commands
...
+	Send            SendCommand            `command:"send" description:"Send a transaction"`
+	SendCustom      SendCustomCommand      `command:"send-custom" description:"Send a transaction, manually selecting inputs from unspent outputs"`
+	Receive         ReceiveCommand         `command:"receive" description:"Show your address to receive funds"`
+	History         HistoryCommand         `command:"history" description:"Show your transaction history"`
+	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"Show details of a transaction"`
+	Help            HelpCommand            `command:"help" description:"Shows general application help. Run help <command-name> to get help message for a specific command"`
```
itswisdomagain: I think we should stick to active verbs for command short descriptions: "Show ..." instead of "Shows". It also tallies with the patterns above.

cli/commands/help.go
```diff
@@ -0,0 +1,123 @@
...
+				longestNameLength = argNameLength
+			}
+		}
+
+		for _, arg := range args {
+			required := ""
+			if arg.Required == 1 {
+				required = "(required)"
+			}
+			fmt.Println(fmt.Sprintf("%s %s%s %s", arg.Name, required, generateWhiteSpace(len(arg.Name), longestNameLength), arg.Description))
```
itswisdomagain: You can use tabs (`\t`) with the TabWriter in `termio/writer/go`

itswisdomagain (changes requested):
This is neat! Some thoughts

itswisdomagain:
> To take advantage of the work done in this PR, the descriptions (both long and short) on the commands should be updated, but this should be done in a separate pull request.

Yes, separate PR, to ensure thoroughness.

cli/commands/help.go
ademuanthony: I am unable to use the TabWriter due to formatting issues;
the extra space between the header and the body of the table is not desirable for my use case.

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	PrintCommandHelp(parser.Name, targetCommand)
+
+	return nil
+}
+
+func PrintCommandHelp(appName string, command *flags.Command) {
+	fmt.Printf("%s. %s\n", command.ShortDescription, command.LongDescription)
+	fmt.Println()
```
itswisdomagain: Create the `tabWriter` earlier in this function and use it for all printing. When done, call `tabWriter.flush()`. Use `tabWriter` as variable name, it better describes the purpose the writer serves.

```
tabWriter := termio.StdoutWriter
fmt.Fprintf(tabWriter, "%s. %s\n", command.ShortDescription, command.LongDescription)
fmt.Fprintln(tabWriter)
```

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+
+	args := command.Args()
+	options := command.Options()
```
itswisdomagain: This declaration and assignment can move to line 75, just before it's usage

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+	usageText := fmt.Sprintf("Usage: %s %s", appName, command.Name)
+	if args != nil && len(args) > 0 {
+		usageText += " [args]"
+	}
+	usageText += " [options]"
+	fmt.Println(usageText)
+	fmt.Println()
```
itswisdomagain:
```
fmt.Fprintln(tabWriter, usageText)
fmt.Fprintln(tabWriter)
```

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if args != nil && len(args) > 0 {
+		fmt.Println("Arguments:")
+		longestNameLength := 0
```
itswisdomagain: This variable and the following loop is are longer necessary

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+			// required args takes extra 10 space for printing the text '(required)'
+			if arg.Required == 1 {
+				argNameLength += 10
+			}
+			if argNameLength > longestNameLength {
+				longestNameLength = argNameLength
+			}
+		}
+
+		writer := termio.StdoutWriter
```
itswisdomagain: Use the `tabWriter` created at the top of the function

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+		for _, arg := range args {
+			required := ""
+			if arg.Required == 1 {
+				required = "(required)"
+			}
+			fmt.Fprintln(writer,fmt.Sprintf("%s %s \t %s", arg.Name, required, arg.Description))
+		}
+		writer.Flush()
+		fmt.Println()
```
itswisdomagain: Don't flush `tabWriter` yet, `fmt.Fprintln(tabWriter)`

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if options != nil && len(options) > 0 {
```
itswisdomagain, line 61: I'm having some troubles reading the code in this if block, compare this snippet, more readable?
```
if options != nil && len(options) > 0 {
	fmt.Fprintln(tabWriter, "Options:")

	for _, option := range options {
		var optionUsages []string
		if option.ShortName != 0 {
			optionUsages = append(optionUsages, fmt.Sprintf("-%c", option.ShortName))
		}
		if option.LongName != "" {
			optionUsages = append(optionUsages, fmt.Sprintf("--%s", option.LongName))
		}
		optionText := strings.Join(optionUsages, ", ")

		fmt.Fprintf(tabWriter, "%s \t %s\n", optionText, option.Description)
	}

	fmt.Fprintln(tabWriter)
}
```

cli/commands/help.go
```diff
@@ -0,0 +1,100 @@
...
+			}else {
+				name += " " + " " + " " + " "
+			}
+			fmt.Fprintln(writer,fmt.Sprintf("%s--%s \t %s", name, option.LongName, option.Description))
+		}
+		writer.Flush()
+		fmt.Println()
+	}
+
+	fmt.Println(fmt.Sprintf("Use %s -h to view application options", appName))
```
itswisdomagain:
```
fmt.Fprintf(tabWriter, "Use %s -h to view application options\n", appName)
tabWriter.Flush()
```

itswisdomagain (changes requested):

cli/commands/help.go
ademuanthony: the current version tries to indent the options, giving 2 whitespaces for those with short and 4 for those without. This is an attempt to stay consistent with the output of parser.WriteHelp
It is not nesecary?

itswisdomagain: Ah, I see. It is indeed necessary. Will this work?
```
if options != nil && len(options) > 0 {
	fmt.Fprintln(tabWriter, "Options:")

	for _, option := range options {
		var optionUsages []string
		optionPrefix := "  "
		if option.ShortName != 0 {
			optionUsages = append(optionUsages, fmt.Sprintf("%s-%c", optionPrefix, option.ShortName))
                         optionPrefix = ""
		} else {
                         optionPrefix = "    "
		}
		if option.LongName != "" {
			optionUsages = append(optionUsages, fmt.Sprintf("%s--%s", optionPrefix, option.LongName))
		}
		optionText := strings.Join(optionUsages, ", ")

		fmt.Fprintf(tabWriter, "%s \t %s\n", optionText, option.Description)
	}

	fmt.Fprintln(tabWriter)
}
```

cli/runner/parsercommand.go
```diff
@@ -0,0 +1,22 @@
...
+// flags.Parser can implement to have it injected at run time
+type ParserCommandRunner interface {
+	Run(parser *flags.Parser, args []string) error
+	flags.Commander
+}
+
+// ParserCommand implements `flags.Commander`, using a noop Execute method to satisfy `flags.Commander` interface
+// Commands embedding this struct should ideally implement `ParserCommandRunner` so that their `Run` method can
+// be invoked by `CommandRunner.Run` which will inject the necessary dependencies to run the command
+type ParserCommand struct{}
```
itswisdomagain: The `commanderStub` is back on master, this can go. Use `commands. commanderStub` instead

cli/runner/parsercommand.go
```diff
@@ -0,0 +1,22 @@
+package runner
+
+import (
+	"github.com/jessevdk/go-flags"
+)
+
+// ParserCommandRunner defines the Run method that cli commands that depends on
+// flags.Parser can implement to have it injected at run time
+type ParserCommandRunner interface {
```
itswisdomagain: Add this to `runner/runners.go`

cli/commands/help.go
```diff
@@ -0,0 +1,80 @@
...
+	PrintCommandHelp(parser.Name, targetCommand)
+
+	return nil
+}
+
+func PrintCommandHelp(appName string, command *flags.Command) {
+	tabWriter := termio.StdoutWriter
+	fmt.Fprintln(tabWriter ,fmt.Sprintf("%s. %s\n", command.ShortDescription, command.LongDescription))
+
+	usageText := fmt.Sprintf("Usage: %s %s", appName, command.Name)
```
itswisdomagain: To maintain consistency with parser.WriteHelp as you rightly recommended, the usage text should be printed in 2 lines like this
screenshot 2019-01-01 at 10 33 23 pm

I think `usageText := fmt.Sprintf("Usage:\n  %s %s", appName, command.Name)` does the trick. (2 spaces after the breakline)

cli/commands/help.go
```diff
@@ -0,0 +1,80 @@
...
+	fmt.Fprintln(tabWriter)
+
+	if args != nil && len(args) > 0 {
+		fmt.Fprintln(tabWriter,"Arguments:")
+		for _, arg := range args {
+			required := ""
+			if arg.Required == 1 {
+				required = "(required)"
+			}
+			fmt.Fprintln(tabWriter,fmt.Sprintf("%s %s \t %s", arg.Name, required, arg.Description))
```
itswisdomagain: Also, in keeping consistent with parser.WriteHelp, this should be prefixed with 2 spaces

cli/commands/help.go
```diff
@@ -0,0 +1,80 @@
...
+
+	options := command.Options()
+	if options != nil && len(options) > 0 {
+		fmt.Fprintln(tabWriter,"Options:")
+		// option printout attempts to add 2 whitespace for options with short name and 4 for those without in order to
+		// This is an attempt to stay consistent with the output of parser.WriteHelp
+		for _, option := range options {
+			optionUsage := " " + " "
+			if option.ShortName != 0 {
+				optionUsage += "-" + string(option.ShortName) + ", "
```
itswisdomagain: There'd be a trailing comma on this output if there's no long name (not all config options) will have long name. Consider this (notice the subtle change to the comment too):

```
// option printout attempts to add 2 whitespace for options with short name and 6 for those without
// This is an attempt to stay consistent with the output of parser.WriteHelp
for _, option := range options {
	var optionUsage string

	if option.ShortName != 0 && option.LongName != "" {
		optionUsage = fmt.Sprintf("  -%c, --%s", option.ShortName, option.LongName)
	} else if option.ShortName != 0 {
		optionUsage = fmt.Sprintf("  -%c", option.ShortName)
	} else {
		optionUsage = fmt.Sprintf("      --%s", option.LongName)
	}

	fmt.Fprintln(tabWriter,fmt.Sprintf("%s \t %s", optionUsage, option.Description))
}
```

itswisdomagain (changes requested):
Last lap, I believe. You'll have to rebase your branch with master to be able to effect some of the requested changes

ademuanthony:
23-29-01-help

cli/runner/runners.go
```diff
@@ -20,3 +20,10 @@ type WalletMiddlewareCommandRunner interface {
 	Run(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	flags.Commander
 }
+
+// ParserCommandRunner defines the Run method that cli commands that depends on
+// flags.Parser can implement to have it injected at run time
+type ParserCommandRunner interface {
+	Run(parser *flags.Parser, args []string) error
```
itswisdomagain: We don't really have a need for `args []string`, it can go

itswisdomagain (changes requested):
1 last change, then `go fmt ./...` your code

itswisdomagain (approved):

PR #99, opened 2018-12-31
add createwallet command and cleanup runner package

itswisdomagain:
Running a godcr command without an existing wallet prompts user to create one.
This PR introduces a new `godcr createwallet` command that a user can run at any time to create a wallet.


cli/walletloader/helpers.go
```diff
@@ -0,0 +1,54 @@
...
+	fmt.Println("Please keep in mind that anyone who has access to the seed can also restore your wallet thereby " +
+		"giving them access to all your funds, so it is imperative that you keep it in a secure location.")
+}
+
+func attemptToCreateWallet(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
+	createWalletPrompt := "No wallet found. Would you like to create one now? (y/N)"
+	validateUserResponse := func(userResponse string) error {
+		userResponse = strings.TrimSpace(userResponse)
+		userResponse = strings.Trim(userResponse, `"`)
+		if userResponse == "" || strings.EqualFold("y", userResponse) || strings.EqualFold("N", userResponse) {
```
barisere: - The response from the user should be treated as case-insensitive.
- When using a polar question with y for yes and n for no, the uppercased option specifies the default. Are we encouraging the user to not create a wallet?

barisere: There may be other use cases for requesting yes/no answers from the user. This can be factored into a function in `package termio`.

barisere (changes requested):

cli/walletloader/helpers.go
itswisdomagain: - The response is currently treated as case-insensitve (strings.EqualFold)
- The "no" default was as suggested by @raedah, it made sense too because prior to asking users whether to create a wallet, if there was no wallet, the program just exits with a message asking users to run the `godcr createwallet` command to create a wallet. This maintains that behaviour, by default, the program exits. User can create the wallet manually by running the appropriate command. That said, what other option does the user have than to create a wallet if none exists? "Yes" default makes sense
- Correct, will do

barisere: For the `createwallet` command, "yes" makes sense as default. No would make sense if we we're suggesting owing to the absence of a wallet, but not when to the user is asking to create a wallet.

cli/termio/terminalprompt/terminalprompt.go
```diff
@@ -83,3 +84,45 @@ func RequestSelection(message string, options []string, validate ValidatorFuncti
 		return value, nil
 	}
 }
+
+func RequestYesNoOption(message, defaultOption string) (string, error) {
```
barisere: This function will be used to ask polar questions, so the return string represents only two possible states: the user either agrees or disagrees. Thus, we can use a boolean instead of string as the first return value.
The benefit is that there won't be a need for a string comparison of the user input at the call sites; we know it was affirmative if `true` is returned, negative otherwise.

cli/termio/terminalprompt/terminalprompt.go
```diff
@@ -83,3 +84,45 @@ func RequestSelection(message string, options []string, validate ValidatorFuncti
...
+	if err != nil {
+		return "", err
+	}
+
+	userResponse = strings.TrimSpace(userResponse)
+	if defaultOption != "" && userResponse == "" {
+		return defaultOption, nil
+	}
+
+	return userResponse, nil
```
barisere: `userResponse` here can be either "y", "yes", "n", or "no", based on the logic in `validateUserResponse`. It's easy to miss these if we do string comparison of the user input at the call site. Hence my previous suggestion to use boolean return value.

cli/walletloader/helpers.go
```diff
@@ -29,23 +29,14 @@ func displayWalletSeed(seed string) {
...
-	userResponse, err := terminalprompt.RequestInput(createWalletPrompt, validateUserResponse)
+	createWalletPrompt := "No wallet found. Would you like to create one now?"
+	userResponse, err := terminalprompt.RequestYesNoOption(createWalletPrompt, "Y")
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "Error reading your response: %s", err.Error())
 		return err
 	}
 
-	if userResponse == "" || strings.EqualFold("N", userResponse) {
+	if strings.EqualFold("n", userResponse) {
```
barisere: My previous addresses a possible error on this line.

cli/walletloader/walletloader.go
```diff
@@ -74,18 +74,8 @@ func CreateWallet(ctx context.Context, walletMiddleware app.WalletMiddleware) (e
...
-		if userResponse == "" || strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
-			return nil
-		} else {
-			return fmt.Errorf("invalid option, try again")
-		}
-	}
-
-	syncBlockchainResponse, err := terminalprompt.RequestInput(syncBlockchainPrompt, validateUserResponse)
+	syncBlockchainPrompt := "Would you like to sync the blockchain now?"
+	syncBlockchainResponse, err := terminalprompt.RequestYesNoOption(syncBlockchainPrompt, "Y")
```
barisere: Check the subsequent use of `syncBlockchainResponse`. It should be a boolean, if you've made the change suggested in the preceding comments.

barisere (changes requested):

cli/commands/receive.go
```diff
@@ -48,21 +48,10 @@ func (receiveCommand ReceiveCommand) Run(wallet walletcore.Wallet) error {
...
-		if userResponse == "" || strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
-			return nil
-		} else {
-			return fmt.Errorf("invalid option, try again")
-		}
-	}
-
-	confirm, err := terminalprompt.RequestInput("Would you like to generate a QR code? (y/N)", validateConfirm)
+	// Print out QR code?
+	confirm, err := terminalprompt.RequestYesNoOption("Would you like to generate a QR code?", "N")
```
barisere: See https://github.com/raedahgroup/godcr/pull/99#discussion_r244636630 and the other comments in that review.

barisere (changes requested):

cli/commands/send.go
```diff
@@ -75,20 +74,9 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
-		userResponse = strings.TrimSpace(userResponse)
-		userResponse = strings.Trim(userResponse, `"`)
-		if strings.EqualFold("Y", userResponse) || strings.EqualFold("n", userResponse) {
-			return nil
-		} else {
-			return fmt.Errorf("invalid option, try again")
-		}
-	}
-	confirm, err := terminalprompt.RequestInput("Are you sure? (y/n)", validateConfirm)
+	confirm, err := terminalprompt.RequestYesNoOption("Are you sure?", "")
```
barisere: See https://github.com/raedahgroup/godcr/pull/99#discussion_r244636630

cli/commands/send.go
```diff
@@ -77,20 +74,9 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
```
barisere: See https://github.com/raedahgroup/godcr/pull/99#discussion_r244636630

barisere (changes requested):

barisere (approved):

PR #104, opened 2019-01-01
http send page cleanup

### LINKED ISSUE #57: cleanup http send page
untitled

yaowork:
this performs cleanup that fixes #57


barisere (changes requested):
image
> If only one account is available, it should just be printed and not a widget.
(See https://github.com/raedahgroup/godcr/issues/57#issuecomment-446794830)

- I have only one account, but it is not selected. Clicking the select menu displays nothing.
- There's a checkbox for custom transaction, and a link to "customize transaction". How do they differ?
- The heading for the form can be worded better. "Send From" is not a good description of what the form is meant for.
- The heading for the form should not be a link.

itswisdomagain (changes requested):
You haven't added code to use the value of the "Spend Unconfirmed" checkbox. That checkbox ought to alter the way a transaction is created. If the box is checked, then the value of `required confirmation` for both normal send and custom send should be 0. Otherwise, a higher value should be used to ensure that only **confirmed** transactions are spent.

web/views/balance.html
```diff
@@ -56,5 +56,11 @@ <h3 class="mb-0">{{ simpleBalance $account.Balance false }}</h3>
             </div>
         </div>
     </div>
+    <script>
+        $(function(){
+            // set current nav item 
+            $("#nav-balance").addClass("active");
```
itswisdomagain: Instead of manually setting this in every page, we should have a single code probably in footer that compares the current `window.href` value against each `a.navlink` element to identify which of the nav links is currently visited and mark that as active instead

itswisdomagain (changes requested):
Just one code that caught my attention. Will take another look and test properly after you are done with the changes you're working to add.

web/views/send.html
```diff
@@ -10,84 +10,87 @@
...
-                                            <select class="form-control" id="sourceAccount" name="sourceAccount">
-                                                {{ range $account := .accounts }}
-                                                <option value="{{ $account.Number }}">
-                                                    {{ $account.Name }} - {{ $account.Balance.Total }}</option>
+                    <div class="card z-depth-0 bordered">
+                        <div class="card-body">
+                            <div class="row">
+                                <div class="col-md-12 col-sm-12">
+                                    <div class="form-group">
+                                       {{ $numberOfAccounts := len .accounts }} {{ if eq $numberOfAccounts 1 }}
```
itswisdomagain: Can simplify this to `{{ if eq (len .accounts) 1 }}`

web/views/send.html
```diff
@@ -10,84 +10,87 @@
...
+                                            {{ $account := index .accounts 0 }}
+                                            <div>Source Account</div>
+                                            <strong>{{ $account.Name }} - {{ $account.Balance }} </strong>
```
itswisdomagain: Add some padding or margin, the default account is too close to the header
screenshot 2019-01-06 at 1 55 22 am

itswisdomagain: Also, `$account.Balance` is a struct object, holding sub values
You can instead use `{{ simpleBalance $account.Balance false }}`, a template function defined in line 28 of templates.go to display a properly formatted amount summary

web/views/send.html
```diff
@@ -10,84 +10,87 @@
...
+                                            <div class="d-none">
```
itswisdomagain, line 22: This div can go, we are not showing a dropdown when there's just one account in the wallet, it's not enough to hide it with `class="d-none"`.

web/views/send.html
```diff
@@ -10,84 +10,87 @@
...
+                            </div>
+                            <div class="row">
+                                <div class="col-md-6 col-sm-12">
+                                    <div class="form-group">
+                                        <label for="amount">Amount (DCR)</label>
+                                        <input type="number" class="form-control" id="amount" name="amount" />
+                                    </div> 
+                                    <div class="form-group">
+                                        <input type="checkbox" name="use-custom" id="use-custom" value="1" />
+                                        <label for="use-custom">Select custom inputs</label>
```
itswisdomagain, line 48: - this checkbox should go below destination address
- text should be "Build custom transaction"
- the section showing the utxos should not have a scroll bar within it, it should extend to fill the page and push the send button down
- also notice that the utxo amounts are not showing
screenshot 2019-01-06 at 2 07 08 am

web/views/send.html
```diff
@@ -10,84 +10,87 @@
...
-                                <div class="alert alert-info">Build custom transaction with wallet inputs</div>
-                                <div id="wallet-outputs">
-                                        <table class="table inverse" height="200px">
+                            <div class="row d-none" id="custom-tx-row">
+                                <div class="col-md-12">
+                                    <div class="scrollable">
+                                        <div class="alert alert-info">Select inputs to build custom transaction</div>
+                                        <div class="status">Fetching Unspent Outputs</div>
+                                        <div class="progress">
+                                            <div class="progress-bar" style="width: 0%"></div>
```
itswisdomagain, line 58: This progress bar is not in any specification I'm aware of. Did @raedah request this? If not, it should go.

web/public/js/send.js
```diff
@@ -6,7 +6,7 @@ function validatePassphrase() {
     $(".passphrase-error").remove();
     var error = "";
 
-    var passphraseEl = $("#walletPassphrase");
+    var passphraseEl = $("#wallet-passphrase");
 
     if (passphraseEl.val() == "") {
```
itswisdomagain: Use strict comparison operator `===` and `!==`. The lose comparison operator `==` and `!=` is used in so many places in this file, please find all occurences and correct.

web/public/js/send.js
```diff
@@ -36,26 +36,40 @@ function getWalletPassphraseAndSubmit() {
...
     }
 
     if (destinationAddressEl.val() == "") {
-        errors["#destinationAddress"] = "The destination address is required"
+        errors["#destination-address"] = "The destination address is required"
+    }
+
+    var isCustomTransaction = $("#use-custom").prop("checked")?true:false;
+    if (isCustomTransaction && (getSelectedInputsSum() < amountEl.val())) { 
+        errors["#custom-tx-row .alert-info"] = "The sum of selected inputs is less than send amount";
```
itswisdomagain, line 72: Displaying errors next to the div that has error makes it hard for the error to be noticed. Display errors immediately before or after the "Next" button instead

itswisdomagain (changes requested):
Required adjustments for now. Will be able to test more when these are fixed

web/views/send.html
yaowork: a visual indicator that indicates whether a user has selected "enough" or "not enough" inputs was requested.

web/views/send.html
yaowork: - The checkbox is there as per instructions
- Same with the "Select custom inputs" text

web/views/send.html
yaowork: it is needed in submitting the form

web/views/send.html
itswisdomagain: Okay, how about the scrolling area? And the progress bar? Both specifically required?

If yes to both, then the progress bar should not be inside the scrollable area. When I scroll away from the top of the utxo list, the progress bar is no longer visible.

web/public/js/send.js
```diff
@@ -36,38 +36,62 @@ function getWalletPassphraseAndSubmit() {
...
     }
 
     if (!$.isEmptyObject(errors)) {
         isClean = false;
         for (var i in errors) {
```
itswisdomagain, line 77: `errors` could simply be an array of error messages instead of a key-value pair, since the keys are no longer used.

app/walletcore/types.go
```diff
@@ -26,6 +26,7 @@ type UnspentOutput struct {
 	Tree            int32          `json:"tree"`
 	ReceiveTime     int64          `json:"receive_time"`
 	Amount          dcrutil.Amount `json:"amount"`
+	FormattedAmount float64        `json:"formatted_amount"`
```
itswisdomagain, line 29: This `FormattedAmount` field in the `UnspentOutput` type is really only needed in the web interface. Other interfaces can quite easily get the dcr amount by calling `utxo.Amount.ToCoin()`.

Also, it's really just duplication of data. We have similar duplication in `Time` and `FormattedTime` in some other types that we'll also have to cleanup.

If an interface requires value in a particular format, there's no need to add a new field to the parent type if the value in the desired format can be easily gotten from a field already existing in the parent type.

In this case, you can get the formatted time in the web interface handler that needs this value, no need to have it here. I understand the data is used in JS code and you cannot call `.ToCoin()` in JS, but you can do so in the `getUnspentOutputs` handler before marshalling the data as JSON.

Or you could change `Amount` type to `float64` and modify it's usages in other interfaces. Don't duplicate value.

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -179,13 +179,16 @@ func (c *WalletPRCClient) UnspentOutputs(account uint32, targetAmount int64) ([]
...
+
 		unspentOutput := &walletcore.UnspentOutput{
 			OutputKey:       fmt.Sprintf("%s:%d", txHash, utxo.OutputIndex),
 			TransactionHash: txHash,
 			OutputIndex:     utxo.OutputIndex,
 			Tree:            utxo.Tree,
 			ReceiveTime:     utxo.ReceiveTime,
-			Amount:          dcrutil.Amount(utxo.Amount),
+			Amount:          amount,
+			FormattedAmount: amount.ToCoin(),
```
itswisdomagain, line 191: As explained above, not needed. At the very least, take off either `Amount` or `FormattedAmount`, one of those values can be used to get the other seamlessly.

web/public/js/send.js
```diff
@@ -36,38 +36,62 @@ function getWalletPassphraseAndSubmit() {
...
-    var sourceAccountEl = $("#sourceAccount");
+    var sourceAccountEl = $("#source-account");
     var amountEl = $("#amount");
-    var destinationAddressEl = $("#destinationAddress");
+    var destinationAddressEl = $("#destination-address");
     var isClean = true;
 
-    if (sourceAccountEl.find(":selected").text() == "") {
-        errors["#sourceAccount"] = "The source account is required";
+    if (sourceAccountEl.find(":selected").text() === "") {
```
itswisdomagain, line 58: You should be creative here, if there's only one account in the wallet, `source account` should not be dropdown that requires the `:selected` element selector to determine its value.

web/public/js/send.js
```diff
@@ -36,38 +36,62 @@ function getWalletPassphraseAndSubmit() {
...
         errors["#amount"] = "The amount is required"
     }
 
-    if (destinationAddressEl.val() == "") {
-        errors["#destinationAddress"] = "The destination address is required"
+    if (destinationAddressEl.val() === "") {
+        errors["#destination-address"] = "The destination address is required"
+    }
+
+    var isCustomTransaction = $("#use-custom").prop("checked")?true:false;
```
itswisdomagain, line 70: `$('#use-custom').is(':checked')`

web/public/js/send.js
```diff
@@ -6,13 +6,13 @@ function validatePassphrase() {
...
+    var passphraseEl = $("#wallet-passphrase");
 
-    if (passphraseEl.val() == "") {
+    if (passphraseEl.val() === "") {
         error = "Your wallet passphrase is required";
     }
 
-    if (error != "") {
+    if (error !== "") {
         passphraseEl.after("<div class='passphrase-error'>" + error + "</div>");
```
itswisdomagain, line 16: Why not add this div to the modal directly and if there's an error, you set the error text and show as appropriate. Also the if statement above and this one can be merged.

itswisdomagain: This function can be simplified thus:
```
$(".passphrase-error").remove();

var passphraseEl = $("#wallet-passphrase");

if (passphraseEl.val() === "") {
    passphraseEl.after("<div class='passphrase-error'>Your wallet passphrase is required</div>");
    return false
}

return true
```

As mentioned, would be better (performance-wise) to be add the error div to the modal directly instead of creating the element and appending to the dom each time an error needs to be displayed and removing the element from the dom every time this function is called

web/public/js/send.js
```diff
@@ -139,59 +163,91 @@ function clearMessages() {
...
+    $("#custom-tx-row .progress-bar").css("width", percentage+"%");
 }
 
-function closeTab(tab_tem) {
-    tab_item.trigger("click");
+function openCustomizePanel() {
+    $("#custom-tx-row").removeClass("d-none");
+    resetCustomizePanel();
+   
+    var account_number = $("#source-account").find(":selected").val();
```
itswisdomagain, line 190: You may need a `getSelectedAccount` function that gets the selected account from dropdown or simply displayed text

web/public/js/send.js
```diff
@@ -139,59 +163,91 @@ function clearMessages() {
...
+    var callback = function(txs) {
+        // populate outputs 
+        var utxoHtml = txs.map(tx => {
+            var receiveDateTime = new Date(tx.receive_time * 1000)
+            return  "<tr>" + 
+                        "<td width='5%'><input type='checkbox' class='custom-input' name='tx' value="+ tx.key+" data-amount='" + tx.formatted_amount + "' /></td>" +
+                        "<td width='50%'>" + tx.key + "</td>" + 
+                        "<td width='20%'>" + tx.formatted_amount + "</td>" + 
+                        "<td width='25%'>" + receiveDateTime.toString().split(' ').slice(0,5).join(' '); + "</td>" +
```
itswisdomagain, line 199: This doesn't look good. We may need a `formatTime` js function, for now, `toLocaleString()` or something similarly simple should do

itswisdomagain: If the particular time display format is a requirement from @raedah, you should consider using moment.js

web/public/js/send.js
```diff
@@ -139,59 +163,91 @@ function clearMessages() {
...
         e.preventDefault();
-        getWalletPassphraseAndSubmit()
+        if (validateSendForm()) {
+            getWalletPassphraseAndSubmit(submitSendForm)
+        }
     })
+
+    // set current nav item 
+    $("#nav-send").addClass("active");
 });
```
itswisdomagain, line 252: This JS file needs a lot of cleanup.
- Code should be simplified. Variables properly named (e.g. `destinationAddressElement` instead of `destinationAddressEl`).
- The functions should be arranged in order of use. `validatePassphrase` should likely not be the first function in this file.
- I notice some potentially unused functions such as `clearMessages()`, confirm and delete unused code.
- Add more comments.
- You should probably make more use of the `setErrorMessage` function in displaying error messages.
- And other considerations I may not be able to point out at this time.

It's worth remembering we want other devs to be able to maintain this code, not just you.

web/views/send.html
itswisdomagain: That's a hack, you should find a better way to solve the issue with form submission. One way is to use a single hidden input field.
`<input type="hidden" id="source-account" name="source-account" value="{{ $account.Number }}">`

Also, since you're submitting the form with JavaScript, you might not even need a hidden field. As introduced in this PR, there are 2 areas where the selected source account could display: as a simple text (if there's only 1 account in the wallet) and as a dropdown (if there are more 2 or more accounts). Your JS code for validating and submitting the send form should check either places for the presence of a selected account value, not resolving to hacks and work arounds.

web/views/send.html
```diff
@@ -10,84 +10,88 @@
...
```
itswisdomagain, line 58: The progress bar and the `Select inputs to build custom transaction` prompt should not be inside the scrollable area. When I scroll away from the top of the utxo list, they are no longer visible

itswisdomagain (changes requested):
Some more cleanup required.

itswisdomagain:
Other observations.

The error message and the button look off, they should probably be in the card. And some spacing should separate them.
screenshot 2019-01-06 at 6 05 03 pm

Repeated error messages:
screenshot 2019-01-06 at 6 34 22 pm

itswisdomagain:
The git history for this PR branch is polluted and no longer in sync with `raedahgroup/master` because of a recent history re-write done on master.

Follow the steps below to re-create this PR
- Keep a copy of this PR branch's code in some folder in your PC for reference
- Delete your forked repo (codemaestro64/godcr)
- Fork https://github.com/raedahgroup/godcr again
- Clone your newly forked repo
- Create a new branch for your work (`git checkout -b webui-cleanup`)
- Re-apply the changes you made on this PR to the newly created branch
- Push the new branch and create a new PR
- Delete the copy of this code you saved initially

PR #87, opened 2018-12-29
[WIP] Add function to fetch tickets. Add tickets page on web ui

yaowork:
This creates a function in "walletrpcclient" that fetches all tickets. The web ui utilizes this in displaying the ticket info. 

This is a partial fix for #65 
screencapture-localhost-1234-tickets-2018-12-29-13_19_53

Issue #65, opened 2018-12-17
add staking menu


raedah:
Show count of tickets. Sum of the below numbers.

getstakeinfo
```
"ownmempooltix": 2,
"immature": 6,
"live": 138,
```

Show a list of all tickets with their current status.
Valid status are Mempool, Immature, Live, Voted, Revoked.

raedah:
Related, make purchaseticket cmd available.

PR #118, opened 2019-01-02
Staking menu cli

barisere:
Partial solution to #65 
This PR provides functionality for `godcr` in CLI mode: purchasing tickets, and viewing stake information for the wallet.
It introduces two new CLI commands, `purchaseticket` and `stakeinfo`.
As follows, `purchaseticket` displays the hashes of the tickets just purchased, if successful.
```
$ ./godcr purchasetickets
Spending Passphrase: 
You have purchased 1 ticket(s)
3d36436937e01d264281144bb7dffd616d94467d965b63c6430b0708915f9d10
```
`stakeinfo` prints the tickets in the wallet, as follows.
```
$ ./godcr stakeinfo
stake info for wallet:
expired 0  immature 0  live 0  revoked 11  unmined 1  unspent 4  allmempooltix 12  poolsize 5094  missed 15  voted 0  total subsidy 0
```
This PR does not provide these features in web mode; #87 is meant for that.


app/walletcore/types.go
```diff
@@ -85,3 +85,29 @@ type TransactionDetails struct {
 	Outputs       []*TxOutput `json:"outputs"`
 	*Transaction
 }
+
+// StakeInfo holds ticket information summary related to the wallet.
+type StakeInfo struct {
+	Total   uint32     `json:"total"`
+	Tickets []Ticket `json:"tickets"`
```
itswisdomagain: go fmt

itswisdomagain: maybe use pointer for the tickets slice

app/walletcore/types.go
```diff
@@ -85,3 +85,29 @@ type TransactionDetails struct {
...
+}
+
+// Ticket represents a ticket related to the wallet.
+type Ticket struct {
+	Hash   string
+	Status string
+}
+
+type PurchaseTicketRequest struct {
```
itswisdomagain: We can and should use `dcrlibwallet.PurchaseTicketsRequest` instead. One objective of making changes to the `dcrlibwallet` library is to minimise code repetition. Several functions that were once in this godcr codebase have been migrated to the `dcrlibwallet` library to allow code reuse (`txhelper.NewUnsignedTx` for example)

app/walletcore/wallet.go
```diff
@@ -1,5 +1,6 @@
 package walletcore
 
+import "context"
```
itswisdomagain: group imports

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -2,10 +2,13 @@ package dcrlibwallet
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"sort"
 	"time"
 
+	"github.com/raedahgroup/dcrlibwallet"
+
```
itswisdomagain: space can go

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -227,3 +230,67 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
 	return nil, fmt.Errorf("not implemented")
 }
+
+func (lib *DcrWalletLib) StakeInfo(ctx context.Context) (*walletcore.StakeInfo, error) {
+	data, err := lib.walletLib.StakeInfo()
+	if err != nil {
+		return nil, fmt.Errorf("error getting stake info: %v", err.Error())
+	}
+
+	total := data.OwnMempoolTix + data.Live + data.Immature
```
itswisdomagain: Also return the number of tickets in mempool, that are live and that are immature. Aside from showing the individual tickets and their statuses, the `godcr getstakeinfo` command should show a nice summary of the status of the wallet's tickets, not just the total.  Something like:
```
Stake Info
Total Tickets:    20
In mempool:       5
Live:             15

Tickets
hash             status
```

No need to show an entry if there are zero tickets of that nature, in the sample above, "Immature" was ommitted from the output because there are no immature tickets

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -227,3 +230,67 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	stakeInfo := &walletcore.StakeInfo{Total: total}
+
+	responseCh, errorCh, err := lib.walletLib.GetTickets(&dcrlibwallet.GetTicketsRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("error getting stake info: %v", err.Error())
+	}
+
+	for {
+		select {
+		case err := <-errorCh:
```
itswisdomagain: also listen for `ctx.Done` in case user sends interrupt signal. `main.go` catches such signals and cancels the context here as long as the context here is the same context created in `main.go` for long running operations. That context is usually passed to all interfaces and interfaces can in turn pass it to functions performing long operations

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -227,3 +230,67 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	}
+	response, err := lib.walletLib.PurchaseTickets(&dcrlibwallet.PurchaseTicketsRequest{
+		Account:               request.FromAccount,
+		Expiry:                request.Expiry,
+		NumTickets:            request.NumTickets,
+		Passphrase:            request.Passphrase,
+		PoolAddress:           request.PoolAddress,
+		PoolFees:              request.PoolFees,
+		RequiredConfirmations: request.MinConfirmations,
+		SpendLimit:            int64(sendAmount),
```
itswisdomagain: `request.SpendLimit` was already int64, conversion above and reverse conversion here unnecessary

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -227,3 +230,67 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	if err != nil {
+		return nil, fmt.Errorf("error purchasing tickets: %s", err.Error())
+	}
+	ticketHashes = make([]string, len(response))
+	for i, ticketHash := range response {
+		hash, err := chainhash.NewHash(ticketHash)
+		if err != nil {
+			return ticketHashes, fmt.Errorf("error purchasing tickets: %s", err.Error())
+		}
+		ticketHashes[i] = hash.String()
```
itswisdomagain: we should probably modify `lib.walletLib.PurchaseTickets` to return []string instead

itswisdomagain: I'll do this, push to dcrlibwallet repo and provide you with the latest version for use in go.mod to access the updated function

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -13,10 +13,10 @@ import (
 	"google.golang.org/grpc/credentials"
 )
 
-// WalletPRCClient implements `WalletMiddleware` using `dcrwalletrpc` as medium for connecting to a decred wallet
+// WalletRPCClient implements `WalletMiddleware` using `mobilewallet.LibWallet` as medium for connecting to a decred wallet
 // Functions relating to operations that can be performed on a wallet are defined in `walletfunctions.go`
 // Other wallet-related functions are defined in `walletloader.go`
-type WalletPRCClient struct {
+type WalletRPCClient struct {
```
itswisdomagain, line 19: Nice catch

app/walletmediums/dcrwalletrpc/helpers.go
```diff
@@ -27,7 +27,7 @@ func (c *WalletPRCClient) unspentOutputStream(account uint32, targetAmount int64
 	return c.walletService.UnspentOutputs(context.Background(), req)
 }
 
-func (c *WalletPRCClient) signAndPublishTransaction(serializedTx []byte, passphrase string) (string, error) {
+func (c *WalletRPCClient) signAndPublishTransaction(serializedTx []byte, passphrase string) (string, error) {
 	ctx := context.Background()
```
itswisdomagain, line 28: Perhaps, dcrwallet rpc function calls should use the context created in `main.go`, the context passed as parameter to different rpc function calls are used by those functions to know when to terminate long-running processes - which is an expected behaviour when user sends SIGINT

Likely work for a different PR

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -340,3 +340,79 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+
+	totalTickets := stakeInfoResponse.GetOwnMempoolTix() + stakeInfoResponse.GetLive() + stakeInfoResponse.GetImmature()
+
+	ticketsResponse, err := c.walletService.GetTickets(ctx, &walletrpc.GetTicketsRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	stakeInfo := &walletcore.StakeInfo{Tickets: make([]walletcore.Ticket, 0), Total: totalTickets}
+	for response, err := ticketsResponse.Recv(); err != io.EOF; response, err = ticketsResponse.Recv() {
```
itswisdomagain: Err... maybe simplify this? Took me a moment to realise this is the `for initializer; condition; next` syntax used in familiar operations like `for i := 0; i < 10; i++`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -340,3 +340,79 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	}
+	response, err := c.walletService.PurchaseTickets(ctx, &walletrpc.PurchaseTicketsRequest{
+		Account:               request.FromAccount,
+		Expiry:                request.Expiry,
+		NumTickets:            request.NumTickets,
+		Passphrase:            request.Passphrase,
+		PoolAddress:           request.PoolAddress,
+		PoolFees:              request.PoolFees,
+		RequiredConfirmations: request.MinConfirmations,
+		SpendLimit:            int64(amount),
```
itswisdomagain: The conversion above and reverse conversion here are unnecessary

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -340,3 +340,79 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+func (c *WalletRPCClient) PurchaseTicket(ctx context.Context, request walletcore.PurchaseTicketRequest) (ticketHashes []string, err error) {
+	amount, err := dcrutil.NewAmount(float64(request.SpendLimit))
+	if err != nil {
+		return nil, fmt.Errorf("invalid amount for spend limit: %s", err.Error())
+	}
+	priceResponse, err := c.walletService.TicketPrice(ctx, &walletrpc.TicketPriceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())
+	}
+	ticketPrice := dcrutil.Amount(priceResponse.GetTicketPrice())
```
itswisdomagain: `priceResponse.GetTicketPrice()` is int64, can be compared directly with `request.SpendLimit`, no need for converting both to dcrutil.Amount before comparing

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -340,3 +340,79 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	if amount < ticketPrice {
+		return nil, fmt.Errorf("insufficient funds: spend limit %v is less that ticket ticketPrice %v", amount, ticketPrice)
+	}
+	balance, err := c.AccountBalance(request.FromAccount)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account: %v", err.Error())
+	}
+	if balance.Spendable < ticketPrice {
```
itswisdomagain: There's a tiny issue with this comparison. `balance` was gotten from `c.AccountBalance`, which uses 0 required confirmations to fetch the account's balance. This purchase ticket request might want to use a higher value for required confirmations (`request.MinConfirmations > 0`). If that happens, `balance.Spendable` (gotten with 0 required confirmations) might be sufficient to cover `ticketPrice` but `balance.Spendable` (gotten with `request.MinConfirmations` required confirmations) might not.

It's time to modify the `wallet.AccountBalance` function to allow passing a required confirmations parameter.

cli/commands/commands.go
```diff
@@ -15,9 +15,11 @@ type AvailableCommands struct {
 	History         HistoryCommand         `command:"history" description:"Show your transaction history"`
 	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"Show details of a transaction"`
 	Help            HelpCommand            `command:"help" description:"Show general application help. Run help <command-name> to get help message for a specific command"`
+	StakeInfo       GetStakeInfoCommand    `command:"getstakeinfo" description:"Show information about the wallet stakes, tickets and their statuses"`
+	PurchaseTicket  PurchaseTicketCommand  `command:"purchaseticket" description:"Purchase a ticket"`
 }
 
-// AvailableCommands defines experimental commands and options available on the cli
+// ExperimentalCommands defines experimental commands and options available on the cli
```
itswisdomagain, line 22: 😆

cli/commands/commands.go
```diff
@@ -15,9 +15,11 @@ type AvailableCommands struct {
 	History         HistoryCommand         `command:"history" description:"Show your transaction history"`
 	ShowTransaction ShowTransactionCommand `command:"show-transaction" description:"Show details of a transaction"`
 	Help            HelpCommand            `command:"help" description:"Show general application help. Run help <command-name> to get help message for a specific command"`
+	StakeInfo       GetStakeInfoCommand    `command:"getstakeinfo" description:"Show information about the wallet stakes, tickets and their statuses"`
```
itswisdomagain: Change command name to `stakeinfo` (discussed with @raedah); and command type to `StakeInfoCommand`

I think `show-transaction` should probably also be `transaction`, no need to prefix with verbs like `gethistory`, `getbalance`, `show-transaction`, `gethelp`.

`show-transaction` may be an exception though. `godcr transaction [hash]` versus `godcr show-transaction [hash]`, the latter is nicer.

cli/commands/getstakeinfo.go
```diff
@@ -0,0 +1,29 @@
+package commands
+
+import (
+	"context"
+	"fmt"
+	"github.com/raedahgroup/godcr/app"
```
itswisdomagain: separate imports

cli/commands/getstakeinfo.go
```diff
@@ -0,0 +1,29 @@
...
+	"strings"
+)
+
+// GetStakeInfoCommand requests statistics about the wallet stakes.
+type GetStakeInfoCommand struct {
+	commanderStub
+}
+
+// Run displays information about wallet sakes, tickets and their statuses.
+func (g GetStakeInfoCommand) Run(ctx context.Context, wallet app.WalletMiddleware) error {
```
itswisdomagain: This `Run` method should implement `WalletCommandRunner` so that the wallet is opened before getting here. You can add `ctx context.Context` to the signature of `WalletCommandRunner.Run`

itswisdomagain (changes requested):

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +250,67 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+				Hash: response.Ticket.Ticket.Hash.String(),
+				Status: response.TicketStatus.String(),
+			})
+		}
+	}
+
+	return stakeInfo, nil
+}
+
+func (lib *DcrWalletLib) PurchaseTicket(ctx context.Context, request walletcore.PurchaseTicketRequest) (ticketHashes []string, err error) {
```
itswisdomagain: This function implementation for dcrwalletrpc performs a check for account balance and ticket price before proceeding. We should do that here also. Will add the `GetTicket` price function to dcrlibwallet and provide the updated dep version

cli/commands/getstakeinfo.go
```diff
@@ -0,0 +1,29 @@
...
+	stakeInfo, err := wallet.StakeInfo(ctx)
+	if err != nil {
+		return err
+	}
+	output := strings.Builder{}
+	output.WriteString(fmt.Sprintf("Tickets: %d\n", stakeInfo.Total))
```
itswisdomagain: When using dcrlibwallet as wallet medium, `stakeInfo` might be nil even if `err != nil`. That would be similar to `stakeInfo.Total == 0` in the case of dcrwalletrpc. In either cases, display a simple status message: `No tickets in wallet`

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,58 @@
...
+	NumTickets       uint32  `long:"num-tickets" default:"1" description:"The number of tickets to purchase."`
+	PoolAddress      string  `long:"pool-address" description:"The address of the stake pool used. Pool mode will be disabled if an empty string is passed."`
+	PoolFees         float64 `long:"pool-fees" description:"The stake pool fees amount." long-description:"This must be set to a positive value in the allowed range of 0.01 to 100.00 to be valid. It must be set when the pool-address is also set."`
+	Expiry           uint32  `long:"expiry" default:"0" description:"The height at which the tickets expire and can no longer enter the blockchain. It defaults to 0 (no expiry)."`
+	TxFee            int64   `long:"tx-fee" description:"Fees per kB to use for the transaction generating outputs to use for buying tickets." long-description:"If 0 is passed, the global value for a transaction fee will be used."`
+	TicketFee        int64   `description:"Fees per kB to use for all purchased tickets." long-description:"If 0 is passed, the global value for a ticket fee will be used."`
+	Args             struct {
```
itswisdomagain: Maybe put this struct in a different type (PurchaseTicketCommandArgs) and reference here

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,58 @@
...
+		FromAccount string `required:"yes" positional-arg-name:"from account"`
```
itswisdomagain: Just "account"?

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,58 @@
...
+		SpendLimit  int64  `required:"yes" positional-arg-name:"spend limit"`
```
itswisdomagain: "spend-limit" to keep with pattern in receive command and to make for nicer output in `godcr help purchaseticket`

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,58 @@
...
+	} `positional-args:"yes"`
+}
+
+func (ptc PurchaseTicketCommand) Run(ctx context.Context, middleware app.WalletMiddleware) error {
```
itswisdomagain: As with `StakeInfoCommand`, this should implement `WalletCommandRunner`

go.mod
```diff
@@ -2,22 +2,40 @@ module github.com/raedahgroup/godcr
...
+	github.com/decred/dcrd/chaincfg v1.2.1
 	github.com/decred/dcrd/chaincfg/chainhash v1.0.1
+	github.com/decred/dcrd/dcrec v0.0.0-20181212224710-c6f60e2c101c // indirect
+	github.com/decred/dcrd/dcrec/edwards v0.0.0-20181212224710-c6f60e2c101c // indirect
 	github.com/decred/dcrd/dcrutil v1.2.0
 	github.com/decred/dcrd/hdkeychain v1.1.1
 	github.com/decred/dcrd/txscript v1.0.2
 	github.com/decred/dcrd/wire v1.2.0
 	github.com/decred/dcrdata v2.1.3+incompatible
-	github.com/decred/dcrwallet v1.2.3-0.20181120205657-8690f1096aa7
```
itswisdomagain: Retain this version of dcrwallet for this import and for `dcrwallet/rpc/walletrpc` below. `go mod tidy` afterwards, most of the golang.org/x/ dependencies should no longer be needed

go.mod
```diff
@@ -2,22 +2,40 @@ module github.com/raedahgroup/godcr
...
+	github.com/decred/dcrwallet/wallet v1.1.2 // indirect
+	github.com/decred/dcrwallet/walletseed v1.0.1
 	github.com/go-chi/chi v3.3.3+incompatible
+	github.com/gorilla/websocket v1.4.0 // indirect
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/mdp/qrterminal v1.0.1
-	github.com/raedahgroup/dcrlibwallet v1.0.0-rc1.0.20190102181321-9f59fe9537892f128c0855149f60d164600829af
+	github.com/onsi/ginkgo v1.7.0 // indirect
+	github.com/onsi/gomega v1.4.3 // indirect
+	github.com/raedahgroup/dcrlibwallet v1.0.0-rc1.0.20190102215219-497892e692815b32cbd80623373fcaa91394d7bb
```
itswisdomagain: Will give you a new version for this dep. The replace directive at the end of the file can be removed then

itswisdomagain (changes requested):

app/walletcore/types.go
barisere: > maybe use pointer for the tickets slice

It holds only simple scalars that won't likely be modified. Making a copy of it is cheap, if we need to modify it.

go.mod
itswisdomagain: `github.com/raedahgroup/dcrlibwallet v0.0.0-20190103100403-47981b2244f5`

That version has `TicketPrice` and `PurchaseTicket` returns hashes as string slice

app/walletcore/types.go
barisere: There'll be repetition nonetheless. `dcrlibwallet.PurchaseTicketsRequest` and `walletrpc.PurchaseTicketsRequest` are not the same type. But I've made the change.

app/walletmediums/dcrlibwallet/walletfunctions.go
barisere: It wasn't an `Amount`, so the purchase failed due to insufficient balance.

app/walletmediums/dcrwalletrpc/walletfunctions.go
barisere: https://github.com/raedahgroup/godcr/pull/118#discussion_r244964200

app/walletmediums/dcrwalletrpc/walletfunctions.go
barisere: https://github.com/raedahgroup/godcr/pull/118#discussion_r244964200

app/walletcore/types.go
itswisdomagain: Yes, as much as we want to avoid repetition, I'm skeptical about using `walletrpc` types in godcr because dcrwalletrpc is out of our control. `dcrlibwallet` isn't. At least for now, and hopefully never, seeing as 1 major goal of dcrlibwallet is to abstract features across `dcrandroid`, `dcrios,` and `godcr`

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,69 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+		return nil, fmt.Errorf("error getting stake info: %v", err.Error())
+	}
+
+	total := data.OwnMempoolTix + data.Live + data.Immature + data.Unspent
+	stakeInfo := &walletcore.StakeInfo{
+		Immature:      data.Immature,
+		Live:          data.Live,
+		OwnMempoolTix: data.OwnMempoolTix,
+		Total:         total,
+		Unspent:       data.Unspent,
```
itswisdomagain, line 294: Not sure this is needed. As the wallet rpc doc says, `Unspent` could include live, missed or expired tickets, adding live tickets to unspent could lead to an incorrect total.

The issue description only mentions displaying count for mempool, live and immature tickets

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,69 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case err := <-errorCh:
+			return nil, err
+		case response, ok := <-responseCh:
+			if !ok {
+				break
+			}
+			stakeInfo.Tickets = append(stakeInfo.Tickets, walletcore.Ticket{
```
itswisdomagain: Adjusted requirement from @raedah says the list of individual tickets is no longer needed with the `StakeInfo` command. I think this information is already provided with the `GetTransaction` command but if not, we can add it in a different PR

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,69 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+				Status: response.TicketStatus.String(),
+			})
+		}
+	}
+
+	return stakeInfo, nil
+}
+
+func (lib *DcrWalletLib) PurchaseTicket(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) (ticketHashes []string, err error) {
+	sendAmount, err := dcrutil.NewAmount(float64(request.SpendLimit))
```
itswisdomagain: This is wrong conversion. `request.SpendLimit` which represents amount in atoms. To convert that to dcutil.Amount, you use `dcrutil.Amount(request.SpendLimit)`. Still not sure the conversion is needed though, the value can be used directly

itswisdomagain: To emphasize, if `request.SpendLimit == 1`, converting it to float will produce 1.0, and `dcrutil.NewAmount(1.0)` is equal to 1 DCR, not 1 Atom which is what `request.SpendLimit == 1` represents. Maybe we need to update the property name to `SpendLimitInAtom` or change the type to `dcutil.Amount` which is an int64 alias?

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,69 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	}
+	request.SpendLimit = int64(sendAmount)
+
+	ticketPrice, err := lib.walletLib.TicketPrice(ctx)
+	if err != nil {
+		return nil, err
+	}
+	if request.SpendLimit < ticketPrice.TicketPrice {
+		return nil, fmt.Errorf("insufficient funds: spend limit %v is less that ticket ticketPrice %v",
+			request.SpendLimit, ticketPrice)
```
itswisdomagain: Both `request.SpendLimit` and `ticketPrice` are int64, might be more informative to the user to convert to dcr amount string using `dcrutil.Amount(ticketPrice).String()` and use `%s` instead of `%v` in both cases

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,69 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	}
+	response, err := lib.walletLib.PurchaseTickets(&request)
+	if err != nil {
+		return nil, fmt.Errorf("error purchasing tickets: %s", err.Error())
+	}
+	ticketHashes = make([]string, len(response))
+	for i, ticketHash := range response {
```
itswisdomagain: you could just assign `lib.walletLib.PurchaseTickets` result to `ticketHashes`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,94 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+
+func (c *WalletRPCClient) PurchaseTicket(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) (ticketHashes []string, err error) {
+	amount, err := dcrutil.NewAmount(float64(request.SpendLimit))
+	if err != nil {
+		return nil, fmt.Errorf("invalid amount for spend limit: %s", err.Error())
+	}
+	request.SpendLimit = int64(amount)
+	priceResponse, err := c.walletService.TicketPrice(ctx, &walletrpc.TicketPriceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())
```
itswisdomagain: "could not determine ticket price, %s"

app/walletmediums/dcrlibwallet/walletfunctions.go
itswisdomagain: Typo too, "less than", and "ticket price" instead of "ticket ticketPrice"

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,94 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	}
+	ticketPrice := priceResponse.GetTicketPrice()
+	if request.SpendLimit < ticketPrice {
+		return nil, fmt.Errorf("insufficient funds: spend limit %v is less that ticket ticketPrice %v", amount, ticketPrice)
```
itswisdomagain: spend limit doesn't mean user does not have more funds in their wallet..., compare the message displayed in dcrlibwallet's version of this function

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,94 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	}
+	balance, err := c.AccountBalance(request.Account)
```
itswisdomagain: Account balance should be retrieved with min confirmations equal to `request.RequiredConfirmations`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,94 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account: %v", err.Error())
+	}
+	if balance.Spendable < dcrutil.Amount(ticketPrice) {
```
itswisdomagain: Similar comparison is not done in dcrlibwallet's version of this function

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,61 @@
...
+	PoolFees         float64            `long:"pool-fees" description:"The stake pool fees amount." long-description:"This must be set to a positive value in the allowed range of 0.01 to 100.00 to be valid. It must be set when the pool-address is also set."`
+	Expiry           uint32             `long:"expiry" default:"0" description:"The height at which the tickets expire and can no longer enter the blockchain. It defaults to 0 (no expiry)."`
+	TxFee            int64              `long:"tx-fee" description:"Fees per kB to use for the transaction generating outputs to use for buying tickets." long-description:"If 0 is passed, the global value for a transaction fee will be used."`
+	TicketFee        int64              `description:"Fees per kB to use for all purchased tickets." long-description:"If 0 is passed, the global value for a ticket fee will be used."`
+	Args             PurchaseTicketArgs `positional-args:"yes"`
+}
+
+type PurchaseTicketArgs struct {
+	Account    string `required:"yes" positional-arg-name:"from-account"`
```
itswisdomagain: New suggestion from @raedah: this shouldn't be required, use default account (account number 0) by default if no value is passed for this argument

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,61 @@
...
+	SpendLimit int64  `required:"yes" positional-arg-name:"spend-limit"`
```
itswisdomagain: Another comment from @raedah: this is obsolete, should no longer be requested from the user. Use the `ticketPrice` value directly instead.

The dcrwalletrpc docs explain that the spend limit is

> The maximum amount to pay for a single ticket. If the current stake difficulty _(aka ticket price)_ is above this amount, the wallet will return an error

Since the spend limit cannot be lower than ticket price and has no meaning if it is higher (the price for a single ticket is constant and equal to ticketPrice), we can safely use the ticketPrice in place of this value gotten from the user

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+	account, err := wallet.AccountNumber(ptc.Args.Account)
+	if err != nil {
+		return err
+	}
+	tickets, err := wallet.PurchaseTicket(ctx, dcrlibwallet.PurchaseTicketsRequest{
+		TxFee:                 ptc.TxFee,
+		TicketFee:             ptc.TicketFee,
+		TicketAddress:         ptc.TicketAddress,
+		SpendLimit:            ptc.Args.SpendLimit,
```
itswisdomagain: No longer required. I will modify the `dcrlibwallet.PurchaseTicketsRequest` struct to no longer include this field. Also when submitting the actual purchase request to the wallet, I'll modify the `dcrlibwallet.PurchaseTicket` function to use the `TicketPrice()` value instead.

itswisdomagain: Will reply this comment with a new dcrlibwallet dep version to use to access the changes

cli/commands/stakeinfo.go
```diff
@@ -0,0 +1,44 @@
...
+	commanderStub
+}
+
+// Run displays information about wallet stakes, tickets and their statuses.
+func (g StakeInfoCommand) Run(ctx context.Context, wallet app.WalletMiddleware) error {
+	stakeInfo, err := wallet.StakeInfo(ctx)
+	if err != nil {
+		return err
+	}
+	if stakeInfo == nil {
```
itswisdomagain, line 22: `stakeInfo.Total` could also be equal to 0, doesn't that mean there is no ticket in wallet? If that's not the meaning, then we may need to reword the term `Total` (maybe `Active`?), I think total should be a count of ALL tickets in wallet regardless of their statuses

cli/runner/runner.go
```diff
@@ -51,20 +52,39 @@ func (runner CommandRunner) Run(parser *flags.Parser, command flags.Commander, a
...
+
+	return commandRunner.Run(runner.ctx, runner.walletMiddleware)
+}
+
+// processWalletCommand handles command execution for commands requiring access to the app.WalletMiddleware
+// Such commands must satisfy `WalletMiddlewareCommandRunner`.
+// The wallet is opened using the provided walletMiddleware, sync operations performed (if requested)
+// then, the command is executed using the Run method of the WalletMiddlewareCommandRunner interface
+func (runner CommandRunner) processWalletMiddlewareCommand(commandRunner WalletMiddlewareCommandRunner, options config.CliOptions) error {
+	if err := prepareWallet(runner.ctx, runner.walletMiddleware, options); err != nil {
```
itswisdomagain: Commands implementing `WalletMiddlewareCommandRunner` should ideally not **require** that the wallet be pre-opened. An example is the `createwallet` command, it certainly does not need a wallet to be open prior to running and probably should even fail if a wallet already exists.

With this change, `createwallet` will fail if a wallet doesn't exist. Maybe we should rename the type but commands requiring wallet to be pre-opened should use the `WalletCommandRunner` interface, not this one

itswisdomagain (changes requested):

cli/commands/stakeinfo.go
itswisdomagain: New feedback from @raedah: `When in doubt, return all available information and we can prune it in review.`

cli/commands/purchase-ticket.go
itswisdomagain: `github.com/raedahgroup/dcrlibwallet v0.0.0-20190104023021-65f2f2be4b06`

app/walletmediums/dcrlibwallet/walletfunctions.go
barisere: wallet rpc returns 0 for Live, and returns the expected value for live as `Unspent`.
```
*walletrpc.StakeInfoResponse {PoolSize: 0, AllMempoolTix: 0, OwnMempoolTix: 0, Immature: 0, Live: 0, Voted: 0, Missed: 0, Revoked: 5, Expired: 0, TotalSubsidy: 0, Unspent: 7}
```
```
$ ./godcr stakeinfo
stake info for wallet:
total 7  revoked 5  unspent 7
```

app/walletcore/wallet.go
```diff
@@ -46,5 +51,13 @@ type Wallet interface {
 	// TransactionHistory
 	TransactionHistory() ([]*Transaction, error)
 
+	// GetTransaction returns information about the transaction witht the given hash.
```
itswisdomagain: *with* typo

app/walletcore/wallet.go
```diff
@@ -46,5 +51,13 @@ type Wallet interface {
...
+	// An error is returned if the no transaction with the given hash is found.
 	GetTransaction(transactionHash string) (*TransactionDetails, error)
+
+	// StakeInfo returns information about wallet stakes, tickets and their statuses.
+	StakeInfo(ctx context.Context) (*StakeInfo, error)
+
+	// PurchaseTicket is used to purchase tickets.
+	PurchaseTicket(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) (ticketHashes []string, err error)
```
itswisdomagain: Since this command supports purchasing more than 1 or more tickets, won't a better name be `purchasetickets`?

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -230,7 +232,7 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
 			Amount:        dcrutil.Amount(tx.Amount),
 			Fee:           dcrutil.Amount(tx.Fee),
 			Type:          tx.Type,
-			Direction:     txDirection(tx.Direction),
+			Direction:     txDirection(int32(tx.Direction)),
```
itswisdomagain: Let's replace `walletcore.TransactionDirection` with `txhelper.TransactionDirection`. The `txDirection()` declared in line 220 above would no longer be necessary

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,42 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+		Unspent:       data.Unspent,
+	}
+
+	return stakeInfo, nil
+}
+
+func (lib *DcrWalletLib) PurchaseTicket(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) ([]string, error) {
+	balance, err := lib.AccountBalance(request.Account)
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account: %v", err.Error())
```
itswisdomagain: - `err.Error()` is of type string, use `%s` instead of `%v`
- message should be could not fetch `account balance`, not `account`
- use the value of `request.RequiredConfirmations` in the call to fetch account balance so your comparison below will be valid

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -247,3 +249,42 @@ func (lib *DcrWalletLib) TransactionHistory() ([]*walletcore.Transaction, error)
...
+	}
+
+	ticketPrice, err := lib.walletLib.TicketPrice(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if balance.Spendable < dcrutil.Amount(ticketPrice.TicketPrice) {
+		return nil, fmt.Errorf("insufficient funds: account balance %v is less than ticket price %v",
+			balance.Spendable, ticketPrice.TicketPrice)
```
itswisdomagain: `return nil, fmt.Errorf("insufficient funds: spendable account balance (%s) is less than ticket price (%s)", balance.Spendable, dcrutil.Amount(ticketPrice.TicketPrice))`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -7,6 +7,8 @@ import (
 	"io"
 	"sort"
 
+	"github.com/raedahgroup/dcrlibwallet"
+
```
itswisdomagain: space can go

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,64 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
 	}, nil
 }
+
+func (c *WalletRPCClient) StakeInfo(ctx context.Context) (*walletcore.StakeInfo, error) {
+	stakeInfo, err := c.walletService.StakeInfo(ctx, &walletrpc.StakeInfoRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	totalTickets := stakeInfo.OwnMempoolTix + stakeInfo.Live + stakeInfo.Immature + stakeInfo.Unspent
```
itswisdomagain: Does this really represent the total number of tickets in the querying wallet? I think the sum of the following better represent the total:
`stakeInfo.OwnMempoolTix + stakeInfo.Live + stakeInfo.Immature + stakeInfo.Expired + stakeInfo.Revoked + stakeInfo.Voted + stakeInfo.Missed`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,64 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+		Revoked:       stakeInfo.Revoked,
+		Total:         totalTickets,
+		Unspent:       stakeInfo.Unspent,
+	}, nil
+}
+
+func (c *WalletRPCClient) PurchaseTicket(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) ([]string, error) {
+	priceResponse, err := c.walletService.TicketPrice(ctx, &walletrpc.TicketPriceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())
```
itswisdomagain: *ticket ticketPrice*

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
+package commands
+
+import (
+	"context"
+	"strings"
+
+	"github.com/raedahgroup/dcrlibwallet"
+
```
itswisdomagain: space can go

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+	PoolAddress      string             `long:"pool-address" description:"The address of the stake pool used. Pool mode will be disabled if an empty string is passed."`
+	PoolFees         float64            `long:"pool-fees" description:"The stake pool fees amount." long-description:"This must be set to a positive value in the allowed range of 0.01 to 100.00 to be valid. It must be set when the pool-address is also set."`
+	Expiry           uint32             `long:"expiry" default:"0" description:"The height at which the tickets expire and can no longer enter the blockchain. It defaults to 0 (no expiry)."`
+	TxFee            int64              `long:"tx-fee" description:"Fees per kB to use for the transaction generating outputs to use for buying tickets." long-description:"If 0 is passed, the global value for a transaction fee will be used."`
+	TicketFee        int64              `description:"Fees per kB to use for all purchased tickets." long-description:"If 0 is passed, the global value for a ticket fee will be used."`
+	Args             PurchaseTicketArgs `positional-args:"yes"`
+}
+
+type PurchaseTicketArgs struct {
+	Account string `positional-arg-name:"from-account"`
```
itswisdomagain: Let's make this an option instead, no default value. I'm thinking `godcr purchaseticket from=default`. If `args.from == ""`, use 0 as account number else use `wallet.AccountNumber` function to look up account number

`godcr receive [account]` is a little intuitive - user can quite easily understand that the `[account]` argument represents the account that funds will be received into.

`godcr purchaseticket [account]` would make sense if `[account]` means the account that the purchased ticket will be added to but tickets are not added to individual accounts but to the wallet as a whole (I think).

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+		Account:               account,
+	})
+	if err != nil {
+		return err
+	}
+	output := strings.Builder{}
+	for _, ticketHash := range tickets {
+		output.WriteString(ticketHash + "\n")
+	}
+	termio.PrintStringResult(strings.TrimSpace(output.String()))
```
itswisdomagain: This currently simply prints out the hashes on terminal, some contextual message would help.
`n tickets purchased with hash <hash1>, <hash2>` or something of that nature, just precede the hash printing with a message that provides context for the command output the user sees

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,64 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+
+func (c *WalletRPCClient) StakeInfo(ctx context.Context) (*walletcore.StakeInfo, error) {
+	stakeInfo, err := c.walletService.StakeInfo(ctx, &walletrpc.StakeInfoRequest{})
+	if err != nil {
+		return nil, err
+	}
+
+	totalTickets := stakeInfo.OwnMempoolTix + stakeInfo.Live + stakeInfo.Immature + stakeInfo.Unspent
+
+	return &walletcore.StakeInfo{
```
itswisdomagain, line 408: This should include counts for `Voted` and `Missed`

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/raedahgroup/dcrlibwallet"
+
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/cli/termio"
+)
+
+type PurchaseTicketCommand struct {
+	commanderStub
+	MinConfirmations uint32             `long:"min-conf" default:"1" description:"The number of required confirmations for funds used to purchase a ticket." long-description:"If set to zero, it will use unconfirmed and confirmed outputs to purchase tickets."`
```
itswisdomagain: `long:"minconf" default:"0"`
the godcr program has used 0 as default number of required/minimum confirmations for various operations, let's stick to that for now. See #125 for more

I think we should avoid hyphens for option names, makes it harder to type. I see a number of such here

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+	TicketFee        int64              `description:"Fees per kB to use for all purchased tickets." long-description:"If 0 is passed, the global value for a ticket fee will be used."`
+	Args             PurchaseTicketArgs `positional-args:"yes"`
+}
+
+type PurchaseTicketArgs struct {
+	Account string `positional-arg-name:"from-account"`
+}
+
+func (ptc PurchaseTicketCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
+	passphrase, err := getWalletPassphrase()
```
itswisdomagain, line 27: Currently, `godcr purchaseticket` just shows "Wallet passphrase" prompt on terminal. This can be "Enter private passphrase to purchase ticket"

cli/commands/stakeinfo.go
```diff
@@ -0,0 +1,46 @@
...
+	if stakeInfo.Revoked > 0 {
+		output += fmt.Sprintf("revoked %d  ", stakeInfo.Revoked)
+	}
+	if stakeInfo.OwnMempoolTix > 0 {
+		output += fmt.Sprintf("unmined %d", stakeInfo.OwnMempoolTix)
+	}
+	if stakeInfo.Unspent > 0 {
+		output += fmt.Sprintf("unspent %d  ", stakeInfo.Unspent)
+	}
+	termio.PrintStringResult(output)
```
itswisdomagain, line 31: Based on recent comment from @raedah, show all counts whether 0 or not. We'll prune later

go.mod
```diff
@@ -2,21 +2,34 @@ module github.com/raedahgroup/godcr
 
 require (
 	github.com/aarzilli/nucular v0.0.0-20181227101716-d1a942545d6d
-	github.com/decred/dcrd/chaincfg v1.2.0
+	github.com/davecgh/go-spew v1.1.1 // indirect
```
itswisdomagain: I'm not sure what introduced some of these dependencies. Can you confirm? `go mod tidy` too if it helps.

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -360,3 +362,64 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+		Account:               request.Account,
+		Expiry:                request.Expiry,
+		NumTickets:            request.NumTickets,
+		Passphrase:            request.Passphrase,
+		PoolAddress:           request.PoolAddress,
+		PoolFees:              request.PoolFees,
+		RequiredConfirmations: request.RequiredConfirmations,
+		TicketAddress:         request.TicketAddress,
+		TicketFee:             request.TicketFee,
+		TxFee:                 request.TxFee,
```
itswisdomagain, line 450: `SpendLimit: priceResponse.TicketPrice,` is missing here. A recent comment mentioned using the ticket price as the spend limit.

Without providing a valid spend limit, the following error was encountered while trying to purchase ticket with dcrwallet rpc:
`error purchasing tickets: rpc error: code = FailedPrecondition desc = Unable to purchase tickets: wallet.PurchaseTickets: invalid operation: ticket price 16.07545648 DCR above spend limit 0 DCR`

itswisdomagain (changes requested):
Additional note:
`godcr purchaseticket` (via dcrlibwallet) gave insufficient balance error:
`Unable to purchase tickets: wallet.PurchaseTickets: insufficient balance:: txauthor.NewUnsignedTransaction`

cli/commands/purchase-ticket.go
itswisdomagain: I think @raedah has mentioned using `Spending Passphrase` before. That also makes sense for a "purchase..." command

app/walletmediums/dcrlibwallet/walletfunctions.go
barisere: Where is `RequiredConfirmations` passed or used in getting the account balance? The function accepts on the account number.

app/walletmediums/dcrwalletrpc/walletfunctions.go
barisere: I'll defer to @raedah on this. A different requirement was given in #65. I'm confused too.

cli/commands/purchase-ticket.go
barisere: `godcr purchasetickets --pay-from=default`
How about that?

app/walletmediums/dcrwalletrpc/walletfunctions.go
barisere: Let @raedah clarify what the total should be.

cli/commands/purchase-ticket.go
barisere: It makes it readable.
```
$ docker run --memory-reservation 
```
Try typing --memoryreservation, or --disablecontenttrust=false. See how readable it gets when you add hyphens. It depends on the words in the flag, and how readable they are when they run together. If the individual words need to stand out in order to be more readable, hyphens (or underscores) work. If they can be read easily when joined, it's fine too. Take `--tlscert`.

app/walletmediums/dcrwalletrpc/walletfunctions.go
raedah: We dont need to calculate a total. Simply display the values that the api offers.

app/walletmediums/dcrlibwallet/walletfunctions.go
itswisdomagain: Check #143

app/walletcore/types.go
```diff
@@ -49,3 +49,20 @@ type TransactionDetails struct {
 	Outputs       []*txhelper.DecodedOutput `json:"outputs"`
 	*Transaction
 }
+
+// StakeInfo holds ticket information summary related to the wallet.
+type StakeInfo struct {
```
itswisdomagain, line 54: `AllMempoolTix`, `PoolSize` and `TotalSubsidy` were not assigned or used. It'd be nice to also show these values on cli.
`Missed` and `Voted` are needed for display on cli but they were not assigned or used either.
`Total` was assigned but never read. It should probably go, we can bring it back when it's usefulness is affirmed after testing by @raedah

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -268,3 +270,42 @@ func (lib *DcrWalletLib) GetTransaction(transactionHash string) (*walletcore.Tra
 		Outputs:       decodedTx.Outputs,
 	}, nil
 }
+
+func (lib *DcrWalletLib) StakeInfo(ctx context.Context) (*walletcore.StakeInfo, error) {
+	data, err := lib.walletLib.StakeInfo()
+	if err != nil {
+		return nil, fmt.Errorf("error getting stake info: %v", err.Error())
```
itswisdomagain: `%s` instead of `%v`

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -268,3 +270,42 @@ func (lib *DcrWalletLib) GetTransaction(transactionHash string) (*walletcore.Tra
...
+	}
+
+	total := data.OwnMempoolTix + data.Live + data.Immature + data.Unspent
```
itswisdomagain: Since it's still unclear what `total` should be, let's do without it for now, stick to displaying the returned data only for now until further direction is received post-testing

app/walletcore/types.go
```diff
@@ -49,3 +49,20 @@ type TransactionDetails struct {
...
+	Expired       uint32 `json:"expired"`
+	Immature      uint32 `json:"immature"`
+	Live          uint32 `json:"live"`
+	Missed        uint32 `json:"missed"`
+	OwnMempoolTix uint32 `json:"ownMempoolTix"`
+	Revoked       uint32 `json:"revoked"`
+	Total         uint32 `json:"total"`
+	Unspent       uint32 `json:"unspent"`
+
+	PoolSize     uint32 `json:"poolSize"`
```
itswisdomagain: `AllMempoolTix` should come here as well, so the properties above the empty line space are those that relate to the querying wallet, while below the line, properties defined are general blockchain info. Comments should be added as well

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -268,3 +270,42 @@ func (lib *DcrWalletLib) GetTransaction(transactionHash string) (*walletcore.Tra
...
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account balance: %s", err.Error())
+	}
+
+	ticketPrice, err := lib.walletLib.TicketPrice(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	if balance.Spendable < dcrutil.Amount(ticketPrice.TicketPrice) {
```
itswisdomagain: This should also account for when user specifies more than 1 ticket to purchase in `request.NumTickets`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -378,3 +378,65 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+
+func (c *WalletRPCClient) PurchaseTickets(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) ([]string, error) {
+	ticketPrice, err := c.walletService.TicketPrice(ctx, &walletrpc.TicketPriceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())
+	}
+
+	balance, err := c.AccountBalance(request.Account, int32(request.RequiredConfirmations))
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account: %v", err.Error())
```
itswisdomagain: `%s`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -378,3 +378,65 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+	}
+
+	if balance.Spendable < dcrutil.Amount(ticketPrice.TicketPrice) {
+		return nil, fmt.Errorf("insufficient funds: account balance %v is less than ticket price %v",
```
itswisdomagain: `%s` in both instances of `%v`, also see other review about balance and number of tickets above, also applies here

cli/commands/helpers.go
```diff
@@ -301,7 +301,11 @@ func getChangeAmount(prompt string) (amount float64, err error) {
 
 // getWalletPassphrase fetches the user's wallet passphrase from the user.
 func getWalletPassphrase() (string, error) {
-	result, err := terminalprompt.RequestInputSecure("Spending Passphrase", terminalprompt.EmptyValidator)
+	return getWalletPassphraseWithPrompt("Wallet passphrase")
```
itswisdomagain: "Spending Passphrase"

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+	"strings"
+
+	"github.com/raedahgroup/dcrlibwallet"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/cli/termio"
+)
+
+type PurchaseTicketCommand struct {
+	commanderStub
+	MinConfirmations uint32  `long:"min-conf" default:"0" description:"The number of required confirmations for funds used to purchase a ticket." long-description:"If set to zero, it will use unconfirmed and confirmed outputs to purchase tickets."`
```
itswisdomagain: default should be 2, that represents fully confirmed tx

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,63 @@
...
+		Account:               account,
+	})
+	if err != nil {
+		return err
+	}
+	if len(tickets) == 0 {
+		return fmt.Errorf("no ticket was purchased")
+	}
+	output := strings.Builder{}
+	output.WriteString(fmt.Sprintf("You have purchased %d ticket(s)\n", len(tickets)))
```
itswisdomagain: You can create the needed string without the builder and without the loop below using:
`output := fmt.Sprintf("You have purchased %d ticket(s)\n%s", strings.Join(tickets, "\n"))`

go.mod
```diff
@@ -10,10 +10,10 @@ require (
 	github.com/decred/dcrd/wire v1.2.0
 	github.com/decred/dcrwallet v1.2.3-0.20181120205657-8690f1096aa7
 	github.com/decred/dcrwallet/rpc/walletrpc v1.0.1-0.20181109211527-ca582da21c08
-	github.com/decred/dcrwallet/walletseed v1.0.0
+	github.com/decred/dcrwallet/walletseed v1.0.1
 	github.com/go-chi/chi v3.3.3+incompatible
 	github.com/jessevdk/go-flags v1.4.0
-	github.com/raedahgroup/dcrlibwallet v0.0.0-20190106235103-e608e7f143affaf1722b1634f1aba61eed8feda0
+	github.com/raedahgroup/dcrlibwallet v1.0.0-rc1.0.20190106225103-e608e7f143af
```
itswisdomagain: Update this to `github.com/raedahgroup/dcrlibwallet v1.0.0-rc1.0.20190108195612-81f0df0be7a3` to get the latest changes to the dcrlibwallet repo that improves the `StakeInfo` function

itswisdomagain (changes requested):

app/walletcore/types.go
```diff
@@ -62,7 +62,9 @@ type StakeInfo struct {
 	Total         uint32 `json:"total"`
 	Unspent       uint32 `json:"unspent"`
 
-	PoolSize     uint32 `json:"poolSize"`
-	Voted        uint32 `json:"voted"`
-	TotalSubsidy int64  `json:"totalSubsidy"`
+	// General blockchain stake info
+	AllMempoolTix uint32 `json:"allMempoolTix"`
+	PoolSize      uint32 `json:"poolSize"`
+	Voted         uint32 `json:"voted"`
```
itswisdomagain: `Voted` is related to the wallet, the number of tickets owned by the wallet that have participated in a vote

app/walletcore/types.go
```diff
@@ -62,7 +62,9 @@ type StakeInfo struct {
 	Total         uint32 `json:"total"`
```
itswisdomagain: `Total` can be removed, no longer assigned or used

cli/commands/commands.go
```diff
@@ -16,7 +16,7 @@ type AvailableCommands struct {
 	ShowTransaction ShowTransactionCommand `command:"showtransaction" description:"Show details of a transaction"`
 	Help            HelpCommand            `command:"help" description:"Show general application help. Run help <command-name> to get help message for a specific command"`
 	StakeInfo       StakeInfoCommand       `command:"stakeinfo" description:"Show information about the wallet stakes, tickets and their statuses"`
-	PurchaseTicket  PurchaseTicketCommand  `command:"purchaseticket" description:"Purchase a ticket"`
+	PurchaseTicket  PurchaseTicketsCommand `command:"purchasetickets" description:"Purchase tickets"`
```
itswisdomagain: `PurchaseTickets` to maintain uniformity
description:Purchase 1 or more tickets

cli/commands/purchase-ticket.go
```diff
@@ -53,11 +53,8 @@ func (ptc PurchaseTicketCommand) Run(ctx context.Context, wallet walletcore.Wall
...
 		return fmt.Errorf("no ticket was purchased")
 	}
-	output := strings.Builder{}
-	output.WriteString(fmt.Sprintf("You have purchased %d ticket(s)\n", len(tickets)))
-	for _, ticketHash := range tickets {
-		output.WriteString(ticketHash + "\n")
-	}
-	termio.PrintStringResult(strings.TrimSpace(output.String()))
+	output := fmt.Sprintf("You have purchased %d ticket(s)\n%s", len(tickets), strings.Join(tickets, "\n"))
+	termio.PrintStringResult(strings.TrimSpace(output))
```
itswisdomagain: `TrimSpace` should no longer be necessary, `strings.Join` does not leave trailing spaces

itswisdomagain (changes requested):
Some little changes here and there and this should be good to merge

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -378,3 +378,67 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+
+func (c *WalletRPCClient) PurchaseTickets(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) ([]string, error) {
+	ticketPrice, err := c.walletService.TicketPrice(ctx, &walletrpc.TicketPriceRequest{})
+	if err != nil {
+		return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())
+	}
+
+	balance, err := c.AccountBalance(request.Account, int32(request.RequiredConfirmations))
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account: %s", err.Error())
```
itswisdomagain: account balance

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -268,3 +270,45 @@ func (lib *DcrWalletLib) GetTransaction(transactionHash string) (*walletcore.Tra
...
+
+func (lib *DcrWalletLib) PurchaseTickets(ctx context.Context, request dcrlibwallet.PurchaseTicketsRequest) ([]string, error) {
+	balance, err := lib.AccountBalance(request.Account, int32(request.RequiredConfirmations))
+	if err != nil {
+		return nil, fmt.Errorf("could not fetch account balance: %s", err.Error())
+	}
+
+	ticketPrice, err := lib.walletLib.TicketPrice(ctx)
+	if err != nil {
+		return nil, err
```
itswisdomagain: dcrwallet uses `return nil, fmt.Errorf("could not determine ticket ticketPrice: %s", err.Error())`, preferred

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -378,3 +378,67 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
...
+		TxFee:                 request.TxFee,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("error purchasing tickets: %s", err.Error())
+	}
+	ticketHashes := make([]string, len(response.GetTicketHashes()))
+	for i, ticketHash := range response.GetTicketHashes() {
+		hash, err := chainhash.NewHash(ticketHash)
+		if err != nil {
+			return ticketHashes, fmt.Errorf("error purchasing tickets: %s", err.Error())
```
itswisdomagain: error purchasing tickets? I believe at this point, the tickets were indeed purchased but the returned hash could not be parsed/processed, user should know they may have been debited so they don't repeat the command and get debited twice

cli/commands/purchase-ticket.go
```diff
@@ -0,0 +1,60 @@
...
+	NumTickets       uint32  `long:"num-tickets" default:"1" description:"The number of tickets to purchase."`
+	PoolAddress      string  `long:"pool-address" description:"The address of the stake pool used. Pool mode will be disabled if an empty string is passed."`
+	PoolFees         float64 `long:"pool-fees" description:"The stake pool fees amount." long-description:"This must be set to a positive value in the allowed range of 0.01 to 100.00 to be valid. It must be set when the pool-address is also set."`
+	Expiry           uint32  `long:"expiry" default:"0" description:"The height at which the tickets expire and can no longer enter the blockchain. It defaults to 0 (no expiry)."`
+	TxFee            int64   `long:"tx-fee" description:"Fees per kB to use for the transaction generating outputs to use for buying tickets." long-description:"If 0 is passed, the global value for a transaction fee will be used."`
+	TicketFee        int64   `description:"Fees per kB to use for all purchased tickets." long-description:"If 0 is passed, the global value for a ticket fee will be used."`
```
itswisdomagain: No short/long name?

itswisdomagain (changes requested):
These also

itswisdomagain:
General stake info returned by dcrlibwallet is different from dcrwallet rpc. This might have something to do with the implementation of the `StakeInfo` functions in both projects, should probably file as a bug and attend to it later.

dcrlibwallet:
```
stake info for wallet:
expired 0  immature 0  live 0  revoked 0  unmined 0  unspent 0  allmempooltix 0  poolsize 0  missed 0  voted 0  total subsidy 0
```

dcrwallet rpc:
```
stake info for wallet:
expired 0  immature 0  live 0  revoked 0  unmined 0  unspent 1  allmempooltix 10  poolsize 5098  missed 1  voted 0  total subsidy 0
```

Created issue #146 to track this

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -430,13 +430,15 @@ func (c *WalletRPCClient) PurchaseTickets(ctx context.Context, request dcrlibwal
 		TxFee:                 request.TxFee,
 	})
 	if err != nil {
-		return nil, fmt.Errorf("error purchasing tickets: %s", err.Error())
+		return nil, fmt.Errorf("could not complete ticket(s) purchase, encountered an error. " +
+			"Please check your balance before retrying.\nError: %s", err.Error())
```
itswisdomagain: - A different error message is displayed for this same scenario when using dcrlibwallet, let's harmonize the messages.
- The first line of this error message begins with a small letter, the next line starts with a capital letter

itswisdomagain: I have to add, I like the simplicity of the error message displayed when using dcrlibwallet. When the error message displayed is long, it becomes harder to make sense of the message displayed. The message displayed here has at least 2 sentences before the main error. The main error itself (knowing how verbose dcrwallet rpc can be) is gonna be another long story.

app/walletmediums/dcrlibwallet/walletfunctions.go
```diff
@@ -302,13 +302,17 @@ func (lib *DcrWalletLib) PurchaseTickets(ctx context.Context, request dcrlibwall
...
 
 	if balance.Spendable < dcrutil.Amount(ticketPrice.TicketPrice*int64(request.NumTickets)) {
 		return nil, fmt.Errorf("insufficient funds: spendable account balance (%s) is less than ticket price %s",
 			balance.Spendable, dcrutil.Amount(ticketPrice.TicketPrice))
 	}
 
-	return lib.walletLib.PurchaseTickets(ctx, &request)
+	tickets, err := lib.walletLib.PurchaseTickets(ctx, &request)
+	if err != nil {
+		return nil, fmt.Errorf("could not complete ticket(s) purchase, encountered an error: %s", err.Error())
```
itswisdomagain: might be better (more readable) to move the actual error message to the next line

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -430,13 +430,15 @@ func (c *WalletRPCClient) PurchaseTickets(ctx context.Context, request dcrlibwal
...
 	}
 	ticketHashes := make([]string, len(response.GetTicketHashes()))
 	for i, ticketHash := range response.GetTicketHashes() {
 		hash, err := chainhash.NewHash(ticketHash)
 		if err != nil {
-			return ticketHashes, fmt.Errorf("error purchasing tickets: %s", err.Error())
+			return ticketHashes, fmt.Errorf("encountered an error while processing purchased ticket(s), " +
+				"please check your balance before retrying\nError: %s", err.Error())
```
itswisdomagain: - first line of error begins with small letter, next line begins with capital letter
- can this error message be simplified?

itswisdomagain (changes requested):

itswisdomagain (approved):

Issue #114, opened 2019-01-02
send-custom - improve utxo display and selection


itswisdomagain:
**Unspent output selection**
Do not show the list by default. Instead, ask the user if to list the outputs for manual selection or to allow the program auto pick the first best sized input(s) to use. The following prompt should do the trick:
```
Would you like to (a)utomatically or (m)anually select inputs? (A/m)
```

**Unspent output display**
When listing unspent outputs, show date, confirms and Received with Address instead of showing the utxo id.
This could look like:
```
[1] 4.1998656 DCR from TsfEKMhofscqRTbr1dunLQgzRb7K8hSiy8U    02/01/2019 5:01 PM    10 confirmation(s)
```
Currently, it looks like:
```
[1]: f54b60c7ac171f42d10871c37808b82e7c22f7ab30ac6647a080be8bbb748123:0 (4.1998656 DCR)
```

Issue #113, opened 2019-01-02
send feature - multiple destination addresses


itswisdomagain:
Both cli and web interface should allow users provide multiple destination addresses, specifying the amount to send to each address. On cli, the flow can be as follows:

```
...
Destination Address:
Amount (DCR):
Destination Address 2 (or blank to continue):
Amount (DCR):
[repeat until done]
...
```

Issue #115, opened 2019-01-02
send-custom - customize change outputs


itswisdomagain:
Allow user to define number of change outputs to use for a transaction. If user choses to use more than 1 change output, ask if user wants to use automatically generated random amounts for each change output, or to manually specify amount for each change output. If user choses to use 1 change output, all the change amount from transaction should be assigned to the single change output.

Addresses should be generated for each change output and displayed to the user (with the assigned change amount) for final confirmation before broadcasting transaction. Total amount assigned to all change addresses should equal total change amount from transaction.

An idea of how this should look on cli:
```
....
How many change outputs would you like to use? (default: 1): 2
Use random amounts for the change outputs? (Y/n): y
Enter your spending passphrase: 
You are about to spend the input
9.423 DCR from TsgUAq54uZ6GW1VQwgjtgG7KB68pJwy73dJ
and send
2.123 DCR to TsgQwgjtgG7KB68pJwyUAq54uZ6GW1V73dJ
1.123 DCR to TsgJwyUAq54uZ6GWQwgjtgG7KB68p1V83dJ
3.123 DCR to TsgJwyWQwgjtgG7KB68pUAq54uZ6G1V83dJ (change)
3.123 DCR to TsgJwyuZ6G1V8WQwgjtgG7KB68pUAq543dJ (change)
Do you want to broadcast it? (y/n): y
Sent. Txid b5f844895a4e5b4e496929be0356e8435a8c30976d6bd59a29539fddd37fc9d7
```

PR #119, opened 2019-01-03
send feature - multiple destination addresses and improved utxo display and selection

ademuanthony:
Resolved [issue 113](https://github.com/raedahgroup/godcr/issues/113) and [issue 114](https://github.com/raedahgroup/godcr/issues/114)
Added the feature to allow sending to multiple destinations

Added a -l flag for the user to control whether or not the output option list should be selected automatically or manually.
If the flag is not provided, the best-sized input is selected automatically, shown to the user and used for the transaction if the user agree to continue

if the -l flag is provided, the best-sized input is provided as default selection for the user who is provided with the option to either use it or manually select the inputs for the transaction.
The list is sorted by the amount given a tabular display showing receiving address, amount, date and no of confirmations


itswisdomagain:
#114 has been updated to no longer include a list flag for the `send-custom` command. 

Instead, after user enters destination addresses and amounts, ask user whether to manually chose inputs for the transaction or not:
```
Would you like to (a)utomatically or (m)anually select inputs? (A/m)
```

If user choses to automatically select inputs, run your best size algorithm to decide what utxos to use as inputs for the transaction. No need to display the automatically selected inputs at this time.

If user choses manual selection, list the utxos and ask user to select the unspent outputs to use as inputs for the new transaction:
```
Select unspent outputs to use as inputs for this transaction (e.g. 1-4,6)
```

Regardless of the user's choice, we should have a list of utxos to use as inputs for this transaction. If the user choses manual selection, it may be appropriate to calculate the total sum selected by the user and confirm that it is not less than the total send amount. If it is less, inform the user and ask the user to make a new selection.

Afterwards, request for the spending password and display a following prompt using the format specified in #115:
```
You are about to spend the input(s)
9.423 DCR from TsgUAq54uZ6GW1VQwgjtgG7KB68pJwy73dJ
and send
2.123 DCR to TsgQwgjtgG7KB68pJwyUAq54uZ6GW1V73dJ
1.123 DCR to TsgJwyUAq54uZ6GWQwgjtgG7KB68p1V83dJ
Do you want to broadcast it? (y/n): y
Sent. Txid b5f844895a4e5b4e496929be0356e8435a8c30976d6bd59a29539fddd37fc9d7
```

cli/commands/helpers.go
```diff
@@ -3,12 +3,18 @@ package commands
...
+	"sort"
 	"strconv"
 	"strings"
+	"time"
 	"unicode"
 
+	"github.com/decred/dcrd/chaincfg"
 	"github.com/decred/dcrd/dcrutil"
+	"github.com/decred/dcrd/txscript"
+
```
itswisdomagain: space can go

app/walletcore/types.go
```diff
@@ -53,6 +53,7 @@ type UnspentOutput struct {
 	Tree            int32          `json:"tree"`
 	ReceiveTime     int64          `json:"receive_time"`
 	Amount          dcrutil.Amount `json:"amount"`
+	PkScript        []byte
```
itswisdomagain: `Address string` instead. Instead of returning the `PkScript` with the `UnspentOutput` object, use it to get the address and return the address with the `UnspentOutput` object

cli/commands/helpers.go
```diff
@@ -58,8 +64,11 @@ func selectAccount(wallet walletcore.Wallet) (uint32, error) {
 }
 
 // getSendDestinationAddress fetches the destination address to send DCRs to from the user.
-func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
+func getSendDestinationAddress(wallet walletcore.Wallet, index int) (string, error) {
```
itswisdomagain: Modify this function to return `[]txhelper.TransactionDestination` instead. Change the function name to `getSendTxDestinations`. The whole process of continuously asking for addresses and amounts until a blank space is encountered should be done here, to keep the send command's `Run` method uncluttered.

cli/commands/send.go
```diff
@@ -2,7 +2,10 @@ package commands
 
 import (
 	"fmt"
+	"sort"
+
 	"github.com/raedahgroup/dcrlibwallet/txhelper"
+
```
itswisdomagain, line 7: space can go

itswisdomagain (changes requested):
Some more things to consider. Will review once more when the desired behaviour is fully actualised.

ademuanthony:
bs-1
bs-2
bs-3
bs-4

cli/commands/helpers.go
```diff
@@ -75,13 +82,49 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
...
+		if err != nil {
+			return nil, fmt.Errorf("error receiving input: %s", err.Error())
+		}
+
+		if destinationAddress == "" {
+			break
+		}
+
+		if _, addressExists := sendAmounts[destinationAddress]; addressExists {
+			promptMessage := fmt.Sprintf("The address %s has already been added. Do you want to change the amount", destinationAddress)
```
itswisdomagain: Should end with a question mark

cli/commands/helpers.go
```diff
@@ -75,13 +82,49 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
...
+		destinationAddress, err := terminalprompt.RequestInput(label, validateAddressInput)
+		if err != nil {
+			return nil, fmt.Errorf("error receiving input: %s", err.Error())
+		}
+
+		if destinationAddress == "" {
+			break
+		}
+
+		if _, addressExists := sendAmounts[destinationAddress]; addressExists {
```
itswisdomagain: I absolutely love your use of the variable name `addressExists` here, instead of the regular `ok`

cli/commands/helpers.go
```diff
@@ -120,8 +163,8 @@ func getWalletPassphrase() (string, error) {
 }
 
 // getUtxosForNewTransaction fetches unspent transaction outputs to be used in a transaction.
-func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount float64) ([]string, error) {
-	var selectedUtxos []string
+func getUtxosForNewTransaction(wallet walletcore.Wallet, utxos []*walletcore.UnspentOutput, sendAmount float64) ([]*walletcore.UnspentOutput, error) {
+	var selectedUtxos []*walletcore.UnspentOutput
 	var err error
```
itswisdomagain: You can use `err` and `selectedUtxos` as named returned parameters instead of declaring them here

cli/commands/helpers.go
```diff
@@ -191,8 +234,16 @@ func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount flo
...
+		return utxos[i].Amount < utxos[j].Amount
+	})
 	for index, utxo := range utxos {
-		options[index] = fmt.Sprintf("%s (%s)", utxo.OutputKey, utxo.Amount.String())
+		date := time.Unix(utxo.ReceiveTime, 0).Format("Mon Jan 2, 2006 3:04PM")
+		txn, err := wallet.GetTransaction(utxo.TransactionHash)
+		if err != nil {
+			return nil, fmt.Errorf("error reading transaction: %s", err.Error())
+		}
+		options[index] = fmt.Sprintf("%s (%s) \t %s \t %v confirmation(s)", utxo.Address, utxo.Amount.String(), date, txn.Confirmations)
```
itswisdomagain: Add `Confirmations` to the `UnspentOutput` type and assign the appropriate value there instead of here. The value will also be needed by other interfaces such as the web and desktop when listing unspent outputs, we want to avoid writing much core code in the different interfaces and abstract as much functionality to the core `wallet` interface and the different mediums implementing the interface

cli/commands/send.go
```diff
@@ -43,27 +45,46 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	var sendAmountTotal float64
+	for _, destination := range sendDestinations {
+		sendAmountTotal += destination.Amount
+	}
 
-	sendAmount, err := getSendAmount()
-	if err != nil {
-		return err
+	if accountBalance.Spendable.ToCoin() < sendAmountTotal {
+		return fmt.Errorf("Selected account has low balance. Cannot proceed")
```
itswisdomagain: has **insufficient** balance

cli/commands/send.go
```diff
@@ -43,27 +45,46 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
 		// get all utxos in account, pass 0 amount to get all
 		utxos, err := wallet.UnspentOutputs(sourceAccount, 0)
 		if err != nil {
 			return err
 		}
 
-		utxoSelection, err = getUtxosForNewTransaction(utxos, sendAmount)
+		choice, err := terminalprompt.RequestInput("Would you like to (a)utomatically or (m)anually select inputs? (A/m)", func(input string) error {
+			switch strings.ToLower(input) {
+			case "":
```
itswisdomagain: Can handle all cases here
`case "", "a", "m":`

cli/commands/send.go
```diff
@@ -72,7 +93,26 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
 		return err
 	}
 
-	fmt.Printf("You are about to send %f DCR to %s\n", sendAmount, destinationAddress)
+	if custom {
+		fmt.Println("You are about to spend the input")
```
itswisdomagain: input(s), in case they are more than 1

cli/commands/send.go
```diff
@@ -72,7 +93,26 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+		for _, output := range utxoSelection {
+			fmt.Println(fmt.Sprintf(" %s from %s", output.Amount.String(), output.Address))
+		}
+		fmt.Println("and send it to")
```
itswisdomagain: `and send` should be sufficient here, each destination below already says "amount to address"

cli/commands/send.go
```diff
@@ -72,7 +93,26 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+		for _, destinatoin := range sendDestinations {
+			fmt.Println(fmt.Sprintf(" %v DCR to %s", destinatoin.Amount, destinatoin.Address))
```
itswisdomagain: You can use `%f` instead of `%v` here for float

cli/commands/send.go
```diff
@@ -72,7 +93,26 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
 	}
 
-	fmt.Printf("You are about to send %f DCR to %s\n", sendAmount, destinationAddress)
+	if custom {
+		fmt.Println("You are about to spend the input")
+		for _, output := range utxoSelection {
+			fmt.Println(fmt.Sprintf(" %s from %s", output.Amount.String(), output.Address))
+		}
+		fmt.Println("and send it to")
+		for _, destinatoin := range sendDestinations {
```
itswisdomagain: `destination` 😉

cli/commands/send.go
```diff
@@ -72,7 +93,26 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+			fmt.Println(fmt.Sprintf("You are about to send %f DCR to %s", sendDestinations[0].Amount, sendDestinations[0].Address))
+		} else {
+			fmt.Println("You are about to send")
+			for _, destination := range sendDestinations {
+				fmt.Println(fmt.Sprintf(" %v DCR to %s", destination.Amount, destination.Address))
+			}
+		}
+	}
+
 	sendConfirmed, err := terminalprompt.RequestYesNoConfirmation("Are you sure?", "")
```
itswisdomagain: I think the question here should be `Do you want to broadcast it?`

go.mod
```diff
@@ -2,6 +2,8 @@ module github.com/raedahgroup/godcr
 
 require (
 	github.com/aarzilli/nucular v0.0.0-20181227101716-d1a942545d6d
+	github.com/btcsuite/go-flags v0.0.0-20150116065318-6c288d648c1c
```
itswisdomagain: We have another go-flags dependency below `(github.com/jessevdk/go-flags)`, which I believe is what we actually use for this project. Please ensure there is no mix of go-flag dependencies and run `go mod tidy` to remove unused imports from go.mod and go.sum

.gitignore
```diff
@@ -6,4 +6,4 @@ vendor/
 godcr
 go.mod
 go.sum
-godcr.conf
+godcr.conf
```
itswisdomagain, line 7: `go fmt ./...`

itswisdomagain (changes requested):

itswisdomagain:
When I update amount for a previously added destination address, the numbering should not increase
screenshot 2019-01-04 at 12 29 47 am

Destination 2 was an update to destination 1, so I should have been asked for Destination address 2 again

Also the default of that prompt should be no. It's likely a mistake from the user, shouldn't assume the user meant to update the amount unless he says "y"es

app/walletcore/helper.go
```diff
@@ -0,0 +1,19 @@
...
+func GetAddressFromPkScript(activeNet *chaincfg.Params, pkScript []byte) (address string, err error) {
+	_, addresses, _, err := txscript.ExtractPkScriptAddrs(txscript.DefaultScriptVersion,
+		pkScript, activeNet)
+	if err != nil {
+		return
+	}
+	if len(addresses) < 1 {
+		return "", errors.New("Cannot extract any address from output")
+	}
+	return addresses[0].EncodeAddress(), nil
```
itswisdomagain: There are cases where a `pkScript` references more than 1 address, haven't seen it before but it exists. Let's not throw other such addresses away, instead let's return all of them. Here's an idea:
```
encodedAddresses := make([]string, len(addresses))
for i, address := range addresses {
     encodedAddresses[i] = address.EncodeAddress()
}

return strings.Join(encodedAddresses, ", ")
```

itswisdomagain: By the way, having this shared code here is really nice. The current direction for godcr is to have shared codes kept in [the WIP branch of dcrlibwallet](https://github.com/raedahgroup/dcrlibwallet/tree/dcrlibwallet-wip) so that such functions are accessible from `godcr`, `dcrandroid` and `dcrios` (all those wallet apps connect to `dcrlibwallet`)

Will move this function over there later, no need to have the PR stalled for that reason. I'll provide more information soon on when and how to put shared codes in `dcrlibwallet` subsequently.

cli/commands/helpers.go
```diff
@@ -75,13 +82,50 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
 		return nil
 	}
 
-	address, err := terminalprompt.RequestInput("Destination Address", validateAddressInput)
-	if err != nil {
-		// There was an error reading input; we cannot proceed.
-		return "", fmt.Errorf("error receiving input: %s", err.Error())
+	var destinationAddresses []string
```
itswisdomagain: This is not important, I see its real use is in line 125. See the comment there for an alternative that uses `sendAmounts` only instead of `destinationAddresses`

cli/commands/helpers.go
```diff
@@ -75,13 +82,50 @@ func getSendDestinationAddress(wallet walletcore.Wallet) (string, error) {
...
+		sendAmount, err := getSendAmount()
+		if err != nil {
+			return nil, fmt.Errorf("error receiving input: %s", err.Error())
+		}
+		sendAmounts[destinationAddress] = sendAmount
+		index++
 	}
 
-	return address, nil
+	for _, address := range destinationAddresses {
```
itswisdomagain:
```
for address, amount := range sendAmounts {
     destinations = append(destinations, txhelper.TransactionDestination{Address: address, Amount: amount})
}
```

cli/commands/helpers.go
```diff
@@ -120,10 +164,7 @@ func getWalletPassphrase() (string, error) {
 }
 
 // getUtxosForNewTransaction fetches unspent transaction outputs to be used in a transaction.
-func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount float64) ([]string, error) {
-	var selectedUtxos []string
-	var err error
-
+func getUtxosForNewTransaction(wallet walletcore.Wallet, utxos []*walletcore.UnspentOutput, sendAmount float64) (selectedUtxos []*walletcore.UnspentOutput, err error) {
```
itswisdomagain: `wallet walletcore.Wallet` is not used

cli/commands/helpers.go
```diff
@@ -191,8 +232,12 @@ func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount flo
 	}
 
 	options := make([]string, len(utxos))
+	sort.Slice(utxos, func(i, j int) bool {
+		return utxos[i].Amount < utxos[j].Amount
+	})
 	for index, utxo := range utxos {
-		options[index] = fmt.Sprintf("%s (%s)", utxo.OutputKey, utxo.Amount.String())
+		date := time.Unix(utxo.ReceiveTime, 0).Format("Mon Jan 2, 2006 3:04PM")
+		options[index] = fmt.Sprintf("%s (%s) \t %s \t %v confirmation(s)", utxo.Address, utxo.Amount.String(), date, utxo.Confirmations)
```
itswisdomagain: `%d` instead of `%v` for int

cli/commands/send.go
```diff
@@ -72,8 +89,27 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+		for _, destination := range sendDestinations {
+			fmt.Println(fmt.Sprintf(" %f DCR to %s", destination.Amount, destination.Address))
+		}
+	} else {
+		if len(sendDestinations) == 1 {
+			fmt.Println(fmt.Sprintf("You are about to send %f DCR to %s", sendDestinations[0].Amount, sendDestinations[0].Address))
+		} else {
+			fmt.Println("You are about to send")
+			for _, destination := range sendDestinations {
+				fmt.Println(fmt.Sprintf(" %v DCR to %s", destination.Amount, destination.Address))
```
itswisdomagain: `%f` instead of `%v`

cli/commands/send.go
```diff
@@ -72,8 +89,27 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
 		return err
 	}
 
-	fmt.Printf("You are about to send %f DCR to %s\n", sendAmount, destinationAddress)
-	sendConfirmed, err := terminalprompt.RequestYesNoConfirmation("Are you sure?", "")
+	if custom {
+		fmt.Println("You are about to spend the input(s)")
+		for _, output := range utxoSelection {
```
itswisdomagain: `utxo` instead of `output`

cli/commands/send.go
```diff
@@ -83,14 +119,13 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
-	sendDestinations := []txhelper.TransactionDestination{{
-		Amount:  sendAmount,
-		Address: destinationAddress,
-	}}
-
 	var sentTransactionHash string
 	if custom {
-		sentTransactionHash, err = wallet.SendFromUTXOs(sourceAccount, utxoSelection, sendDestinations, passphrase)
+		var outputs []string
+		for _, utox := range utxoSelection {
```
itswisdomagain: `utxo`

go.sum
```diff
@@ -2,102 +2,75 @@ github.com/AndreasBriese/bbloom v0.0.0-20180913140656-343706a395b7 h1:PqzgE6kAMi
 github.com/AndreasBriese/bbloom v0.0.0-20180913140656-343706a395b7/go.mod h1:bOvUY6CB00SOBii9/FifXqc0awNKxLFCL/+pkDPuyl8=
 github.com/BurntSushi/xgb v0.0.0-20160522221800-27f122750802 h1:otE7gFP0KwnPTKtMvPLOcWDZxx6rXYyJGlrXbhoD7og=
 github.com/BurntSushi/xgb v0.0.0-20160522221800-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
-github.com/aarzilli/nucular v0.0.0-20181227101716-d1a942545d6d h1:BLhrKkQNOBeRDRJdevy7tE7ca2+sp3daQ6xG7vagEzg=
+github.com/aarzilli/nucular v0.0.0-20181227101716-d1a942545d6d h1:F486Q0BO+uFwzGSCfNkuQYr86h0Oap7zYdimuCyr0/0=
```
itswisdomagain: I can't build your code as is, without deleting go.sum before building:
```
go: verifying github.com/aarzilli/nucular@v0.0.0-20181227101716-d1a942545d6d: checksum mismatch
        downloaded: h1:BLhrKkQNOBeRDRJdevy7tE7ca2+sp3daQ6xG7vagEzg=
        go.sum:     h1:F486Q0BO+uFwzGSCfNkuQYr86h0Oap7zYdimuCyr0/0=
```

The checksum currently on master `(BLhrKkQNOBeRDRJdevy7tE7ca2+sp3daQ6xG7vagEzg=)` was built with go 1.11.4. Are you using same version? Remember to delete go.sum and perhaps clear mod cache at least once after updating to go 1.11.4 before building. Also run `go mod tidy`

We can't merge this in until this is fixed as it will not build on machines running go 1.11.4 which is the go version now used for this project

itswisdomagain (changes requested):

itswisdomagain:
Some anomalies to note and correct

- **Inputs or outputs**
See the screenshot below. The prompt asks: `Would you like to (a)utomatically or (m)anually select **inputs**?`, then says `Select unspent **outputs**`. Also, invalid selection message reads: ` Total amount from selected **outputs** is smaller than amount to send`
screenshot 2019-01-05 at 7 03 37 pm

- **utxo confirmation message**
I have 3 available utxos and I select all 3. Confirmation message says I selected 6 utxos, no idea where the rest came from...
screenshot 2019-01-05 at 7 10 15 pm

cli/commands/helpers.go
```diff
@@ -180,25 +217,57 @@ func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount flo
 		for _, n := range selection {
 			utxo := utxos[n]
 			totalAmountSelected += dcrutil.Amount(utxo.Amount).ToCoin()
-			selectedUtxos = append(selectedUtxos, utxo.OutputKey)
+			selectedUtxos = append(selectedUtxos, utxo)
```
itswisdomagain, line 222: Clear `selectedUtxos` every time this validate function runs before appending to `selectedUtxos`, to prevent repeated utxos. See https://github.com/raedahgroup/godcr/pull/119#issuecomment-451679344 for an example of the repeated utxos issue.

itswisdomagain (changes requested):

itswisdomagain:
Everything looks good, just this https://github.com/raedahgroup/godcr/pull/119#issuecomment-451311595 I thought was resolved, still there.

screenshot 2019-01-06 at 12 47 54 am

itswisdomagain (approved):

PR #123, opened 2019-01-04
Multi change output

### LINKED ISSUE #115: send-custom - customize change outputs

ademuanthony:
Resolves #115 

This PR is about controlling the change output. 
The user can specify the number of output to use and decide whether to send a random amount to each output or specify the amount for each output
ch-001
ch-002


cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
 	return amount, nil
 }
 
+// getSendAmount fetches the amout of DCRs to send from the user.
+func getNChangeOutput() (int, error) {
+	var number int64
+	var err error
```
itswisdomagain: Use named variables for instances like this

cli/commands/helpers.go
```diff
@@ -8,10 +8,17 @@ import (
 	"unicode"
 
 	"github.com/decred/dcrd/dcrutil"
+	"github.com/raedahgroup/dcrlibwallet/txhelper"
 	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/raedahgroup/godcr/cli/termio/terminalprompt"
 )
 
+const (
+	// todo dcrwallet should define this as typed errors
```
itswisdomagain: As discussed, this can go

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
 }
 
+// getSendAmount fetches the amout of DCRs to send from the user.
+func getNChangeOutput() (int, error) {
+	var number int64
+	var err error
+
+	validateNumber := func(input string) error {
+		if input == "" {
+			input = "1"
```
itswisdomagain: you can just assign `number = 1` and `return nil` here, no need to proceed to parse "1" and see if an error will occur

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+		return nil
+	}
+
+	_, err = terminalprompt.RequestInput("How many change outputs would you like to use? (default: 1)", validateNumber)
+	if err != nil {
+		// There was an error reading input; we cannot proceed.
+		return 0, fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	return int(number), nil
```
itswisdomagain: `number` and `err` are assigned in line 130, once you make them named return parameters, you can just do `return` here

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+}
+
+// getSendAmount fetches the amount of DCRs to send from the user.
+func getChangeAmount(prompt string) (float64, error) {
+	var amount float64
+	var err error
```
itswisdomagain: named parameters

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+
+	validateAmount := func(input string) error {
+		if input == "" {
+			input = "0"
```
itswisdomagain: just assign `amount = 0` here and `return nil`

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+		return nil
+	}
+
+	_, err = terminalprompt.RequestInput(prompt, validateAmount)
+	if err != nil {
+		// There was an error reading input; we cannot proceed.
+		return 0, fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	return amount, nil
```
itswisdomagain: can just `return` here after using named return parameters

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+	_, err = terminalprompt.RequestInput("How many change outputs would you like to use? (default: 1)", validateNumber)
+	if err != nil {
+		// There was an error reading input; we cannot proceed.
+		return 0, fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	return int(number), nil
+}
+
+// getSendAmount fetches the amount of DCRs to send from the user.
```
itswisdomagain: `// getChangeAmount...`

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
+		return 0, fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	return amount, nil
+}
+
+// getChangeOutputDestinations fetches the amount to be sent to each change address
+func getChangeOutputDestinations(addresses []string, changeAmountInDcr float64) (destinations []txhelper.TransactionDestination, err error) {
+	var commutativeAmount float64
+	fmt.Println(fmt.Sprintf("Enter the amount to be sent to each change address. The sum of all the amounts should not be more than %f", changeAmountInDcr))
```
itswisdomagain: `fmt.Printf("Enter the amount to be sent to each change address. The sum of all the amounts should not be more than %f DCR\n", changeAmountInDcr)
`

cli/commands/helpers.go
```diff
@@ -110,18 +117,96 @@ func getSendAmount() (float64, error) {
...
 	}
 	return result, nil
 }
 
 // getUtxosForNewTransaction fetches unspent transaction outputs to be used in a transaction.
-func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount float64) ([]string, error) {
+func getUtxosForNewTransaction(utxos []*walletcore.UnspentOutput, sendAmount float64) ([]string, float64, error) {
 	var selectedUtxos []string
+	var totalAmountSelected float64
 	var err error
```
itswisdomagain: named variables

cli/commands/send.go
```diff
@@ -53,20 +56,119 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	}
+
+	amountInAtom, err := txhelper.AmountToAtom(totalInputAmount)
+	if err != nil {
+		return err
+	}
+
+	var changeOutputDestinations []txhelper.TransactionDestination
+
+	useRandomNo, err := terminalprompt.RequestYesNoConfirmation("Use random amounts for the change outputs?", "y")
```
itswisdomagain: `useRandomNo` sounds like use random? no
`useRandomChangeAmounts` is better

cli/commands/send.go
```diff
@@ -53,20 +56,119 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+			if changeAmount == 0 {
+				if lastChangeAmountInDcr > 0 {
+					fmt.Println("using default value")
+				}
+				//todo the display of the transaction summary should show the last amount to the user if it is grater
+				// than 0 as we can add it to the list currently since the change amount is a raw estimate
+				break
+			}
+
+			allDestination := append(sendDestinations, changeOutputDestinations...)
```
itswisdomagain: `changeOutputDestinations` is an empty slice at this point, this operation does nothing, what are you trying to achieve?

cli/commands/send.go
```diff
@@ -53,20 +56,119 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+
+			changeAddress, err := wallet.GenerateReceiveAddress(sourceAccount)
```
itswisdomagain: Line 126 above generated a change address that was discarded, don't generate another one here, use that address instead

cli/commands/send.go
```diff
@@ -53,20 +56,119 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+			changeDestination := txhelper.TransactionDestination{Address: changeAddress, Amount: changeAmount}
+			allDestination = append(allDestination, changeDestination)
+
+			_, remainingChange, err := wallet.GenerateChangeAddresses(sourceAccount, 1, len(utxoSelection),
```
itswisdomagain: The generated address here should be used in the next cycle, just as `remainingChange` is stored in `lastChangeAmountInDcr` for use in the next cycle

cli/commands/send.go
```diff
@@ -53,20 +56,119 @@ func send(wallet walletcore.Wallet, custom bool) (err error) {
...
+	}
+
+	var changeOutputDestinations []txhelper.TransactionDestination
+
+	useRandomNo, err := terminalprompt.RequestYesNoConfirmation("Use random amounts for the change outputs?", "y")
+	if err != nil {
+		return fmt.Errorf("error reading your response: %s", err.Error())
+	}
+
+	if useRandomNo {
```
itswisdomagain: This whole `send` function has grown too big. Split into smaller functions. The if and else case here can be made into separate functions; perhaps other parts of this function can be split out too.

itswisdomagain (changes requested):
The `send` function is a mess, please let's simplify it. The code for generating change addresses as user manually enters amounts progressively can be improved, even if it means creating or modifying core wallet functions.

I haven't tested this PR yet, will test after these changes are made. The code might work now but after simplifying the `send` function, a re-test would definitely be needed, so will just wait till then.

Also, this https://github.com/raedahgroup/godcr/pull/119#discussion_r245484660 affects this PR as well.

cli/commands/send.go
ademuanthony: In this loop, at first run, `changeOutputDestinations` will be empty but a `changeDestination` will be appended to it after every successful run (line 164). This list we are building, the `sendDestinations` and the `changeDestination` we want to check and see if it can be added to our `changeOutputDestinations` makes up `allDestination` that we are sending to the `wallet.GenerateChangeAddresses` to get the estimated `remainingChange`.

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
 	return amount, nil
 }
 
+// getSendAmount fetches the amout of DCRs to send from the user.
+func getNChangeOutput() (number int, err error) {
```
itswisdomagain: This function is only called in `getChangeDestinationsWithRandomAmounts`. That function is not very big, the code here can go there for simplicity.

itswisdomagain: Alternatively, preferably, we can have a `RequestNumberInput` function in the `terminalprompt ` package that performs the number validation done here and returns the valid number entered by a user

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+	if err != nil {
+		// There was an error reading input; we cannot proceed.
+		return 0, fmt.Errorf("error receiving input: %s", err.Error())
+	}
+
+	return
+}
+
+// getChangeAmount fetches the amount of DCRs to send from the user.
+func getChangeAmount(prompt string) (amount float64, err error) {
```
itswisdomagain: This `getChangeAmount` function is called by `getChangeDestinationsFromUser`. Please move it to below the `getChangeDestinationsFromUser` function for easier code scanning. Functions should appear in the order they are called.

`getChangeOutputDestinations` is called by `send.go`, which then calls `getChangeDestinationsWithRandomAmounts` or `getChangeDestinationsFromUser` in that order. When scanning, I expect to see those functions in the following order:
- `getChangeOutputDestinations`
- `getChangeDestinationsWithRandomAmounts`
- `getChangeDestinationsFromUser`
- `getChangeAmount`

Also note that `getChangeDestinationsWithRandomAmounts` and `getChangeDestinationsFromUser` would need re-placement to conform to the order described above.

Ordered placement of functions allow for easier code scanning.

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+		changeOutputDestinations, err = getChangeDestinationsWithRandomAmounts(wallet, amountInAtom, sourceAccount,
+			nUtxoSelection, sendDestinations)
+	} else {
+		changeOutputDestinations, err = getChangeDestinationsFromUser(wallet, amountInAtom, sourceAccount,
+			nUtxoSelection, sendDestinations)
+	}
+	return
+}
+
+// getChangeDestinationsFromUser fetches change destination from the user
```
itswisdomagain: `// getChangeDestinationsFromUser fetches change destination from the user progressively until the total available change amount is covered`

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+func getChangeDestinationsFromUser(wallet walletcore.Wallet, amountInAtom int64, sourceAccount uint32, nUtxoSelection int, sendDestinations []txhelper.TransactionDestination) (changeOutputDestinations []txhelper.TransactionDestination, err error) {
```
itswisdomagain: Since `err` is shadowed in the loop below, this function can survive without named parameters. You'll only need to declare `var changeOutputDestinations []txhelper.TransactionDestination` below

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+// getChangeAmount fetches the amount of DCRs to send from the user.
+func getChangeAmount(prompt string) (amount float64, err error) {
+	validateAmount := func(input string) error {
+		if input == "" {
+			amount = 0
+			return nil
+		}
+
+		amount, err = strconv.ParseFloat(input, 64)
+		if err != nil {
```
itswisdomagain, line 191: Reject negative amounts and cases where user enters "0" as amount. `amount` must be greater than 0 here. Only treat amount as 0 if user enters "" (which is already done above)

I'm thinking, respond with invalid amount if `err != nil || amount <= 0`

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+
+	return
+}
+
+// getChangeOutputDestinations fetches the amount to be sent to each change address
+func getChangeOutputDestinations(wallet walletcore.Wallet, totalInputAmount float64, sourceAccount uint32,
+	nUtxoSelection int, sendDestinations []txhelper.TransactionDestination) (changeOutputDestinations []txhelper.TransactionDestination, err error) {
+	useRandomChangeAmounts, err := terminalprompt.RequestYesNoConfirmation("Use random amounts for the change outputs?", "y")
+	if err != nil {
+		err = fmt.Errorf("error reading your response: %s", err.Error())
```
itswisdomagain: If you don't use named variables, you can just return the error directly here

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+		return
+	}
+
+	amountInAtom, err := txhelper.AmountToAtom(totalInputAmount)
+	if err != nil {
+		return
+	}
+
+	if useRandomChangeAmounts {
+		changeOutputDestinations, err = getChangeDestinationsWithRandomAmounts(wallet, amountInAtom, sourceAccount,
```
itswisdomagain: If you don't use named variables, you can just return the result of calling this function directly instead of assigning:
`return getChangeDestinationsWithRandomAmounts(wallet, amountInAtom, sourceAccount, nUtxoSelection, sendDestinations)`

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+
+		prompt := fmt.Sprintf("[%d]: Change Amount (DCR) (default: %f)", index+1, dcrutil.Amount(defaultAmount).ToCoin())
+		changeAmount, err := getChangeAmount(prompt)
+		if err != nil {
+			// we cannot just use return here as the named varaible `err` has been shadowed in line 207 above
+			return nil, err
+		}
+
+		if changeAmount > dcrutil.Amount(defaultAmount).ToCoin() {
+			fmt.Println(fmt.Sprintf("Error: %s. Try again with a smaller amount", errMsgInputNotEnoughForTxn))
```
itswisdomagain: A more correct error message would be `fmt.Sprintf("Invalid amount. Enter an amount less than %f", dcrutil.Amount(defaultAmount).ToCoin())`

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+
+	var index int
+	for {
+		address, err := wallet.GenerateReceiveAddress(sourceAccount)
+		if err != nil {
+			return nil, fmt.Errorf("error in generating address: %s", err.Error())
+		}
+		changeAddresses = append(changeAddresses, address)
+		totalChangeAmount, err := txhelper.EstimateChange(nUtxoSelection, amountInAtom, sendDestinations, changeAddresses)
+		defaultAmount := totalChangeAmount - amountAssigned
```
itswisdomagain: `defaultAmount` is always converted to float64 everywhere it's used, could just make the value a float64 value directly here instead:

`defaultAmount := dcrutil.Amount(totalChangeAmount - amountAssigned).ToCoin()`

cli/commands/helpers.go
itswisdomagain: This change would mean both `errMsgInputNotEnoughForTxn` and `errMsgInputNotEnoughForFee` constants defined above will no longer be used and can be deleted

cli/commands/helpers.go
```diff
@@ -151,17 +157,184 @@ func getSendAmount() (float64, error) {
...
+		portion := rand.Float64()
+		portionRations = append(portionRations, portion)
+		rationSum += portion
+	}
+	for _, portion := range portionRations {
+		amount := portion * float64(changeAmount) / rationSum
+		portions = append(portions, amount)
+	}
+
+	for i, address := range changeAddresses {
```
itswisdomagain: This last loop can be merged with the previous. Here's something I played with:
```
for i, portion := range portionRations {
	portionPercentage := portion / rationSum
	amount := portionPercentage * float64(changeAmount)

	changeOutput := txhelper.TransactionDestination{
		Address: changeAddresses[i],
		Amount: dcrutil.Amount(amount).ToCoin(),
	}
	changeOutputDestinations = append(changeOutputDestinations, changeOutput)
}
```

cli/commands/send.go
```diff
@@ -60,7 +59,9 @@ func send(wallet walletcore.Wallet, custom bool) error {
 		return fmt.Errorf("Selected account has insufficient balance. Cannot proceed")
 	}
 
+	var changeOutputDestinations []txhelper.TransactionDestination
 	var utxoSelection []*walletcore.UnspentOutput
+	var totalInputAmount float64
 
 	if custom {
```
itswisdomagain: Let's split this point onwards into 2 functions `completeCustomSend` and `completeNormalSend` to avoid the  multiple `if custom` blocks that make this function a harder to read. Both `completeCustomSend` and `completeNormalSend` should return `hash string, err error`

The ending part of this function should then look like this:
```
var sentTxHash string
if custom {
      sentTxHash, err = completeCustomSend(....)
} else {
      sentTxHash, err = completeNormalSend(....)
}

if err != nil {
	return err
}

fmt.Println("Sent txid", sentTxHash)
return nil
```

cli/commands/send.go
```diff
@@ -80,9 +81,15 @@ func send(wallet walletcore.Wallet, custom bool) error {
 			return fmt.Errorf("error in reading choice: %s", err.Error())
 		}
 		if strings.ToLower(choice) == "a" || choice == "" {
-			utxoSelection = bestSizedInput(utxos, sendAmountTotal)
+			utxoSelection, totalInputAmount = bestSizedInput(utxos, sendAmountTotal)
 		} else {
-			utxoSelection, err = getUtxosForNewTransaction(utxos, sendAmountTotal)
+			utxoSelection, totalInputAmount, err = getUtxosForNewTransaction(utxos, sendAmountTotal)
```
itswisdomagain: any `err` returned is not handled

cli/commands/send.go
```diff
@@ -100,6 +107,9 @@ func send(wallet walletcore.Wallet, custom bool) error {
 		for _, destination := range sendDestinations {
 			fmt.Println(fmt.Sprintf(" %f DCR to %s", destination.Amount, destination.Address))
```
itswisdomagain: tabs would help here to achieve cleaner message display here:
` %f DCR \t to %s`

cli/commands/send.go
```diff
@@ -100,6 +107,9 @@ func send(wallet walletcore.Wallet, custom bool) error {
...
 		}
+		for _, destination := range changeOutputDestinations {
+			fmt.Println(fmt.Sprintf(" %f DCR to %s(change)", destination.Amount, destination.Address))
```
itswisdomagain: there should be a single space between the address and `(change)`

cli/commands/send.go
itswisdomagain: do same when printing the selected utxos

itswisdomagain (changes requested):
This is great improvement. Some more code adjustments below.

Also, please modify `getSendTxDestinations` in `cli/commands/helpers.go` to also return `sendAmountTotal` and eliminate the loop performed in line 54 of `cli/commands/send.go`

cli/termio/terminalprompt/terminalprompt.go
```diff
@@ -43,6 +44,35 @@ func RequestInput(message string, validate ValidatorFunction) (string, error) {
 	}
 }
 
+// RequestNumberInput uses RequestInput requests numeric input from the user.
```
itswisdomagain: `// RequestNumberInput uses RequestInput **to** request numeric input from the user.`

cli/commands/helpers.go
```diff
@@ -15,6 +16,11 @@ import (
 	"github.com/raedahgroup/godcr/cli/termio/terminalprompt"
 )
 
+const (
+	errMsgInputNotEnoughForTxn = "total input amount not enough to cover transaction"
+	errMsgInputNotEnoughForFee = "total input amount not enough to cover transaction fee"
```
itswisdomagain: These constants are not necessary, the second one is never used and the first was wrongly used, it actually shouldn't be used either.

cli/commands/helpers.go
```diff
@@ -113,9 +119,10 @@ func getSendTxDestinations(wallet walletcore.Wallet) (destinations []txhelper.Tr
 
 		sendAmount, err := getSendAmount()
 		if err != nil {
-			return nil, fmt.Errorf("error receiving input: %s", err.Error())
+			return nil, 0, fmt.Errorf("error receiving input: %s", err.Error())
 		}
 		sendAmountAddressMap[destinationAddress] = sendAmount
+		sendAmountTotal += sendAmount
```
itswisdomagain: This should be done in the loop in line 130 to prevent the possibility of incorrect sum in the event an address is entered multiple times and the amount updated. Line 130 is where the final map of addresses and amounts collected from the users is processed.

cli/commands/helpers.go
```diff
@@ -151,17 +158,162 @@ func getSendAmount() (float64, error) {
 	return amount, nil
 }
 
+// getChangeOutputDestinations fetches the amount to be sent to each change address
+func getChangeOutputDestinations(wallet walletcore.Wallet, totalInputAmount float64, sourceAccount uint32,
+	nUtxoSelection int, sendDestinations []txhelper.TransactionDestination) (changeOutputDestinations []txhelper.TransactionDestination, err error) {
```
itswisdomagain: No need to use named parameters, see [this unresolved comment on line 213 of helpers.go](https://github.com/raedahgroup/godcr/pull/123#discussion_r245670635) and [this other one on line 233 of same file](https://github.com/raedahgroup/godcr/pull/123/files#r245670953)

itswisdomagain: Add an empty line below this line to separate where the function signature ends and where the body begins.

cli/commands/helpers.go
```diff
@@ -151,17 +158,162 @@ func getSendAmount() (float64, error) {
...
+
+	changeAmount, err := txhelper.EstimateChange(nUtxoSelection, amountInAtom, sendDestinations, changeAddresses)
+	if err != nil {
+		return nil, fmt.Errorf("error in getting change amount: %s", err.Error())
+	}
+	if changeAmount <= 0 {
+		return
+	}
+
+	var portions []float64
```
itswisdomagain: Unused variable

cli/commands/helpers.go
```diff
@@ -151,17 +158,162 @@ func getSendAmount() (float64, error) {
...
+			return nil, fmt.Errorf("error in generating address: %s", err.Error())
+		}
+		changeAddresses = append(changeAddresses, address)
+		totalChangeAmount, err := txhelper.EstimateChange(nUtxoSelection, amountInAtom, sendDestinations, changeAddresses)
+		defaultAmount := totalChangeAmount - amountAssigned
+
+		prompt := fmt.Sprintf("[%d]: Change Amount (DCR) (default: %f)", index+1, dcrutil.Amount(defaultAmount).ToCoin())
+		changeAmount, err := getChangeAmount(prompt)
+		if err != nil {
+			// we cannot just use return here as the named varaible `err` has been shadowed in line 207 above
```
itswisdomagain: Comment no longer necessary

cli/commands/send.go
```diff
@@ -47,113 +46,137 @@ func send(wallet walletcore.Wallet, custom bool) error {
...
-		var totalInputAmount int64
-		for _, utxo := range utxoSelection {
-			utxos = append(utxos, utxo.OutputKey)
-			totalInputAmount += int64(utxo.Amount)
-		}
+	var outputKeys []string
+	for _, utxo := range utxoSelection {
+		outputKeys = append(outputKeys, utxo.OutputKey)
+	}
+	sentTransactionHash, err = wallet.SendFromUTXOs(sourceAccount, outputKeys, sendDestinations, changeOutputDestinations, passphrase)
```
itswisdomagain: You can just return `wallet.SendFromUTXOs(...)` directly, no need to check if there was an error, the calling function does that check

cli/commands/send.go
```diff
@@ -47,113 +46,137 @@ func send(wallet walletcore.Wallet, custom bool) error {
...
+	}
 
-		sentTransactionHash, err = wallet.SendFromUTXOs(sourceAccount, utxos, sendDestinations, changeDestinations, passphrase)
+	if len(sendDestinations) == 1 {
+		fmt.Println(fmt.Sprintf("You are about to send %f DCR to %s", sendDestinations[0].Amount, sendDestinations[0].Address))
 	} else {
-		sentTransactionHash, err = wallet.SendFromAccount(sourceAccount, sendDestinations, passphrase)
+		fmt.Println("You are about to send")
+		for _, destination := range sendDestinations {
+			fmt.Println(fmt.Sprintf(" %f DCR to %s", destination.Amount, destination.Address))
```
itswisdomagain: tabs here as well

cli/commands/send.go
```diff
@@ -47,113 +46,137 @@ func send(wallet walletcore.Wallet, custom bool) error {
...
 	}
 
-	fmt.Println("Sent. Txid", sentTransactionHash)
-	return nil
+	if !sendConfirmed {
+		return "",errors.New("transaction cancelled")
+	}
+
+	var sentTransactionHash string
+	sentTransactionHash, err = wallet.SendFromAccount(sourceAccount, sendDestinations, passphrase)
```
itswisdomagain: `return wallet.SendFromAccount(...)` directly

itswisdomagain (changes requested):
Some more changes. Also, I marked some comments from the previous review as resolved. Look at the other ones that I have not marked resolved, you didn't attend to them

ademuanthony:
change-ouptut

itswisdomagain (approved):

Issue #96, opened 2018-12-30
dcrwallet returns incorrect data for a transaction's block hash


barisere:
On [PR 83](http://github.com/raedahgroup/godcr/pull/83), the block hash returned by dcrwallet is the same as the transaction hash. This has been confirmed as a bug on the dcrwallet project, and [an issue has been submitted](https://github.com/decred/dcrwallet/issues/1365). This bug shall be documented for now, and this issue shall await a fix from the dcrwallet project.

barisere:
This issue has been fixed in dcrwallet https://github.com/decred/dcrwallet/pull/1367. You need to update `dcrwallet`, and maybe `dcrd`, to get the fix; updating both to the latest versions worked for me.  See the output now has a correct block hash.
```
$ ./godcr show-transaction dc8f20a53ab10246abf19b78296e183f7cdd84040f48f3fd8b05b21068e33ca1
Looking for wallets...
Hash              dc8f20a53ab10246abf19b78296e183f7cdd84040f48f3fd8b05b21068e33ca1
Confirmations     1528
Included in block 000000000022476e6de9b642986624d87e7fd05e18b7cebdd5cca5a82dac1049
Type              REVOCATION
Amount sent       0 DCR
Date              Fri Jan 4, 2019 6:48PM
Size              0.2 kB
Fee               0.00022 DCR
Rate              0.00101851 DCR/kB
```

itswisdomagain:
Latest `dcrd` and `dcrwallet` binaries are available for download [here: https://github.com/decred/decred-binaries/releases](https://github.com/decred/decred-binaries/releases)

PR #127, opened 2019-01-05
implement dcrlibwallet get tx, improve same function for wallet rpc 

itswisdomagain:
The output of cli `show-transaction` via the updated rpc function or the newly implemented dcrlibwallet function. Major changes are addition of account name when displaying outputs and displaying block height instead of block hash. The block height is gotten using the block hash and since the blockhash is invalid (see #96), the block height is not currently found. A default value of -1 is displayed whenever the block hash cannot be used to get a matching block height.

screenshot 2019-01-05 at 8 17 19 am

UPDATE:
dcrwallet rpc block hash issue is resolved, and block heights for transactions now display


ademuanthony (changes requested):
When the transaction hash is not found, the error in the output is too verbose, can we catch the error code from the response just print transaction not found?
not-found
I suspect the error to be of `dcrdwallet/errors.Error`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -324,18 +354,29 @@ func (c *WalletPRCClient) TransactionHistory() ([]*walletcore.Transaction, error
...
-	getTxResponse, err := c.walletService.GetTransaction(context.Background(), getTxRequest)
+	getTxResponse, err := c.walletService.GetTransaction(ctx, getTxRequest)
 	if err != nil {
 		return nil, err
 	}
 
-	transactionHex := fmt.Sprintf("%x", getTxResponse.GetTransaction().GetTransaction())
-	msgTx, err := txhelpers.MsgTxFromHex(transactionHex)
+	// func to attempt to check if an address belongs to the wallet to retrieve it's account name or return external if not
+	getWalletAddressInfo := func(address string) *txhelper.AddressInfo {
```
barisere: This function looks like an exact copy of [this one here](https://github.com/raedahgroup/godcr/pull/127/files#diff-b081aa982693a57870847580a9b9cd71R266). The can be extracted into a package-level function.

barisere (changes requested):
#96 has been fixed, see https://github.com/raedahgroup/godcr/issues/96#issuecomment-451650393

itswisdomagain:
Block height now display with the latest versions of dcrd and dcrwallet.
screenshot 2019-01-05 at 9 42 28 pm

Transaction not found errors now show simpler error message:
`transaction not found`

app/walletmediums/dcrwalletrpc/walletfunctions.go
```diff
@@ -379,6 +382,7 @@ func (c *WalletPRCClient) GetTransaction(transactionHash string) (*walletcore.Tr
 
 	var blockHeight int32 = -1
 	if getTxResponse.BlockHash != nil {
+		//fmt.Printf("%x\n", getTxResponse.GetBlockHash())
```
barisere: The comment is not needed any longer.

barisere (changes requested):

ademuanthony (approved):

barisere (approved):

Issue #124, opened 2019-01-04
eliminate some command-line options


itswisdomagain:
Recent work done in #109 as fix to #106 separated config file options from command-line options. Config file options are options or flags that we expect the user to set in their `godcr.conf` file, while command-line options are options/flags we permit users to pass as command-line options when executing a command.

Currently, godcr accepts config file options in command-line. This shouldn't be so. Options like `usewalletrpc` and `testnet` should ONLY be set in `godcr.conf`. If a user passes such options/flags in command-line, print a message informing the user to set such options in `godcr.conf` and not pass them in command-line.

PR #135, opened 2019-01-06
Eliminate config file option from command line

ademuanthony:
Resolves [Issues](https://github.com/raedahgroup/godcr/issues/124)

stop config file options from being passed in command line args
cong-cli


itswisdomagain (changes requested):
This looks good. Bring back the `go.sum` file so I can test and probably merge

itswisdomagain (approved):

PR #141, opened 2019-01-08
separate ReceiveAddress from GenerateAddress

itswisdomagain:
Currently, there is one core wallet function for generating address in a wallet's account: `GenerateReceiveAddress`. This function is implemented differently for `dcrlibwallet` and `dcrwalletrpc`.

While `dcrwalletrpc` almost always generates a new address every time the function is called, `dcrlibwallet` only generates a new address if the previous address has been used to receive funds. This would cause problems with the new feature being implemented in #123 to support multiple change outputs, where each change output requires generation of a unique address even though the previously generated address is yet to be used to receive funds.

Still, when a user requests an address to use for funds receipt, by default, godcr should not generate a new address unless the previously generated address has been used. However, if the user really wants a new address generated, we can make provision for that in the different interfaces. A new issue will cover that scenario.

For now, this PR comes with the following changes:
- 2 new core wallet functions: `ReceiveAddress` and `GenerateNewAddress`
- Receive feature uses `ReceiveAddress` which may not always generate a new address. Ideally, it would return previously generated and unused address
- Send feature uses `GenerateNewAddress` when dealing with change outputs to prevent address re-use because no 2 change outputs for a single transaction should use the same address

### DIFF
--- a/app/config/config.go
+++ b/app/config/config.go
@@ -99,9 +99,9 @@ func LoadConfig(ignoreUnknownOptions bool) ([]string, Config, *flags.Parser, err
 			optionName = arg[1:]
 		}
 		if isFileOption := isConfigFileOption(optionName); isFileOption {
-			return args, config, parser, fmt.Errorf("Unexpected command-line flag/option, " +
-				"see godcr -h for supported command-line flags/options" +
-				"\nSet other flags/options in %s", AppConfigFilePath)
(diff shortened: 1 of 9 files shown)


ademuanthony (changes requested):
Clean work.
You have to rebase onto master, fix conflict and test again
also note; app/config/config.go:104:40: undefined: configFilePath

ademuanthony (approved):

Issue #144, opened 2019-01-08
can't use godcr with missing godcr.conf file


itswisdomagain:
Currently, if the `godcr.conf` file does not exist in the app data dir, the godcr program displays an error message similar to this:
`open ~/godcr/godcr.conf: no such file or directory`

Instead of throwing an error, the desired behaviour is to create the config file using the contents of the `sample-godcr.conf` file. This file may need to be refactored into a `.go` file so that its content is included with the godcr binary when building. That way, when the godcr binary is run and there is no config file found, the program will have the sample config data it needs to create a config file.

The default config file created for the user should use default values specified in `app/config/config.go` and comment out options/flags without a default value. User should not be required to provide any information in a prompt. Instead, a simple message can be displayed to the user to inform the user that the config file was not found but it has been created with default values.

PR #145, opened 2019-01-09
create new config file if missing

ademuanthony:
Resolves [Issue 144](https://github.com/raedahgroup/godcr/issues/144)
Currently, if the godcr.conf file is not found in the app data dir, the godcr program displays an error message similar to this:
`open ~/godcr/godcr.conf: no such file or directory`
This PR solve this solves this problem by creating a new config file using the default values and notifies the user of the action.
missing-config


app/config/config.go
```diff
@@ -152,3 +156,27 @@ func isConfigFileOption(name string) ( isFileOption bool) {
 	}
 	return
 }
+
+// createConfigFile create the configuration file using the default values in the AppConfigFilePath
```
itswisdomagain: Are the default values in AppConfigFilePath?
Or do you mean it creates the file in `AppConfigFilePath` using default values

app/config/config.go
```diff
@@ -152,3 +156,27 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+func createConfigFile() error {
+	configFile, err := os.Create(AppConfigFilePath)
```
itswisdomagain: `error in creating config file: open /Users/itswisdomagain/Library/Application Support/Godcr/godcr.conf: no such file or directory`

Make provision for scenarios where the appdata folder is not created

app/config/config.go
```diff
@@ -152,3 +156,27 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+	if err != nil {
+		return fmt.Errorf("error in creating config file: %v", err)
```
itswisdomagain: `%s` instead of `%v`

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
+package config
+
+func configText() string {
```
itswisdomagain: - Rename this file to `godcr.conf.go`
- Delete `sample-godcr.conf`, leaving this file as the single source of truth for sample config file content
- To achieve simplicity similar to what we had in `sample-godcr.conf`, this should be a constant, the text beginning at the next line like this:
```
const configTextTemplate =
`[Application Options]

; ------------------------------------------------------------------------------
; Network settings
; ------------------------------------------------------------------------------
```

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; ------------------------------------------------------------------------------
+; Network settings
+; ------------------------------------------------------------------------------
+
+; Connect to a running drcwallet daemon over rpc to perform wallet operations.
+; By default godcr uses dcrlibwallet (usewalletrpc=false or usewalletrpc=0).
+; You can switch to dcrwallet rpc using (usewalletrpc=true or usewalletrpc=1).
+; usewalletrpc=false
+
+; Wallet RPC server
```
itswisdomagain: `dcrwallet gRPC server address. Required if usewalletrpc=true`

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; walletrpcserver=localhost:19111
+
+; Specify dcrwallet's RPC certificate, or disable TLS for the connection
```
itswisdomagain: Another way to put lines 18-20 more cleanly is:
```
; Disable TLS for rpc connection (nowalletrpctls=1) or specify path to dcrwallet's RPC certificate
; nowalletrpctls=0
walletrpccert={{.WalletRPCCert}}
```

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+walletrpccert={{.WalletRPCCert}}
+; nowalletrpctls=0
+
+;Connects to testnet wallet instead of mainnet
```
itswisdomagain: space => `; Co...`

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; testnet=false
+
+; ------------------------------------------------------------------------------
+; HTTP Server
```
itswisdomagain: `Godcr HTTP Mode Options`

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; Wallet RPC server
+; walletrpcserver=localhost:19111
+
+; Specify dcrwallet's RPC certificate, or disable TLS for the connection
+walletrpccert={{.WalletRPCCert}}
+; nowalletrpctls=0
+
+;Connects to testnet wallet instead of mainnet
+; testnet=false
+
```
itswisdomagain: The desktop mode and http mode interface options in lines 33 to 43 are deprecated. We now have just one option for specifying interface mode: `mode=[http/cli/desktop]`.

This information should be included here, before interface-specific options (e.g. http server options) are mentioned:
```
; ------------------------------------------------------------------------------
; Godcr Interface Modes
; ------------------------------------------------------------------------------
; godcr can run in any of the following modes: cli, http, desktop
; mode=cli
```

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; ------------------------------------------------------------------------------
+; HTTP Server
+; ------------------------------------------------------------------------------
+
+; Server address to serve godcr via HTTP (when http=true or http=1)
```
itswisdomagain: `host and port for godcr http server when mode=http`

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+httphost={{.HTTPHost}}
+httpport={{.HTTPPort}}
+
+; Turn HTTP mode on (http=true or http=1) or off (http=false or http=0).
```
itswisdomagain: This point to line 43 are no more relevant, should be deleted

app/config/godcr.config.go
```diff
@@ -0,0 +1,51 @@
...
+; Turn Desktop mode on (desktop=true or desktop=1) or off (desktop=false or desktop=0).
+; Desktop mode is off by default.
+; desktop=false
+
+; ------------------------------------------------------------------------------
+; App Data
+; ------------------------------------------------------------------------------
+
+; Path to application data directory.
+appdata={{.AppDataDir}}`
```
itswisdomagain: Should probably be at the top after `[Application Options]` and before `Network settings`

app/config/config.go
```diff
@@ -152,3 +156,27 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+	}
+	defer configFile.Close()
+
+	tmpl := template.New("config")
+
+	tmpl, err = tmpl.Parse(configText())
+	if err != nil {
+		return fmt.Errorf("error in parsing the config template: %v", err.Error())
```
itswisdomagain: `%s` instead of `%v`

itswisdomagain: since we're displaying this to user, the message could be simplified:
`error preparing default config file content`

app/config/config.go
```diff
@@ -152,3 +156,27 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+	}
+
+	err = tmpl.Execute(configFile, defaultFileOptions())
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("The config file was not found but it has been created with the default values at %s\n", AppConfigFilePath)
```
itswisdomagain: Simplified: `config file created with default values at %s`

app/config/config.go
```diff
@@ -131,7 +132,10 @@ func parseConfigFile(parser *flags.Parser) error {
 		if _, ok := err.(*os.PathError); !ok {
 			return fmt.Errorf("Error parsing configuration file: %v", err.Error())
 		}
-		return err
+		if err = createConfigFile(); err != nil {
+			return err
```
itswisdomagain: display the error: `fmt.Println(err.Error())` but return nil, we should not stop program execution because config file was not found and could not be created

a simple message as to what went wrong should suffice, then the program execution continues

itswisdomagain (changes requested):
Some needed changes.

The readme also needs to be updated to dereference `sample-godcr.conf`. Could use the following for the `Running godcr` section of the readme:

--------
## Running godcr
### General usage
```bash
$ godcr [options] <command> [args]
```

### Commands and options
* Use `godcr -h` or `godcr help` to view options, commands and path to the godcr config file
* Some options can only be set in the godcr configuration (godcr.conf) file. Those options are not displayed in the output of `godcr -h`. Edit the config file to view and set those options.
* Use `godcr <command> -h` or `godcr help <command>` to view command args and detailed help information for a command

app/config/config.go
```diff
@@ -131,12 +133,16 @@ func parseConfigFile(parser *flags.Parser) error {
 		if _, ok := err.(*os.PathError); !ok {
 			return fmt.Errorf("Error parsing configuration file: %v", err.Error())
 		}
-		return err
+		if err = createConfigFile(); err != nil {
+			fmt.Println(fmt.Sprintf("config file was not found and could not be created: %s", err))
```
itswisdomagain: `err` returned from `createConfigFile()` may be descriptive already, adding this text here makes it unnecessarily more verbose

an alternative thought is to print all errors in the `createConfigFile` function and modify the function to return a boolean, true if the operation was successful, false if not

app/config/config.go
```diff
@@ -152,3 +158,33 @@ func isConfigFileOption(name string) ( isFileOption bool) {
 	}
 	return
 }
+
+// createConfigFile create the configuration file in AppConfigFilePath using the default values
+func createConfigFile() error {
+	configFile, err := os.Create(AppConfigFilePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("error in creating config file: %s", err)
```
itswisdomagain: Just seeing this, `err` should be `err.Error()`
Also, as mentioned in a comment above, simply print the error here and return. Use a named return parameter so that the return value is false by default, and only true at the end of the program where you can `return true`

app/config/config.go
```diff
@@ -131,12 +133,16 @@ func parseConfigFile(parser *flags.Parser) error {
 		if _, ok := err.(*os.PathError); !ok {
 			return fmt.Errorf("Error parsing configuration file: %v", err.Error())
```
itswisdomagain: `%s` instead of `%v`, `err.Error()` is of type string

app/config/config.go
itswisdomagain: This section could then look like this:
```
if createConfigFile() {
	return parseConfigFile(parser)
}
```

app/config/config.go
```diff
@@ -152,3 +158,33 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+
+// createConfigFile create the configuration file in AppConfigFilePath using the default values
+func createConfigFile() error {
+	configFile, err := os.Create(AppConfigFilePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			return fmt.Errorf("error in creating config file: %s", err)
+		}
+		err = os.Mkdir(appConfigDir, os.ModePerm)
+		if err != nil {
```
itswisdomagain: Successfully creating the directory does not necessarily mean the config file is created and ready for use. You'll need to re-attempt to create the file and only proceed if the file creation was successful

app/config/config.go
itswisdomagain: For clarity, here's what I mean:
```
// createConfigFile create the configuration file in AppConfigFilePath using the default values
func createConfigFile() (successful bool) {
	configFile, err := os.Create(AppConfigFilePath)
	if err != nil {
		if !os.IsNotExist(err) {
			fmt.Fprintf(os.Stderr,"error in creating config file: %s\n", err.Error())
			return
		}
		err = os.Mkdir(appConfigDir, os.ModePerm)
		if err != nil {
			fmt.Fprintf(os.Stderr,"error in creating appdata directory: %s\n", err.Error())
			return
		}
	}
	defer configFile.Close()

	tmpl := template.New("config")

	tmpl, err = tmpl.Parse(configTextTemplate)
	if err != nil {
		fmt.Fprintf(os.Stderr, "error preparing default config file content: %s", err.Error())
		return
	}

	err = tmpl.Execute(configFile, defaultFileOptions())
	if err != nil {
		fmt.Fprintf(os.Stderr, "error saving default configuration to file: %s\n", err.Error())
		return
	}

	fmt.Println("Config file created with default values at", AppConfigFilePath)
	return true
}
```

itswisdomagain (changes requested):
Additional notes:
- The config file should be created (if it doesn't exist) before the `Unexpected command-line flag/option` message is displayed. That message points users to the config file and it'd be quite a hassle if users go there and can't find a file

README.md
```diff
@@ -40,12 +40,13 @@ $ GO111MODULE=on go install
...
 
 ## Running godcr
-* Create the godcr configuration file by copying the sample configuration file (`sample-godcr.conf`) to the default config location and editing as necessary. The default config location can be gotten by running `godcr -h`.
-The settings in the configuration file can also be provided on the command line as options to the program. Flags passed on the command line override those set in the configuration file. Run `godcr -h` to see the available options.
-* Once you've configured godcr, you can perform various wallet-related operations by running
+* You can perform various wallet-related operations by running
 ```bash
 $ godcr [options] <command> [args]
 ```
+* Running the app for the first time will create the config file in the default config location. This file can be edited as necessary.
```
itswisdomagain: > Running the app for the first time will create **_the_** config file in the **_default_** config location

3 questions will arise in the minds of readers that are new to godcr:
- What's **_the_** config file?
- Where's the default config location?
- Can I change the config location? (You used the word "default config location", suggesting it can be changed...)

The second question is answered below, for the other questions, new contributors will have to find out on their own. We can simplify, eliminate or reduce the questions new folks will have after reading this readme

README.md
```diff
@@ -40,12 +40,13 @@ $ GO111MODULE=on go install
...
+Note that not all settings in the configuration file can be provided on the command line as options to the program. Run `godcr -h` to see the available command line options or view default config location.
```
itswisdomagain: Ideally, no setting on the config file should be provided on the command line. Options in command-line can be provided in config file, but we shouldn't even say so, it's not recommended. The recommended approach is to keep config options distinct from command-line.

Currently, the program will not reject any of the 2 command-line options (`mode` and `sync`) if they're found in the config file. We can keep that behaviour for now, but we'll likely need to review it soon enough. Until then, even though we accept command-line options in `godcr.conf`, the desired approach we should remember is:
> keep config options distinct from command-line options

It'd be more correct to say, some options are supported in the command-line whereas others have to be set in `godcr.conf`

app/config/config.go
```diff
@@ -152,3 +155,43 @@ func isConfigFileOption(name string) ( isFileOption bool) {
 	}
 	return
 }
+
+// createConfigFile create the configuration file in AppConfigFilePath using the default values
+// createConfigFile create the configuration file in AppConfigFilePath using the default values
```
itswisdomagain, line 127: duplicated comment

app/config/config.go
```diff
@@ -152,3 +155,43 @@ func isConfigFileOption(name string) ( isFileOption bool) {
...
+		if !os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr,"error in creating config file: %s\n", err.Error())
+			return
+		}
+		err = os.Mkdir(defaultAppDataDir, os.ModePerm)
+		if err != nil {
+			fmt.Fprintf(os.Stderr,"error in creating appdata directory: %s\n", err.Error())
+			return
+		}
+		configFile, err = os.Create(AppConfigFilePath)
```
itswisdomagain, line 142: add a comment here to show why re-attempting to create the file is necessary, afterall, we just tried to create the same file some lines of code ago

app/config/godcr.conf.go
```diff
@@ -0,0 +1,45 @@
...
+; Connects to testnet wallet instead of mainnet
+; testnet=false
+
+; ------------------------------------------------------------------------------
+; Godcr Interface Modes
+; ------------------------------------------------------------------------------
+
+; godcr can run in any of the following modes: cli, http, desktop
+; mode=cli
+
```
itswisdomagain, line 37: double empty lines here, use 1 to keep the style used throughout this file

itswisdomagain (changes requested):

README.md
```diff
@@ -39,21 +39,17 @@ $ GO111MODULE=on go build
 $ GO111MODULE=on go install
 ```
 
+--------
```
itswisdomagain: this line isn't necessary, I used it in my comment to show where the suggested readme changes began

itswisdomagain (changes requested):

PR #153, opened 2019-01-10
create new config file if missing

ademuanthony:
Resolves [Issue 144](https://github.com/raedahgroup/godcr/issues/144)
Currently, if the godcr.conf file is not found in the app data dir, the godcr program displays an error message similar to this:
`open ~/godcr/godcr.conf: no such file or directory`
This PR solve this solves this problem by creating a new config file using the default values and notifies the user of the action.
missing-config

### DIFF
--- a/app/config/config.go
+++ b/app/config/config.go
@@ -6,6 +6,7 @@ import (
 	"path/filepath"
 	"reflect"
 	"strings"
+	"text/template"
 
 	"github.com/decred/dcrd/dcrutil"
 	flags "github.com/jessevdk/go-flags"
@@ -75,6 +76,13 @@ func defaultConfig() Config {
 // If ignoreUnknownOptions is true, then unknown options seen on the command line are ignored.
 // However, unknown options in the configuration file must return an error.
 func LoadConfig(ignoreUnknownOptions bool) ([]string, Config, *flags.Parser, error) {
+	var configFileExists bool
+	if _, err := os.Stat(AppConfigFilePath); os.IsNotExist(err) {
+		configFileExists = createConfigFile()
+	} else if !os.IsNotExist(err) {
+		configFileExists = true
+	}
+
 	// load defaults first
 	config := defaultConfig()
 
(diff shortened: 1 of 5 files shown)


app/config/config.go
```diff
@@ -75,6 +76,10 @@ func defaultConfig() Config {
 // If ignoreUnknownOptions is true, then unknown options seen on the command line are ignored.
 // However, unknown options in the configuration file must return an error.
 func LoadConfig(ignoreUnknownOptions bool) ([]string, Config, *flags.Parser, error) {
+	if _, err := os.Stat(AppConfigFilePath); os.IsNotExist(err) {
+		createConfigFile()
```
itswisdomagain: This could return an error and the file would not be created.
Later in line 114, `parseConfigFile` may return an error because the file does not exist; and we don't want that.
What's desired is that, `parseConfigFile` should ONLY be called if we know for a fact that the config file exists or if it doesn't exist, that is has been created.

Building on what you have, here's an idea:
```
var configFileExists bool
if _, err := os.Stat(AppConfigFilePath); os.IsNotExist(err) {
	configFileExists = createConfigFile()
} else if !os.IsNotExist(err) {
	configFileExists = true
}
```

Then before calling `parseConfigFile`, first check if the file exists:
```
// if config file doesn't exist, no need to attempt to parse and then re-parse command-line args
if !configFileExists {
	return args, config, parser, nil
}
```

app/config/config.go
```diff
@@ -120,6 +125,47 @@ func LoadConfig(ignoreUnknownOptions bool) ([]string, Config, *flags.Parser, err
...
+func createConfigFile() (successful bool) {
+	configFile, err := os.Create(AppConfigFilePath)
+	if err != nil {
+		if !os.IsNotExist(err) {
+			fmt.Fprintf(os.Stderr,"error in creating config file: %s\n", err.Error())
+			return
+		}
+		err = os.Mkdir(defaultAppDataDir, os.ModePerm)
+		if err != nil {
+			fmt.Fprintf(os.Stderr,"error in creating appdata directory: %s\n", err.Error())
```
itswisdomagain: `error in creating config file directory`

the config directory (which is `defaultAppDataDir`) is not necessarily the same as the appdata directory, because the latter may be modified by the user (`config.AppDataDir`)

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #154, opened 2019-01-11
auto-detect network type when using drcwallet rpc

### LINKED ISSUE #100: auto-detect network type when using drcwallet rpc
This is one of 2 efforts to eliminate the need for a network type flag/option in godcr. The other effort is described in #89. Fixing this issue and #89 should totally render the `testnet` config flag unnecessary and the flag should be removed.

There are 2 current usages of the `testnet` flag in dcrwallet rpc:
- For "constructing" a dcrwallet rpc server address if user does not provide one. [See here](https://github.com/raedahgroup/godcr/blob/master/app/walletmediums/dcrwalletrpc/dcrwalletrpc.go#L71).
- For defining the network parameters for the connected wallet. [See here](https://github.com/raedahgroup/godcr/blob/master/app/walletmediums/dcrwalletrpc/dcrwalletrpc.go#L56). The defined network parameter is used by different core wallet functions to differentiate addresses and keys for one network from those intended for use on another network.

For the first use case, if the user does not provide a rpc server address, the immediate solution is to display an error message to the user, asking the user to set the dcrwallet rpc server address in config file. See #152

For the second use case, the network type can be determined once godcr is successfully connected to dcrwallet via gRPC. [See this](https://github.com/decred/dcrwallet/blob/master/rpc/documentation/api.md#network). Once the network type is determined, the network params can be defined.

ademuanthony:
Resolve #100

This work removes the need for setting testnet option in config flag when using dcrwallet rpc.
We now depend on the supplied address and a successful connection to get the network type.
If the address is not specified, the execution will terminate. The function for getting default address has also been removed


app/config/config.go
```diff
@@ -124,6 +125,10 @@ func LoadConfig(ignoreUnknownOptions bool) ([]string, Config, *flags.Parser, err
 		return args, config, parser, err
 	}
 
+	if config.UseWalletRPC && config.WalletRPCServer == "" {
+		return args, config, parser, errors.New("you must set walletrpcserver to use wallet rpc in config file")
```
itswisdomagain: `you must set walletrpcserver in config file to use wallet rpc`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -53,21 +51,45 @@ func New(ctx context.Context, rpcAddress, rpcCert string, noTLS, isTestnet bool)
...
-			walletService: walletrpc.NewWalletServiceClient(connectionResult.conn),
+			walletService: walletService,
 			activeNet:     activeNet,
 		}
 
 		return client, nil
 	}
 }
 
+func isTestNet(walletService walletrpc.WalletServiceClient) (bool, error) {
```
itswisdomagain: Modify this method to return the active net params `*chaincfg.Params, error` instead of simply checking for testnet/mainnet.

This will make it easier to add support for other network types when the need arises

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -53,21 +51,45 @@ func New(ctx context.Context, rpcAddress, rpcCert string, noTLS, isTestnet bool)
...
+	req := &walletrpc.NetworkRequest{}
+	res, err := walletService.Network(context.Background(), req)
+	if err != nil {
+		return false, err
```
itswisdomagain: `fmt.Errorf("error checking wallet rpc network type: %s", err.Error())`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -53,21 +51,45 @@ func New(ctx context.Context, rpcAddress, rpcCert string, noTLS, isTestnet bool)
...
+	}
+
+	switch res.GetActiveNetwork() {
+	default:
```
itswisdomagain: Make default the last option, makes for easier readability

itswisdomagain (changes requested):
Looks very good. Just these few changes

itswisdomagain (approved):

Issue #152, opened 2019-01-09
auto-detect dcrwallet rpc server address


itswisdomagain:
This issue applies to scenarios where a user runs godcr with `usewalletrpc=true`, but does not provide a dcrwallet rpc server address

- Currently, an address is "constructed" using the `testnet` config flag. [See here](https://github.com/raedahgroup/godcr/blob/master/app/walletmediums/dcrwalletrpc/dcrwalletrpc.go#L71)
- After a fix to #100 is merged, a "default" address will no longer be constructed, instead an error message will be displayed to the user

The desired behaviour is to try to "detect" the rpc server address before showing the user an error message, if an address cannot be detected.

Here are some ideas to explore for auto-detecting dcrwallet rpc server address:

1. Check running processes to see if dcrwallet is currently running. If it is, determine the address (host and port) that the running dcrwallet process is bound to.

2. Attempt to connect to dcrwallet using a [testnet-constructed address](https://github.com/raedahgroup/godcr/blob/master/app/walletmediums/dcrwalletrpc/dcrwalletrpc.go#L73). If that doesn't connect, perform a second connection attempt using a [mainnet-constructed address](https://github.com/raedahgroup/godcr/blob/master/app/walletmediums/dcrwalletrpc/dcrwalletrpc.go#L75)

3. Locate and parse the dcrwallet config file to determine the dcrwallet rpc server address configured by the user. See [here](https://github.com/decred/dcrwallet/blob/master/config.go#L78) for the default path to the config file, and [here](https://github.com/decred/dcrwallet/blob/master/sample-dcrwallet.conf#L93-L107) for how to extract the address from the config file.

Each of the above possibilities have their strengths and weaknesses, it _might_ be a good idea to combine multiple approaches in a bid to auto-detect the dcrwallet rpc server address.

However it is very possible that a **valid** rpc server address cannot be detected - perhaps because the user has not started the dcrwallet daemon or probably doesn't even have the dcrwallet binary. Whatever the case, the user should be asked to confirm that dcrwallet is running and if it is running, the user should be allowed to set the rpc server address directly from the terminal.

```
> Could not detect a valid rpc address to connect to dcrwallet
> Do you want to set the address now? [y/n]: y
> Enter dcrwallet rpc address: 127.0.0.1:19111
> rpc address saved, connecting...
```

Or

```
> Could not detect a valid rpc address to connect to dcrwallet
> Do you want to set the address now? [y/n]: n
> Okay. Bye.
[...]

PR #156, opened 2019-01-11
auto detecting rpc address

ademuanthony:
Resolves [Issue 152](https://github.com/raedahgroup/godcr/issues/152)

Added the ability to detect dcrwallet RPC address from a dcrwallet running process, the default wallet addresses and/or the dcrwallet config file.

The user is asked to enter the address manually if all these options fail. The captured address will be added to the config file for subsequent use
auto-detect-address


cli/termio/terminalprompt/terminalprompt.go
```diff
@@ -15,6 +16,14 @@ var EmptyValidator = func(v string) error {
 	return nil
 }
 
+// RequiredValidator throws if the input value is empty
+var RequiredValidator = func(v string) error {
```
itswisdomagain: `InputRequiredValidator`

main.go
```diff
@@ -63,19 +66,72 @@ func main() {
...
 	}
 
-	if !config.UseWalletRPC {
-		return dcrlibwallet.New(config.AppDataDir, netType)
+	if !_config.UseWalletRPC {
+		return dcrlibwallet.New(_config.AppDataDir, netType)
 	}
-
-	walletMiddleware, err := dcrwalletrpc.New(ctx, config.WalletRPCServer, config.WalletRPCCert, config.NoWalletRPCTLS, config.UseTestNet)
+	if _config.WalletRPCServer == "" {
```
itswisdomagain: Move this whole op to `app/walletmediums/dcrwalletrpc`, it's specific to that medium

app/address.go
```diff
@@ -0,0 +1,11 @@
+package app
```
itswisdomagain: This file should be in `app/walletmediums/dcrwalletrpc`, it contains code specific to that medium

app/config/config.go
```diff
@@ -206,3 +206,9 @@ func isConfigFileOption(name string) (isFileOption bool) {
 	}
 	return
 }
+
+func SaveRpcAddress(address string) error {
```
itswisdomagain: This can be made more generic, I'm thinking a function that can accept any form of update to config file: `UpdateConfigFile(option string, newValue interface{})`

main.go
```diff
@@ -63,19 +66,72 @@ func main() {
 
 // connectToWallet opens connection to a wallet via any of the available walletmiddleware
 // default walletmiddleware is dcrlibwallet, alternative is dcrwalletrpc
-func connectToWallet(ctx context.Context, config config.Config) app.WalletMiddleware {
+func connectToWallet(ctx context.Context, _config config.Config) app.WalletMiddleware {
```
itswisdomagain: maybe `cfg` instead of `_config` or `appConfig` as used in some other parts of this file or `parsedConfig` to signify that it has been parsed and values set for it's various fields

itswisdomagain (changes requested):

app/walletmediums/dcrwalletrpc/address.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/jessevdk/go-flags"
+	"github.com/raedahgroup/godcr/app/config"
+)
+
+type ExplicitString struct {
+	Value         string
+	explicitlySet bool
+}
+
+type walletConfig struct {
```
itswisdomagain, line 15: Several fields in this struct can be done without, retaining just what is needed: `GRPCListeners`, `DisableServerTLS` and `RPCCert`

app/walletmediums/dcrwalletrpc/address.go
```diff
@@ -0,0 +1,55 @@
...
+			err = fmt.Errorf("Error parsing configuration file: %v", err.Error())
+			return
+		}
+		return
+	}
+	var rpcCert string
+	if wConfig.RPCCert != nil {
+		rpcCert = wConfig.RPCCert.Value
+	}
+	return wConfig.LegacyRPCListeners, wConfig.DisableServerTLS, rpcCert, nil
```
itswisdomagain: `wConfig.GRPCListeners` instead of `wConfig.LegacyRPCListeners`
LegacyRPCListeners are for JSON-RPC connections, which godcr does not support (yet, maybe never will).

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+	}
+
+	// parse dcrwallet config first, so as to use it's cert path and tls state
+	rpcConfAddresses, wnoTLS, wrpcCert, err := walletAddressFromDcrdwalletConfig()
+	if wrpcCert != "" {
+		rpcCert = wrpcCert
+	}
+	noTLS = wnoTLS
+
+	promt := "No valid RPC address is provided to connect to. Press 'Y' to automatically detect or 'N' to manually enter the address?"
```
itswisdomagain: - typo: _prompt_
- `No RPC address provided to connect to dcrwallet` (No valid address means address was provided but it is not valid)
- this should be the first op in this method, before even parsing the dcrwallet config file

app/walletmediums/dcrwalletrpc/address.go
```diff
@@ -0,0 +1,55 @@
...
+	walletConfigFilePath = filepath.Join(config.DefaultDcrwalletAppDataDir, defaultWalletConfigFilename)
+)
+
+func walletAddressFromDcrdwalletConfig() (addresses []string, notls bool, certpath string, err error) {
+	wConfig := walletConfig{}
+
+	parser := flags.NewParser(&wConfig, flags.IgnoreUnknown)
+	err = flags.NewIniParser(parser).ParseFile(walletConfigFilePath)
+	if err != nil {
+		if _, ok := err.(*os.PathError); !ok {
```
itswisdomagain: simply `return`, we don't need details of what went wrong, we simply need to know if this parsing operation was a success or not

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+		err := config.UpdateConfigFile("walletrpcserver", address, true)
+		if err == nil {
+			fmt.Println("rpc address saved")
+		} else {
+			fmt.Printf("unable to save rpc address: %s", err.Error())
+		}
+	}
+
+	// parse dcrwallet config first, so as to use it's cert path and tls state
+	rpcConfAddresses, wnoTLS, wrpcCert, err := walletAddressFromDcrdwalletConfig()
```
itswisdomagain: you're not using `err`, you should check that `err == nil` before overriding values for `rpcCert` and `noTLS`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
 	}
 }
 
+func autoDetectAddressAndConnect(ctx context.Context, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
+	saveRPCAddress := func(address string) {
+		err := config.UpdateConfigFile("walletrpcserver", address, true)
+		if err == nil {
+			fmt.Println("rpc address saved")
```
itswisdomagain: Only display a message if saving the address fails

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+	}
+	noTLS = wnoTLS
+
+	promt := "No valid RPC address is provided to connect to. Press 'Y' to automatically detect or 'N' to manually enter the address?"
+	automaticDetectionConfirmed, err := terminalprompt.RequestYesNoConfirmation(promt, "y")
+	if err != nil {
+		return nil, fmt.Errorf("error in getting input: %s", err.Error())
+	}
+	if !automaticDetectionConfirmed {
+		address, err := terminalprompt.RequestInput("Enter dcrwallet rpc address", terminalprompt.InputRequiredValidator)
```
itswisdomagain: Move line 125 to 136 to a different function, say `requestRPCAddressFromUser`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+		fmt.Println("connecting...")
+		walletMiddleware, err := New(ctx, address, rpcCert, noTLS)
+		if err == nil {
+			saveRPCAddress(address)
+			return walletMiddleware, nil
+		}
+		return nil, err
+	}
+
+	for {
```
itswisdomagain: This `for` is unnecessary, the code here never loops, only runs once

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+					address = net.JoinHostPort("localhost", port)
+					walletMiddleware, err := New(ctx, address, rpcCert, noTLS)
+					if err == nil {
+						saveRPCAddress(address)
+						return walletMiddleware, nil
+					}
+				}
+			} else {
+				//todo log err to file
+				fmt.Println("error occured while trying to get port by process: %s", err.Error())
```
itswisdomagain: any need? we should swallow all errors, no need to provide details of what went wrong to the user
if at the end of the day, an address could not be detected, we just say address not detected

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+		walletMiddleware, err := New(ctx, address, rpcCert, noTLS)
+		if err == nil {
+			saveRPCAddress(address)
+			return walletMiddleware, nil
+		}
+		return nil, err
+	}
+
+	for {
+		var address string
```
itswisdomagain: No real use for creating a general variable for address. It's usage is limited to different scopes below, never used aside those scopes. Declare, assign and use where needed for simplicity.

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+
+		proc, err := ps.ProcessByName(dcrwalletExecutableName)
```
itswisdomagain: - `dcrwalletProcess` instead of `proc`
- move this line 142 to 159 to separate function, say `detectAddressFromRunningProcess`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+						return walletMiddleware, nil
+					}
+				}
+			} else {
+				//todo log err to file
+				fmt.Println("error occured while trying to get port by process: %s", err.Error())
+			}
+		}
+
+		if err == nil && len(rpcConfAddresses) > 0 {
```
itswisdomagain: No need to check if `err == nil`, `err` here is gotten from line 142 above where `proc, err := ps.ProcessByName(dcrwalletExecutableName)`

Getting the process may return an error, but doesn't mean we should not attempt to connect to addresses found in config file

itswisdomagain: Also move this line 161 to line 169 to a different function, say `useParsedConfigAddresses`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+			for _, address = range rpcConfAddresses {
+				walletMiddleware, err := New(ctx, address, rpcCert, noTLS)
+				if err == nil {
+					saveRPCAddress(address)
+					return walletMiddleware, nil
+				}
+			}
+		}
+
+		// try connecting with default testnet3 params
```
itswisdomagain: move line 171 - 185 to a different function, say `connectToDefaultAddresses` and only call that function if the process of getting ports from running process fails.

If there is no running process or process was found and ports were retrieved, don't see a point in trying the default addresses.

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+				fmt.Println(fmt.Sprintf("error in reading input: %s", err.Error()))
+			}
+			walletMiddleware, err = New(ctx, address, rpcCert, noTLS)
+			if err == nil {
+				saveRPCAddress(address)
+				return walletMiddleware, nil
+			}
+
+			fmt.Println("connecting...")
+			return New(ctx, address, rpcCert, noTLS)
```
itswisdomagain: No need to call `New` again, simply return the `walletMiddleware, err` gotten from the last call in line 199

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+
+		fmt.Println("Could not detect a valid rpc address to connect to dcrwallet")
+		promt := "Do you want to set the address now?"
+		setAddressConfirmed, err := terminalprompt.RequestYesNoConfirmation(promt, "y")
+		if err != nil {
+			fmt.Println(fmt.Sprintf("error in getting input: %s", err.Error()))
+		}
+
+		if setAddressConfirmed {
+			address, err := terminalprompt.RequestInput("Enter dcrwallet rpc address", terminalprompt.InputRequiredValidator)
```
itswisdomagain: Seeing some repeated code here, similar code exists in line 125 to 136, which should be moved to a function and possibly called here

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -91,6 +99,119 @@ func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.P
...
+		// try connecting with default mainnet params
+		mainnetAddress := net.JoinHostPort("localhost", netparams.MainNetParams.GRPCServerPort)
+		walletMiddleware, err = New(ctx, mainnetAddress, rpcCert, noTLS)
+		if err == nil {
+			saveRPCAddress(mainnetAddress)
+			return walletMiddleware, nil
+		}
+
+		fmt.Println("Could not detect a valid rpc address to connect to dcrwallet")
+		promt := "Do you want to set the address now?"
```
itswisdomagain: It's important to check after each connection attempt if `ctx.Err() != nil`, because a user could hit ctrl+c to terminate the app while the detection process is ongoing.

I noticed a scenario where hitting ctrl+c halted the program but I still got this prompt, which shouldn't be.

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
itswisdomagain: Also, do not make input compulsory, allow user to enter empty string and then quit the app.
Currently, if I don't set rpc address, I'm asked if to detect or set address. What if I don't want to do either? I have to hit ctrl+c. If we allow empty input, user could say enter address and enter nothing to quit.

itswisdomagain (changes requested):

app/walletmediums/dcrwalletrpc/address.go
ademuanthony: I added ```TBOpts ticketBuyerOptions 'group:"Ticket Buyer Options" namespace:"ticketbuyer"'``` because of the `group`, the parser is throwing if it's not present.

itswisdomagain: I tried it without and it didn't throw. Will try again

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
 	conn *grpc.ClientConn
 }
 
+const dcrwalletExecutableName = "dcrwallet"
```
itswisdomagain: this is no longer used, can be removed

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
+	if err == nil {
+		saveRPCAddress(mainnetAddress)
+		return walletMiddleware, nil
+	}
+	return nil, errSkip
+}
+
+func createConnection(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
 	// check if user has provided enough information to attempt connecting to dcrwallet
 	if rpcAddress == "" {
```
itswisdomagain: This should never happen, if an rpc connection attempt is being made, then the user has to have set the rpc address. This if statement can go

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
 	rpcConnectionTimeout = 5 * time.Second
+	errSkip              = errors.New("skip")
 )
 
 // New establishes gRPC connection to a running dcrwallet daemon at the specified address,
 // create a WalletServiceClient using the established connection and
 // returns an instance of `dcrwalletrpc.Client`
 func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
+	client, err1 := createConnection(ctx, rpcAddress, rpcCert, noTLS)
+	if err1 == nil {
```
itswisdomagain: Rename `err1` to `originalConnectionError`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if wrpcCert != "" {
+		rpcCert = wrpcCert
+	}
+	noTLS = wnoTLS
+
+	if client, err = useParsedConfigAddresses(ctx, rpcConfAddresses, rpcCert, noTLS); err != errSkip {
```
itswisdomagain: The only error returned by the `useParsedConfigAddresses` function is `errSkip`, if statement not necessary

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
+		return client, err
+	}
+
+	if client, err = connectToDefaultAddresses(ctx, rpcCert, noTLS); err != errSkip {
```
itswisdomagain: The only error returned by the `connectToDefaultAddresses ` function is `errSkip`, if statement not necessary
Also, the `errSkip` variable declaration can also be removed and it's usage replaced with a generic error

Since `err` is always `errSkip`, the only thing we should be checking is if `client != nil`. As a matter of fact, both `useParsedConfigAddresses` and `connectToDefaultAddresses` can be modified to only return `*WalletRPCClient`, which would be nil if the connection attempt failed

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
 	rpcConnectionDone    = make(chan *rpcConnectionResult)
 	rpcConnectionTimeout = 5 * time.Second
+	errSkip              = errors.New("skip")
 )
 
 // New establishes gRPC connection to a running dcrwallet daemon at the specified address,
 // create a WalletServiceClient using the established connection and
 // returns an instance of `dcrwalletrpc.Client`
 func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
+	client, err1 := createConnection(ctx, rpcAddress, rpcCert, noTLS)
```
itswisdomagain: `walletRPCClient` instead of `client`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
+
 var (
 	rpcConnectionDone    = make(chan *rpcConnectionResult)
 	rpcConnectionTimeout = 5 * time.Second
+	errSkip              = errors.New("skip")
 )
 
 // New establishes gRPC connection to a running dcrwallet daemon at the specified address,
 // create a WalletServiceClient using the established connection and
 // returns an instance of `dcrwalletrpc.Client`
```
itswisdomagain, line 42: Will need to update this comment to explain how this function behaves if the specified address does not connect

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
 func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
+	client, err1 := createConnection(ctx, rpcAddress, rpcCert, noTLS)
+	if err1 == nil {
+		return client, err1
+	}
+	rpcConfAddresses, wnoTLS, wrpcCert, err := walletAddressFromDcrdwalletConfig()
+	if err != nil {
+		return nil, err
```
itswisdomagain: Since the error being returned here is not a direct error from a connection attempt, would be better to wrap in an explanatory message:
`fmt.Errorf("Cannot connect to %s. Trying to check dcrwallet config for different address failed. %s", rpcAddress, err.Error())`

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -28,15 +31,74 @@ type rpcConnectionResult struct {
...
 
 // New establishes gRPC connection to a running dcrwallet daemon at the specified address,
 // create a WalletServiceClient using the established connection and
 // returns an instance of `dcrwalletrpc.Client`
 func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRPCClient, error) {
+	client, err1 := createConnection(ctx, rpcAddress, rpcCert, noTLS)
+	if err1 == nil {
+		return client, err1
+	}
+	rpcConfAddresses, wnoTLS, wrpcCert, err := walletAddressFromDcrdwalletConfig()
```
itswisdomagain: `dcrwalletConfAddresses, dcrwalletConfNoTLS, dcrwalletConfCert` are more descriptive variable names. Computer would always understand the variables, however short or long, so we should go with whatever is more readable for the sake of humans, even if it is long (but not too long).

app/walletmediums/dcrwalletrpc/address.go
```diff
@@ -0,0 +1,46 @@
...
+	Value         string
+	explicitlySet bool
+}
+
+type walletConfig struct {
+	GRPCListeners    []string        `long:"grpclisten" description:"Listen for gRPC connections on this interface/port"`
+	DisableServerTLS bool            `long:"noservertls" description:"Disable TLS for the RPC servers -- NOTE: This is only allowed if the RPC server is bound to localhost"`
+	RPCCert          *ExplicitString `long:"rpccert" description:"File containing the certificate file"`
+
+	TBOpts ticketBuyerOptions `group:"Ticket Buyer Options" namespace:"ticketbuyer"`
```
itswisdomagain: Can just use `struct{}` here instead of creating a `ticketBuyerOptions` and referencing it here (since the type is never really used)

itswisdomagain: Also need to go fmt

app/walletmediums/dcrwalletrpc/address.go
```diff
@@ -0,0 +1,46 @@
...
+
+const (
+	defaultWalletConfigFilename = "dcrwallet.conf"
+)
+
+var (
+	walletConfigFilePath = filepath.Join(config.DefaultDcrwalletAppDataDir, defaultWalletConfigFilename)
+)
+
+func walletAddressFromDcrdwalletConfig() (addresses []string, notls bool, certpath string, err error) {
```
itswisdomagain: - `dcrwallet`, not `dcrdwallet` 😉 
- this function returns more than just the addresses, so maybe `connectionParamsFromDcrwalletConfig`?

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -74,6 +136,13 @@ func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRP
 	}
 }
 
+func saveRPCAddress(address string) {
+	err := config.UpdateConfigFile("walletrpcserver", address, true)
+	if err != nil {
+		fmt.Printf("unable to save rpc address: %s", err.Error())
```
itswisdomagain: Printing this error would have been necessary if the user was prompted for an address and saving it failed. Since we're autodetecting, ignore this error. That would also mean that this function can be removed in favor of simply using `config.UpdateConfigFile("walletrpcserver", address, true)` in the call sites

app/walletmediums/dcrwalletrpc/dcrwalletrpc.go
```diff
@@ -74,6 +136,13 @@ func New(ctx context.Context, rpcAddress, rpcCert string, noTLS bool) (*WalletRP
...
+	}
+}
+
 func getNetParam(walletService walletrpc.WalletServiceClient) (param *chaincfg.Params, err error) {
```
itswisdomagain: This function should go below `connectToRPC` to follow the order in which they're called for easier code scanning

cli/termio/terminalprompt/terminalprompt.go
```diff
@@ -15,6 +16,14 @@ var EmptyValidator = func(v string) error {
 	return nil
 }
 
+// InputRequiredValidator throws if the input value is empty
+var InputRequiredValidator = func(v string) error {
```
itswisdomagain: Don't think this is used anymore, can remove

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #160, opened 2019-01-11
implemented send and custom send spend unconfirmed feature for web and cli

### LINKED ISSUE #138: send feature improvement (web+cli) - spend confirmed or unconfirmed
To send funds, godcr fetches unspent outputs from an account's transaction history. Unspent outputs 
are only considered for use in a new transaction if they belong to a transaction that has a specified number of confirmations.

The actual selection of unspent outputs may not be done in godcr core code (except for custom send). Still, regardless of how and where the selection is done, a value must be specified for the minimum number of confirmations to use.

A default value of 2 is currently used - where 2 represents fully confirmed transactions on the decred blockchain. However, a user should be able to spend from previous transactions that are not yet unconfirmed (< 2 confirmations).

To achieve this, the web interface should have a `Spend unconfirmed` checkbox to allow users override the default of 2 minimum confirmations and send funds using 0 confirmations.

The cli `send` and `sendcustom` commands should register a `--spendunconfirmed` flag to allow users chose to use 0 confirmations for send transactions.

Sirmorrison:
fix #138 

implemented send and custom send spend unconfirmed feature for web and cli

### DIFF
--- a/cli/commands/send.go
+++ b/cli/commands/send.go
@@ -14,38 +14,33 @@ import (
 // SendCommand lets the user send DCR.
 type SendCommand struct {
 	commanderStub
-	Spendunconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
+	SpendUnconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
 }
 
 // Run runs the `send` command.
 func (s SendCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
-	var requireConfirmation int32 
-	if s.Spendunconfirmed{
-		requireConfirmation = 0
-	}else{
-		requireConfirmation = walletcore.DefaultRequiredConfirmations
-	}
-	return send(wallet, requireConfirmation, false)
+	return send(wallet, s.SpendUnconfirmed, false)
 }
 
 // SendCustomCommand sends DCR using coin control.
 type SendCustomCommand struct {
 	commanderStub
-	Spendunconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
+	SpendUnconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
 }
 
 // Run runs the `send-custom` command.
 func (s SendCustomCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
-	var requireConfirmation int32 
-	if s.Spendunconfirmed{
-		requireConfirmation = 0
-	}else{
-		requireConfirmation = walletcore.DefaultRequiredConfirmations
-	}
-	return send(wallet, requireConfirmation, true)
+	return send(wallet, s.SpendUnconfirmed, true)
 }
 
-func send(wallet walletcore.Wallet, requireConfirmation int32, custom bool) error {
+func send(wallet walletcore.Wallet, spendUnconfirmed bool, custom bool) error {
+	var requireConfirmations int32 
+	if spendUnconfirmed{
+		requireConfirmations = 0
+	}else{
+		requireConfirmations = walletcore.DefaultRequiredConfirmations
+	}
+
 	sourceAccount, err := selectAccount(wallet)
 	if err != nil {
 		return err
@@ -53,7 +48,7 @@ func send(wallet walletcore.Wallet, requireConfirmation int32, custom bool) erro
 
 	// check if account has positive non-zero balance before proceeding
 	// if balance is zero, there'd be no unspent outputs to use
-	accountBalance, err := wallet.AccountBalance(sourceAccount, requireConfirmation)
+	accountBalance, err := wallet.AccountBalance(sourceAccount, requireConfirmations)
 	if err != nil {
 		return err
 	}
--- a/cli/commands/send.go
+++ b/cli/commands/send.go
@@ -50,7 +50,6 @@ func send(wallet walletcore.Wallet, requireConfirmation int32, custom bool) erro
 	if err != nil {
 		return err
 	}
-			fmt.Println(requireConfirmation)
 
 	// check if account has positive non-zero balance before proceeding
 	// if balance is zero, there'd be no unspent outputs to use
--- a/cli/commands/send.go
+++ b/cli/commands/send.go
@@ -34,11 +34,9 @@ func (s SendCustomCommand) Run(ctx context.Context, wallet walletcore.Wallet) er
 }
 
 func send(wallet walletcore.Wallet, spendUnconfirmed bool, custom bool) error {
-	var requiredConfirmations int32 
-	if spendUnconfirmed{
+	var requiredConfirmations int32 = walletcore.DefaultRequiredConfirmations
+	if spendUnconfirmed {
 		requiredConfirmations = 0
-	}else{
-		requiredConfirmations = walletcore.DefaultRequiredConfirmations
 	}
 
 	sourceAccount, err := selectAccount(wallet)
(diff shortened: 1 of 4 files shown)


cli/commands/send.go
```diff
@@ -14,35 +14,50 @@ import (
 // SendCommand lets the user send DCR.
 type SendCommand struct {
 	commanderStub
+	Spendunconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
```
itswisdomagain: CamelCasing for variable/field names: `SpendUnconfirmed`

cli/commands/send.go
```diff
@@ -14,35 +14,50 @@ import (
...
 }
 
 // Run runs the `send` command.
 func (s SendCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
-	return send(wallet, false)
+	var requireConfirmation int32
```
itswisdomagain: `requiredConfirmations`

cli/commands/send.go
```diff
@@ -14,35 +14,50 @@ import (
...
 type SendCustomCommand struct {
 	commanderStub
+	Spendunconfirmed bool `short:"u" long:"spendunconfirmed" description:"Use unconfirmed outputs for send transactions."`
 }
 
 // Run runs the `send-custom` command.
 func (s SendCustomCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
-	return send(wallet, true)
+	var requireConfirmation int32 
+	if s.Spendunconfirmed{
```
itswisdomagain: this code is repeated, you should instead pass a `spendUnconfirmed bool` parameter to the `send` function and do this if/else statement there

cli/commands/send.go
```diff
@@ -14,35 +14,50 @@ import (
...
+		requireConfirmation = 0
+	}else{
+		requireConfirmation = walletcore.DefaultRequiredConfirmations
+	}
+	return send(wallet, requireConfirmation, true)
 }
 
-func send(wallet walletcore.Wallet, custom bool) error {
+func send(wallet walletcore.Wallet, requireConfirmation int32, custom bool) error {
```
itswisdomagain: modify this to `func send(wallet walletcore.Wallet, sendUnconfirmed, custom bool) error {`
see above comment

web/routes/handlers.go
```diff
@@ -82,6 +82,7 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 	selectedAccount := req.FormValue("sourceAccount")
 	destAddress := req.FormValue("destinationAddress")
 	passphrase := req.FormValue("walletPassphrase")
+	unconfirmed := req.FormValue("SpendUnconfirmed")
```
itswisdomagain: why not `spendUnconfirmed`?
what does `unconfirmed` mean?

web/routes/handlers.go
```diff
@@ -101,6 +102,13 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 		Address: destAddress,
 	}}
 
+	var RequiredConfirmations int32
+	if unconfirmed == "on"{
```
itswisdomagain: it might be a good idea to accept any value for `spendUnconfirmed`
we know for sure that the value of `spendUnconfirmed` would be an empty string if the checkbox was not checked, use that instead

`if spendUnconfirmed != "" {`

web/routes/handlers.go
```diff
@@ -101,6 +102,13 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 		Address: destAddress,
 	}}
 
+	var RequiredConfirmations int32
```
itswisdomagain: `requiredConfirmations`

web/views/send.html
```diff
@@ -57,6 +57,12 @@
                                         </div>
                                     </div>
                                 </div>
+                                <div>
+                                    <div class="form-group form-check">
+                                        <input type="checkbox" class="form-check-input" name="SpendUnconfirmed" id="SpendUnconfirmed">
```
itswisdomagain: `spendUnconfirmed`
follow the pattern used for ids and names in other input fields

itswisdomagain (changes requested):
Looks neat, a few corrections

cli/commands/send.go
Sirmorrison: tried the camel casing didnt work for 'Spendunconfirmed'. had to capitalize the 'Spend' before it was passed across the send function

itswisdomagain: The review said to use `SpendUnconfirmed`, are you saying that won't work?

cli/commands/send.go
```diff
@@ -14,35 +14,45 @@ import (
...
 
 // Run runs the `send-custom` command.
 func (s SendCustomCommand) Run(ctx context.Context, wallet walletcore.Wallet) error {
-	return send(wallet, true)
+	return send(wallet, s.SpendUnconfirmed, true)
 }
 
-func send(wallet walletcore.Wallet, custom bool) error {
+func send(wallet walletcore.Wallet, spendUnconfirmed bool, custom bool) error {
+	var requireConfirmations int32
```
itswisdomagain: This `if/else` block can be simplified. Here's what I'm thinking:
```
requiredConfirmations := walletcore.DefaultRequiredConfirmations
if spendUnconfirmed {
     requiredConfirmations = 0
}
```

Also, notice the space at the end of the if line, before the curly brace. That's the correct formatting.
Also run `go fmt ./...` to ensure your code is properly formatted before pushing again

itswisdomagain: Also notice I used `requiredConfirmations` as against `requireConfirmations`

cli/commands/send.go
```diff
@@ -14,35 +14,45 @@ import (
...
+
 	sourceAccount, err := selectAccount(wallet)
 	if err != nil {
 		return err
 	}
 
 	// check if account has positive non-zero balance before proceeding
 	// if balance is zero, there'd be no unspent outputs to use
-	accountBalance, err := wallet.AccountBalance(sourceAccount, walletcore.DefaultRequiredConfirmations)
+	accountBalance, err := wallet.AccountBalance(sourceAccount, requireConfirmations)
```
itswisdomagain: The `completeCustomSend` and `completeNormalSend` functions in this file also need to be updated to use the value of `requiredConfirmations` instead of `walletcore.DefaultRequiredConfirmations` which is currently used in both functions

web/routes/handlers.go
```diff
@@ -101,6 +102,13 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 		Address: destAddress,
 	}}
 
+	var requireConfirmations int32
```
itswisdomagain: This `if/else` block can be simplified. Here's what I'm thinking:
```
requiredConfirmations := walletcore.DefaultRequiredConfirmations
if spendUnconfirmed != "" {
     requiredConfirmations = 0
}
```

itswisdomagain (changes requested):

cli/commands/send.go
Sirmorrison: having type issues so had to declare it explicitly

web/routes/handlers.go
```diff
@@ -101,6 +102,12 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 		Address: destAddress,
 	}}
 
+	var requiredConfirmations int32 = walletcore.DefaultRequiredConfirmations
+	if spendUnconfirmed != "" {
+		requiredConfirmations = 0
+	}
+
+	fmt.Println(requiredConfirmations)
```
itswisdomagain: Testing, I guess. Delete please

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #161, opened 2019-01-11
detect wallets for use with dcrlibwallet

### LINKED ISSUE #89: dcrlibwallet - autodetect wallets and network type
Currently, there' a `t/--testnet` flag for instructing dcrcli to use testnet and not mainnet. This is primarily used by the godcr program for connecting to a wallet. [dcrlibwallet](https://github.com/raedahgroup/dcrlibwallet/tree/dcrlibwallet-wip) uses that information to decide which wallet database to load, testnet wallet database is stored in godcr's appdata dir in a "testnet" subdirectory (for linux, that is `~/.godcr/testnet`) while mainnet is stored in "mainnet" subdirectory (`~/.godcr/mainnet` for linux).

The flag should be removed, and godcr should autodetect wallets on a user's PC and the network type for each wallet, save this information in a `wallets.conf` file and use the information saved when establishing a connection to a wallet.

A list of potential directories to find wallet databases should be maintained on the program. Currently,
- dcrwallet stores wallet database files in dcrwallet's appdata dir, under subdirectories "testnet3" for testnet and "mainnet" for mainnet network types.
- dcrlibwallet stores wallet database files in godcr's appdata dir, under subdirectories "testnet" for testnet and "mainnet" for mainnet network types.
- decredition uses `$HOME/.config/decrediton/wallets`

### Desired behaviour on cli mode
- Add `godcr detect` command to perform a scan for wallet databases that godcr can connect to. A `wallets.conf` file should be created in the godcr appdata dir to store information about discovered wallets and their network type. Regardless of the wallet connection mode set by the user, `godcr discover` should detect and save information for all "detectable" wallets.

- The command should list all wallets detected and prompt user to select a default wallet to connect to. A `--wallet` config option should be added to `godcr.conf` to store the "id" of the default wallet. If only one wallet is detected, that should be made default without prior confirmation from user.

- Every run of `godcr` should check if the `wallets.conf` file exists in the appdata dir. If no wallets are configured, it should ask if user wants to detect existing wallets. If godcr wallet config already exists, then no detect is needed (unless explicitly called with the `detect` command).

- If godcr wallets.conf exists but contains no wallet information, user should be asked if to create a new wallet. The wallet creation process should ask the user what network type he wants to create a wallet for. After wallet is created, information about the newly created wallet should be added to the wallets.conf file.

- If godcr wallets.conf exists and has data for just one wallet, godcr should automatically load that wallet. If multiple wallets were detected, the default wallet as specified by the `--wallet` option should be loaded. The user should be able to override the default wallet option saved in godcr.conf by passing `--wallet="id"` in command-line options.

- Similar functionalities should be added to the web and desktop interfaces (preferably in separate PRs). The `testnet` flag and wallet connection current behaviour can be retained for those interfaces until this new feature is implemented for all interfaces. After which, the `testnet` config flag should be discontinued for `dcrlibwallet` connections.

itswisdomagain:
Fixes #89

This PR finally retires the `--testnet` flag and relies on the godcr.conf file to determine network type for connecting to wallets via dcrlibwallet.

If wallets information does not exist in godcr.conf, user is prompted to allow godcr scan for existing wallets. During this scan, network types for existing wallets is detected and this information saved to godcr.conf. This scan can also be triggered at any time by running `godcr detect`

If more than 1 wallet is detected, the user is asked to chose a default wallet which will be used by dcrlibwallet subsequently.

The create wallet command is also modified to ask user for the network type for the new wallet.

If during a scan, no wallet is detected, the user is asked to create one and the created wallet added to godcr.conf


cli/walletloader/helpers.go
```diff
@@ -3,13 +3,41 @@ package walletloader
 import (
 	"context"
 	"fmt"
+	"github.com/raedahgroup/godcr/app/config"
```
ademuanthony: Line 6 and 7 can be moved down

ademuanthony:
I am unable to select a wallet
no-wallet-selection

ademuanthony:
The app stays without any visible activity after selecting the wallet.
Is it running a time-consuming background task? Can we show some notification to the user?
wallet-hanging-after-selection

ademuanthony (approved):

PR #164, opened 2019-01-12
http send page cleanup

### LINKED ISSUE #57: cleanup http send page

yaowork:
this performs cleanup that fixes #57

### DIFF
--- a/web/public/css/style.css
+++ b/web/public/css/style.css
@@ -100,7 +100,10 @@ body {
     padding-bottom: 0 !important;
 }
 
-select#source-account[:disabled] {
+select#source-account[disabled] {
     border: none !important;
     background-color: transparent !important;
+    padding: 0;
+    -webkit-appearance: none;
+    -moz-appearance: none;
 }
--- a/web/public/css/style.css
+++ b/web/public/css/style.css
@@ -72,52 +72,6 @@ body {
     -webkit-border-radius: 0;
 }
 
-.stepper .nav-tabs {
-    border-color: transparent;
-    background-color: #fff;
-}
-
-.stepper .nav-tabs .nav-item {
-    transition: all 0.5s;
-    padding: 0 !important;
-}
-
-.stepper .nav-link {
-    color: inherit;
-    display: block;
-    text-align: center;
-    padding: 20px 0 !important;
-    border-color: transparent !important;
-    border-radius: 0;
-    -moz-border-radius: 0;
-    -webkit-border-radius: 0;
-}
-
-.stepper .nav-link{
-    pointer-events: none;
-    cursor: default;
-}
-
-.stepper .nav-tabs .stepper-active {
-    background-color: #09c !important;
-    color: #fff;
-}
-
-.stepper .nav-link span {
-    font-size: 14px !important;
-}
-
-.stepper .progress {
-    height: 3px;
-}
-
-.stepper .tab-content {
-    padding: 20px 30px;
-    background-color: #fff;
-    border: 1px solid #d0d0d0;
-    border-top-color: transparent;
-}
-
 .table td, .table th{
     padding: .3rem
 }
@@ -130,4 +84,23 @@ body {
 
 .breadcrumb-item{
     font-size: 15px;
+}
+
+.collapsible .card-header {
+    padding: 0 !important;
+}
+
+.card-header .btn-link {
+    text-decoration: none;
+    display: block;
+    cursor: pointer;
+}
+
+.no-btm-pad {
+    padding-bottom: 0 !important;
+}
+
+select#source-account[:disabled] {
+    border: none !important;
+    background-color: transparent !important;
 }
--- a/web/public/js/send.js
+++ b/web/public/js/send.js
@@ -16,13 +16,13 @@ function validateSendForm() {
     var errors = [];
     var isClean = true;
 
-    if ($("#source-account").find(":selected").text() == "") {
+    if ($("#source-account").find(":selected").text() === "") {
         errors.push("The source account is required");
     }
 
     isClean = validateAmountField();
 
-    if ($("#destination-address").val() == "") {
+    if ($("#destination-address").val() === "") {
         errors.push("The destination address is required");
     }
 
@@ -80,11 +80,11 @@ function openCustomizePanel() {
         // populate outputs 
         var utxoHtml = txs.map(tx => {
             var receiveDateTime = new Date(tx.receive_time * 1000);
-            var amount = tx.amount / 100000000;
+            var dcrAmount = tx.amount / 100000000;
             return  "<tr>" + 
-                        "<td width='5%'><input type='checkbox' class='custom-input' name='utxo' value="+ tx.key+" data-amount='" + amount + "' /></td>" +
+                        "<td width='5%'><input type='checkbox' class='custom-input' name='utxo' value="+ tx.key+" data-amount='" + dcrAmount + "' /></td>" +
                         "<td width='50%'>" + tx.key + "</td>" + 
-                        "<td width='20%'>" + amount + "</td>" + 
+                        "<td width='20%'>" + dcrAmount + " DCR</td>" + 
                         "<td width='25%'>" + receiveDateTime.toString().split(' ').slice(0,5).join(' '); + "</td>" +
                     "</tr>"
         });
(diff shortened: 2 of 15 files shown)


web/views/send.html
```diff
@@ -10,84 +10,77 @@
...
                                         <div class="form-group">
-                                            <label for="sourceAccount">Source Account</label>
-                                            <select class="form-control" id="sourceAccount" name="sourceAccount">
-                                                {{ range $account := .accounts }}
-                                                <option value="{{ $account.Number }}">
-                                                    {{ $account.Name }} - {{ $account.Balance.Total }}</option>
-                                                {{ end }}
-                                            </select>
+                                            <label for="source-account">Source Account</label>
+                                            {{ if eq (len .accounts) 1 }}
```
itswisdomagain: This if/else block can be simplified to use just one select:
```
<select class="form-control" id="source-account" name="source-account" {{ if eq (len .accounts) 1 }}disabled{{ end }}>
    <option value="{{ $account.Number }}">
    {{ $account.Name }} - {{ simpleBalance $account.Balance false }}
    </option>
{{ end }}
</select>
```

web/views/send.html
```diff
@@ -10,84 +10,77 @@
...
                                             <input type="number" class="form-control" id="amount" name="amount" />
                                         </div> 
-                                    </div>
-                                </div>
-                                <div class="row">
-                                    <div class="col-md-6 col-sm-6">
                                         <div class="form-group">
                                             <label for="destinationAddress">Destination Address</label>
-                                            <input type="text" class="form-control" id="destinationAddress" name="destinationAddress" />
+                                            <input type="text" class="form-control" id="destination-address" name="destination-address" />
```
itswisdomagain, line 42: The name attribute of most elements on this form was changed. The `submitSendTxForm` handler has not been updated to reflect this change, this page doesn't work currently. Did you try sending funds from this page?

web/views/send.html
```diff
@@ -10,84 +10,77 @@
...
+                                        </div>
+                                        <div>
+                                            <input type="checkbox" name="use-custom" id="use-custom" value="1" />
```
itswisdomagain: The `submitSendTxForm` handler still uses `len(utxos) > 0` to check if to create a custom send tx or a normal one. It's possible the user selects some utxos but later unchecks this box, meaning he wants to do a normal send tx. `req.FormValue("use-custom") != ""` should be used to determine if to build a custom tx or normal tx.

itswisdomagain (changes requested):
This send page doesn't work. Can't test much until that is fixed. Here are some required changes meanwhile.

web/public/css/style.css
```diff
@@ -130,4 +84,23 @@ body {
...
+    text-decoration: none;
+    display: block;
+    cursor: pointer;
+}
+
+.no-btm-pad {
+    padding-bottom: 0 !important;
+}
+
+select#source-account[:disabled] {
```
itswisdomagain: This selector doesn't look correct. bootstrap does this: `.form-control:disabled`
A correct way to write what you have is  `select#source-account[disabled]`

I was able to produce a more widget-like appearance with the following css code:
```
select#source-account[disabled] {
    border: none !important;
    background-color: transparent !important;
    padding: 0;
    -webkit-appearance: none;
    -moz-appearance: none;
}
```

itswisdomagain: screenshot 2019-01-14 at 11 34 15 am

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
-    var sourceAccountEl = $("#sourceAccount");
-    var amountEl = $("#amount");
-    var destinationAddressEl = $("#destinationAddress");
+    $(".errors").empty();
+    var errors = [];
     var isClean = true;
 
-    if (sourceAccountEl.find(":selected").text() == "") {
-        errors["#sourceAccount"] = "The source account is required";
+    if ($("#source-account").find(":selected").text() == "") {
```
itswisdomagain: `===` instead of `==`
line 25 also

web/routes/handlers.go
```diff
@@ -108,7 +109,7 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 	}
 
 	var txHash string
-	if len(utxos) > 0 {
+	if useCustom == "1" {
```
itswisdomagain: I prefer `if useCustom != ""` since the yes value may defer but if the box is not checked, useCustom will always be empty, so checking if it's not empty is a better guarantee that the box was indeed checked

itswisdomagain (changes requested):

itswisdomagain:
Some observations:
screenshot 2019-01-14 at 12 00 00 pm

- The height of each row in the utxo table should be reduced
- The progress bar isn't really "helpful" as is, a user looking at this will not even know that that is a progress bar and even when he finally realizes that he's looking at a progress bar, he may not understand what information the progress bar is trying to depict. We can discuss to come up with something more intuitive to satisfy this requirement
- The "Spend unconfirmed" checkbox should certainly not be between the "Select custom inputs" checkbox and the table. The UI should clearly show that there's a relationship between the "Select custom inputs" and the table displayed

itswisdomagain:
More observations:
screenshot 2019-01-14 at 12 05 52 pm

My spendable balance as seen in the UI above is 0. Ideally, I should not be able to send funds, hence the utxo table doesn't show any inputs to select, because no inputs is confirmed

However, when I check "spend unconfirmed", I expect to see unconfirmed inputs in the list and the transaction should go through.

I'm thinking the "Spend unconfirmed" checkbox should be moved to just below the "Source Account" form group because it is quite related to the selected account. The value of that checkbox (whether checked or not) should be used in listing utxos for selection and also when performing the send operation itself.

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
+function openCustomizePanel() {
+    resetCustomizePanel();
+    $("#custom-tx-row").slideDown();
+   
+    var account_number = $("#source-account").find(":selected").val();
+    var callback = function(txs) {
+        // populate outputs 
+        var utxoHtml = txs.map(tx => {
+            var receiveDateTime = new Date(tx.receive_time * 1000);
+            var amount = tx.amount / 100000000;
```
itswisdomagain: rename `amount` to `dcrAmount`

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
+            return  "<tr>" + 
+                        "<td width='5%'><input type='checkbox' class='custom-input' name='utxo' value="+ tx.key+" data-amount='" + amount + "' /></td>" +
+                        "<td width='50%'>" + tx.key + "</td>" + 
+                        "<td width='20%'>" + amount + "</td>" +
```
itswisdomagain: Include the unit for `dcrAmount` here
`"<td width='20%'>" + dcrAmount + " DCR</td>"`

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
+        $(".custom-input").on("click", function(){
+           calculateSelectedInputPercentage();
+        });
+
+        $("#amount").on("keyup", function(){
+            validateAmountField();
+            calculateSelectedInputPercentage();
+        });
+    }
+    getUnspentOutputs(account_number, callback);
```
itswisdomagain: bring the `getUnspentOutputs` function just below this function for easier readability

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
+}
+
+
 function submitSendForm() {
     var form = $("#send-form");
     var submit_btn = $("#send-form #submit-btn");
     submit_btn.attr("disabled", "disabled").html("Sending...");
 
     // get total of selected inputs and add to form post data
     totalSelectedInputAmount = 0;
```
itswisdomagain: delete lines 113 to 119, there's a function to get the total selected input amount we can use

web/public/js/send.js
```diff
@@ -1,89 +1,131 @@
...
-    $('.utxo:checkbox:checked').each(function () {
+    $('.custom-input').each(function () {
         if (this.checked) {
-            totalSelectedInputAmount += $(this).attr("data-amount");
+            totalSelectedInputAmount += parseInt($(this).attr("data-amount"));
         }
     });
 
     var postData = form.serialize();
     postData += "&totalSelectedInputAmount=" + totalSelectedInputAmount;
```
itswisdomagain: `postData += "&totalSelectedInputAmountDcr=" + getSelectedInputsSum();`
- use the function instead of the variable
- rename totalSelectedInputAmount to totalSelectedInputAmountDcr since the sum is in dcr (float64) but the required value in the `submitSendTxForm` handler is atoms (int64), to avoid ambiguity

web/routes/handlers.go
```diff
@@ -79,10 +79,11 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 	utxos := req.Form["utxo"]
 	totalSelectedInputAmount := req.FormValue("totalSelectedInputAmount")
```
itswisdomagain: `totalSelectedInputAmountDcr := req.FormValue("totalSelectedInputAmountDcr")`

web/routes/handlers.go
```diff
@@ -108,7 +109,7 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
 	}
 
 	var txHash string
-	if len(utxos) > 0 {
+	if useCustom == "1" {
 		totalInputAmount, err := strconv.ParseInt(totalSelectedInputAmount, 10, 64)
```
itswisdomagain: This section should read the value as float64 and convert to dcrutil.Amount (int64 alias)
```
totalInputAmountDcr, err := strconv.ParseFloat(totalSelectedInputAmountDcr, 64)
if err != nil {
	data["error"] = err.Error()
	return
}
totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
if err != nil {
	data["error"] = err.Error()
	return
}
```

itswisdomagain: This would prevent the insufficient funds error when using custom utxo selection

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #166, opened 2019-01-14
Basic framework for Qt interface mode 

itswisdomagain:
### Important
This PR introduces some serious changes to how godcr is built. Care should be taken to follow every listed step correctly to have a successful build.

### Summary of build steps
```
// the following only need to be done once
go get -u -v github.com/therecipe/qt/cmd/...
qtsetup

// the following should be done whenever you want to build godcr
export GO111MODULE=on
go mod download
go mod vendor
export GO111MODULE=off
go build
```

### Breakdown of build steps
First, install and setup the go qt binding to be able to build this code. Linux users might need to install some other dependencies, see https://github.com/therecipe/qt/wiki/Installation-on-Linux. This should be a one time setup.
```
go get -u -v github.com/therecipe/qt/cmd/...
qtsetup
```

After installing qt binding, follow the steps below to build `godcr`
- Enable go modules. On Linux and Mac, simply run `export GO111MODULE=on`. Windows users will have to set the `GO111MODULE=on` environment variable and restart command prompt afterwards.
- While go modules is enabled, download the dependencies listed in `go.mod` into a `vendor` subdirectory in the `godcr` project folder. Run `go mod download` followed by `go mod vendor`. Both commands will only work if go modules is enabled.
- Confirm that the previous step created a `vendor` subdirectory in your project folder
- Disable go modules. Run `export GO111MODULE=off` on your Mac or Linux terminal. If using Windows, set `GO111MODULE=off` in environment variable and restart command prompt to load the new environment variable value.
- Build as normal: `go build`
- After building, run `godcr --mode=qt`

### To note
Qt does not yet properly support go modules, which is why go modules is disabled before running `go build`. This also means that the source code being built must be placed inside the `$GOPATH/src/github.com/raedahgroup/godcr` for the build to be successful


itswisdomagain:
result of `godcr --mode=http`

screenshot 2019-01-16 at 3 22 45 pm

ademuanthony (approved):

engprodigy:
> result of `godcr --mode=http`
> 
> screenshot 2019-01-16 at 3 22 45 pm

@itswisdomagain the cli command is `godcr --mode=qt` and not `godcr --mode=http` to launch the qt app. I guess it was an oversight.

Issue #162, opened 2019-01-12
add terminal ui interface mode


itswisdomagain:
Add `mode=terminal` for running godcr as a full UI app running on terminal using this library: [https://github.com/rivo/tview](https://github.com/rivo/tview)

The new interface would be quite similar to the http interface (`mode=http`), providing a set of UI widgets/elements on the terminal for users to perform wallet operations.

As with the http interface, the behaviour of the terminal gui should start out with the following features:
- Open wallet
- Sync on load (show a sync progress view, no wallet-related function should be accessible till sync completes)
- Some sort of "menu" to provide click access to different godcr functionalities
- Close / Shutdown (could be through a button on the UI, or when a user hits `ctrl + C` or both)

A PR should be created once this basic skeleton is achieved. Then work should continue to implement each godcr feature as separate "pages":
- Balance page (shows account balances for all accounts, similar to `Balance` page in http mode and `godcr balance` command in cli mode)
- Receive page (to allow user select an account to generate an address for receiving funds)
- All other cli commands and http pages

Implementing godcr features in this `terminal` mode should be done progressively, perhaps creating PRs for each feature.

PR #167, opened 2019-01-15
Skeletal framework tview

raphaelint:
Initial fix for #162 

Created a terminal package with a terminal.go file that initialize the rivo/tview to create the skeletal terminal ui

I created a new mode **--mode=terminal**

tview


terminal/terminal.go
```diff
@@ -0,0 +1,24 @@
...
+
+func OpenTerminal() {
+
+	app := tview.NewApplication()
+
+	list := tview.NewList().
+		AddItem("Open wallet", "", 'a', nil).
+		AddItem("Sync on load", "", 'b', nil).
+		AddItem("Account Balance", "", 'c', nil).
+		AddItem("Close / Shutdown", "Press to exit", 'q', func() {
```
itswisdomagain: This list should only have
- (b) Balance (not Account Balance)
- (r) Receive
- (s) Send
- (h) History
- (q) Exit (not Close / Shutdown, also the explanatory text is not necessary)

main.go
```diff
@@ -75,6 +76,9 @@ func main() {
 		enterHttpMode(ctx, walletMiddleware, appConfig)
 	} else if appConfig.InterfaceMode == "nuklear" {
 		enterDesktopMode(ctx, walletMiddleware)
+	} else if appConfig.InterfaceMode == "terminal" {
+		//enterDesktopMode(ctx, walletMiddleware)
```
itswisdomagain: comment is unnecessary

terminal/terminal.go
```diff
@@ -0,0 +1,24 @@
+package terminal
+
+import (
+	"github.com/rivo/tview"
+)
+
+func OpenTerminal() {
```
itswisdomagain: We're not actually opening a terminal, are we? This can be renamed to `StartTerminalApp`

terminal/terminal.go
```diff
@@ -0,0 +1,24 @@
...
+
```
itswisdomagain: This space can go

terminal/terminal.go
```diff
@@ -0,0 +1,24 @@
...
+		AddItem("Sync on load", "", 'b', nil).
+		AddItem("Account Balance", "", 'c', nil).
+		AddItem("Close / Shutdown", "Press to exit", 'q', func() {
+			app.Stop()
+		})
+
+	list.SetBorder(true).SetTitle("GoDCR Terminal")
+
+	if err := app.SetRoot(list, true).SetFocus(list).Run(); err != nil {
+		panic(err)
```
itswisdomagain: Don't panic from an interface, interfaces should return any execution errors back to main for proper handling
- modify this function's signature to return `error`
- in main.go, modify the `enterTerminalMode` function to receive any errors returned from this function: `opError = terminal.OpenTerminal()`
- instead of this if statement, do the following, include the comment too, helps whoever reads/maintains your code to understand what's happening here:
```
// `Run` blocks until app.Stop() is called before returning
return app.SetRoot(list, true).SetFocus(list).Run()
```

itswisdomagain (changes requested):

terminal/terminal.go
raphaelint: Ohk... Done

main.go
```diff
@@ -156,6 +159,15 @@ func enterDesktopMode(ctx context.Context, walletMiddleware app.WalletMiddleware
 	beginShutdown <- true
 }
 
+func enterTerminalMode(ctx context.Context, walletMiddleware app.WalletMiddleware) {
+	fmt.Println("Launching Terminal...")
+	opError := terminal.StartTerminalApp(ctx, walletMiddleware)
+	//Shutdown if error occurred
+	if opError != nil {
```
itswisdomagain: this if statement is unnecessary, because regardless of whether `opError` is nil or not, you trigger shutdown

main.go
```diff
@@ -156,6 +159,15 @@ func enterDesktopMode(ctx context.Context, walletMiddleware app.WalletMiddleware
 	beginShutdown <- true
 }
 
+func enterTerminalMode(ctx context.Context, walletMiddleware app.WalletMiddleware) {
+	fmt.Println("Launching Terminal...")
+	opError := terminal.StartTerminalApp(ctx, walletMiddleware)
```
itswisdomagain: there's already `opError` declared earlier in this file that is used in the `handleShutdown` function at the end of this file, simply assign to that variable instead of creating a new one here

terminal/terminal.go
```diff
@@ -0,0 +1,41 @@
...
+	"context"
+	"fmt"
+
+	"github.com/rivo/tview"
+	"github.com/raedahgroup/godcr/cli/walletloader"
+	"github.com/raedahgroup/godcr/app"
+)
+
+func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
+	app := tview.NewApplication()
```
itswisdomagain: `app` is the name of an import above, rename this to `tviewApp` instead

terminal/terminal.go
```diff
@@ -0,0 +1,41 @@
...
+
+	list := tview.NewList().
+		AddItem("Balance", "", 'b', nil).
+		AddItem("Receive", "", 'r', nil).
+		AddItem("Send", "", 's', nil).
+		AddItem("History", "", 'h', nil).
+		AddItem("Exit", "", 'q', func() {
+			app.Stop()
+		})
+	list.SetBorder(true).SetTitle("GoDCR Terminal")
```
itswisdomagain: Avoid hardcoding the app name anywhere in code, instead use 
`fmt.Sprintf("%s Terminal", strings.ToUpper(app.Name))`

raphaelint:
This is the Terminal view ui

new terminal

terminal/terminal.go
```diff
@@ -0,0 +1,41 @@
...
+		fmt.Println(err)
+	}
+
+	// `Run` blocks until app.Stop() is called before returning
+	return app.SetRoot(list, true).SetFocus(list).Run()
+}
+
+func syncBlockChain(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
+	walletExists, err := walletloader.OpenWallet(ctx, walletMiddleware)
+	if err != nil || !walletExists {
```
itswisdomagain: if wallet does not exist, err will not be nil. Checking if `err != nil` is enough

terminal/terminal.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil{
+		fmt.Println(err)
+	}
+
+	// `Run` blocks until app.Stop() is called before returning
+	return app.SetRoot(list, true).SetFocus(list).Run()
+}
+
+func syncBlockChain(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
+	walletExists, err := walletloader.OpenWallet(ctx, walletMiddleware)
```
itswisdomagain: This is applaudable, but ideally should only be used in the `cli` interface. We'll leave it for now and correct it later

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #173, opened 2019-01-17
Nuklear refactor

yaowork:
This refactor improves the performance of the desktop UI.


nuklear/desktop.go
```diff
@@ -8,45 +8,48 @@ import (
...
+	// initialize master window and set style
+	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, desktop.render)
+	window.SetStyle(helpers.GetStyle())
+	desktop.masterWindow = window
+
+	// register handlers
+	handlers := getHandlers()
+	desktop.handlers = make(map[string]Handler, len(handlers))
+	for _, handler := range handlers {
+		handler.handler.SetWalletMiddleware(walletMiddleware)
```
itswisdomagain: No need to maintain an instance of `walletMiddleware` in all handlers. Remove the `handler.SetWalletMiddleware ` function and modify the `handler.Render` function to accept `wallet walletcore.Wallet`. Whenever you call `handler.Render`, pass `desktop.wallet` as parameter to the function.

nuklear/desktop.go
```diff
@@ -57,87 +60,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+
 	// style content area
-	d.setPageStyle()
+	helpers.SetPageStyle(desktop.masterWindow)
 
-	contentArea = rect
-	w.LayoutSpacePushScaled(rect)
+	w.LayoutSpacePushScaled(contentRect)
+	handler, ok := desktop.handlers[desktop.currentPage]
+	if !ok { // ideally, this should only be false once in the lifetime of an instance
```
itswisdomagain: instead of this, it would be better to check if `desktop.currentPage == ""` which is the only reason why `ok == false`
If `desktop.currentPage == ""`, change page to home page

nuklear/desktop.go
```diff
@@ -57,87 +60,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+		desktop.pageChanged = false
 	}
+
+	handler.Render(w)
+}
+
+func (desktop *Desktop) changePage(page string) {
+	desktop.currentPage = page
+	desktop.pageChanged = true
+	desktop.masterWindow.Changed()
```
itswisdomagain, line 132: I have some concern about having to draw the navigation section everytime the page changes. Might be better to add the content window to desktop (`desktop.pageWindow`) and update that instead

nuklear/handler.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	"github.com/aarzilli/nucular"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/nuklear/handlers"
+)
+
+type Handler interface {
+	BeforeRender()
+	Render(*nucular.Window)
+	SetWalletMiddleware(walletcore.Wallet)
```
itswisdomagain: As explained above, this should go and `Render` should be modified to accept `walletcore.Wallet`

nuklear/handlers/transactions.go
```diff
@@ -0,0 +1,75 @@
...
+
+type TransactionsHandler struct {
+	err                    error
+	isRendering            bool
+	transactions           []*walletcore.Transaction
+	wallet                 walletcore.Wallet
+	hasFetchedTransactions bool
+}
+
+func (handler *TransactionsHandler) fetchTransactions() {
```
itswisdomagain: This method is called by `Render`, it should appear below the `Render` function

nuklear/desktop.go
```diff
@@ -57,87 +60,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+		H: area.H,
 	}
 
-	w.LayoutSpacePushScaled(rect)
-	// style navigation pane
-	setNavStyle(d.window)
+	w.LayoutSpacePushScaled(navRect)
+	// render nav
+	helpers.SetNavStyle(desktop.masterWindow)
 	if sw := w.GroupBegin("Navigation Group", 0); sw != nil {
```
itswisdomagain: - You can use `helpers.NewWindow` here
- What does `sw` mean? Maybe use a more descriptive variable name
- If `sw` is nil, then we have no navigation window, without a navigation window, the app is not usable, this scenario should be attended to, perhaps show an error message

itswisdomagain (changes requested):

nuklear/desktop.go
```diff
@@ -57,87 +61,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+	helpers.SetContentArea(contentRect)
+
 	// style content area
-	d.setPageStyle()
+	helpers.SetPageStyle(desktop.masterWindow)
 
-	contentArea = rect
-	w.LayoutSpacePushScaled(rect)
+	w.LayoutSpacePushScaled(contentRect)
+	handler := desktop.handlers[desktop.currentPage]
```
itswisdomagain: this should go after the if block below, only makes sense to do this after confirming that `desktop.currentPage != nil`

nuklear/desktop.go
```diff
@@ -57,87 +61,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+	if desktop.currentPage == "" { // ideally, this should only be false once in the lifetime of an instance
+		desktop.changePage(homePage)
+		return
+	}
 
-	if pageHandler, ok := d.pageHandlers[d.currentPage]; ok {
-		pageHandler(w)
-	} else {
-		d.pageHandlers[homePage](w)
+	// ensure that the handler's BeforeRender function is called only once per page call
```
itswisdomagain, line 119: could add more information to this comment on why this is necessary; why should BeforeRender be called only once per page call?

nuklear/desktop.go
```diff
@@ -57,87 +61,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
-		sw.Row(40).Dynamic(1)
-		if sw.Button(label.TA("Balance", "LC"), false) {
-			d.gotoPage("balance")
-		}
-		if sw.Button(label.TA("Send (WIP)", "LC"), false) {
-			d.gotoPage("send")
+	w.LayoutSpacePushScaled(navRect)
+	// render nav
+	helpers.SetNavStyle(desktop.masterWindow)
+	if contentWindow := helpers.NewWindow("Navigation Group", w, 0); contentWindow != nil {
```
itswisdomagain: more like `navWindow` or `navArea` than `contentWindow`

nuklear/desktop.go
```diff
@@ -57,87 +61,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
 
-func (d *Desktop) createContentPane(w *nucular.Window, height, width int) {
-	rect := rect.Rect{
+	// create content pane
+	contentRect := rect.Rect{
 		X: navWidth - 45,
 		Y: 0,
-		W: (width + 55) - navWidth,
-		H: height,
+		W: (area.W + 55) - navWidth,
```
itswisdomagain: What's the significance of `45` and `55` used here? Would be a good idea to use constant variables with meaningful, descriptive names

nuklear/desktop.go
```diff
@@ -57,87 +61,65 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
-	setNavStyle(d.window)
-	if sw := w.GroupBegin("Navigation Group", 0); sw != nil {
-		sw.Row(40).Dynamic(1)
-		if sw.Button(label.TA("Balance", "LC"), false) {
-			d.gotoPage("balance")
-		}
-		if sw.Button(label.TA("Send (WIP)", "LC"), false) {
-			d.gotoPage("send")
+	w.LayoutSpacePushScaled(navRect)
+	// render nav
```
itswisdomagain, line 86: Should this comment come before the previous line? If not, then there should be an empty single line space to show separation of concerns

nuklear/handlers/balance.go
```diff
@@ -0,0 +1,60 @@
...
+
+func (handler *BalanceHandler) Render(w *nucular.Window, wallet walletcore.Wallet) {
+	if !handler.isRendering {
+		handler.isRendering = true
+		handler.accounts, handler.err = wallet.AccountsOverview(walletcore.DefaultRequiredConfirmations)
+	}
+
+	// draw page
+	if page := helpers.NewWindow("Balance Page", w, 0); page != nil {
+		page.DrawHeader("Balance")
```
itswisdomagain, line 29: This creates an irregular shape on my mac, wonder why:

screenshot 2019-01-24 at 5 22 10 am

nuklear/desktop.go
itswisdomagain: Have you considered using [menu](https://godoc.org/github.com/aarzilli/nucular#Window.Menu) instead of manually drawing the nav area?

itswisdomagain (changes requested):
Really loving the changes introduced by this PR. A few things to note. Also, `droid-sans.bold.ttf` is not used, should be removed in next commit

nuklear/helpers/style.go
```diff
@@ -55,63 +66,42 @@ var colorTable = nstyle.ColorTable{
...
+	var fontInit sync.Once
 
-func newWindow(title string, w *nucular.Window, flags nucular.WindowFlags) *window {
-	if nw := w.GroupBegin(title, flags); nw != nil {
-		return &window{
-			nw,
+	fontInit.Do(func() {
+		fontData, err := ioutil.ReadFile(asset)
+		if err != nil {
+			log.Fatal(err)
```
itswisdomagain: Interfaces should ideally propagate errors back to main.go so that the loaded wallet is shutdown properly. The following modifications should make this work.

```
func InitFonts() (err error) {
	robotoMediumFont, err := getFont(navFontSize, navFontDPI, "nuklear/assets/font/Roboto-Medium.ttf")
	if err != nil {
		return
	}
	PageHeaderFont, NavFont = robotoMediumFont, robotoMediumFont

	PageContentFont, err = getFont(pageContentFontSize, pageContentFontDPI, "nuklear/assets/font/Roboto-Light.ttf")
	return
}

func getFont(fontSize, DPI int, asset string) (font.Face, error) {
	fontData, err := ioutil.ReadFile(asset)
	if err != nil {
		return nil, err
	}

	ttfont, err := freetype.ParseFont(fontData)
	if err != nil {
		return nil, err
	}

	size := int(float64(fontSize) * scaling)
	options := &truetype.Options{
		Size:    float64(size),
		Hinting: font.HintingFull,
		DPI:     float64(DPI),
	}

	return truetype.NewFace(ttfont, options), nil
}
```
#### Key changes
- Modify `InitFonts` and `getFont` to return `error`
- Modify line 42 of `nuklear/desktop.go` to propagate any `InitFonts` error
```
err := helpers.InitFonts()
if err != nil {
	return err
}
```

#### Other things to note:
- Don't see a need for `sync.Once`, the function is only called once when the app is started
- `PageHeaderFont` and `NavFont` use the same roboto-medium font, no need to re-read the font file multiple times
- `InitFonts` calls `getFont`, the functions should be arranged in that order for easier scanning

nuklear/helpers/style.go
```diff
@@ -66,18 +64,44 @@ var colorTable = nstyle.ColorTable{
...
-	})
+	PageContentFont, err = getFont(pageContentFontSize, pageContentFontDPI, robotoLightFontData)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func readFontFile(file string) ([]byte, error) {
```
itswisdomagain: can just use `ioutil.ReadFile` directly instead of creating a wrapper around the function

nuklear/helpers/style.go
```diff
@@ -66,18 +64,44 @@ var colorTable = nstyle.ColorTable{
...
+	if err != nil {
+		return err
+	}
+
+	robotoLightFontData, err := readFontFile("nuklear/assets/font/Roboto-Light.ttf")
+	if err != nil {
+		return err
+	}
+
+	NavFont, err = getFont(navFontSize, navFontDPI, robotoMediumFontData)
```
itswisdomagain, line 78: I prefer the previous `getFont` function that takes the asset string rather than this, above, there are 2 references to `readFontFile` which really is `ioutil.ReadFile` - a repetition that can be avoided if the call to `ioutil.ReadFile` is made in the `getFont` function

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #176, opened 2019-01-20
send feature improvements (web)

### LINKED ISSUE #140: send feature improvements (web) - multiple destination addresses, utxo selection improvments
Add support for multiple destination addresses in send page. This has been implemented in the cli interface (#119) following instructions in #113.

The format for displaying utxos during a send tx operation has also been updated on cli (see #114). Depending on the approved fix for #57, there may be a need to implement similar improvements in the web interface.

ademuanthony:
Resolves #140 

Added the functionality for sending tx to multiple destinations and worked on the display of the utxos on the web to match that of the cli
send-feature-web

### DIFF
--- a/web/public/js/send.js
+++ b/web/public/js/send.js
@@ -181,22 +181,23 @@ function submitSendForm() {
  *                    MULTI-ADDRESS FUNCTIONS                        *
  *===================================================================*/
 function newDestination() {
-    let html = `<div class="col-md-6 col-sm-12">
-                    <div class="form-group">
-                        <label>Destination Address</label>
-                        <input type="text" class="form-control" name="destination-address" />
+    let html = `<div class="row">
+                    <div class="col-md-6 col-sm-12">
+                        <div class="form-group">
+                            <label>Destination Address</label>
+                            <input type="text" class="form-control" name="destination-address" />
+                        </div>
                     </div>
-                </div>
-                <div class="col-md-6 col-sm-12">
-                    <div class="form-group">
-                        <label for="amount-">Amount (DCR)</label>
-                        <input type="number" class="form-control amount" name="destination-amount" />
+                    <div class="col-md-6 col-sm-12">
+                        <div class="form-group">
+                            <label for="amount-">Amount (DCR)</label>
+                            <input type="number" class="form-control amount" name="destination-amount" />
+                        </div>
                     </div>
                 </div>
     `
     $("#destinations").append(html)
 }
-
 
 /**==================================================================*
  *                    PASSPHRASE FUNCTIONS                           *
@@ -282,7 +283,17 @@ $(function(){
     });
 
     $("#add-destination-btn").on("click", function (e) {
-        newDestination()
+        newDestination();
+        $("#remove-destination-btn").show();
+    })
+
+    $("#remove-destination-btn").hide();
+
+    $("#remove-destination-btn").on("click", function () {
+        $("#destinations .row:last-child").remove();
+        if($("#destinations .row").length < 2) {
+            $("#remove-destination-btn").hide();
+        }
     })
 
     $("#submit-btn").on("click", function(e){
--- a/web/public/js/send.js
+++ b/web/public/js/send.js
@@ -49,7 +49,7 @@ function getSelectedInputsSum() {
 function getTotalSendAmount() {
     let amount = 0;
     for(let i = 1; i <= destinationCount; i++) {
-        amount += $(`#amount-${i}`).val()
+        amount += parseFloat($(`#amount-${i}`).val())
     }
     return amount;
 }
@@ -88,9 +88,10 @@ function openCustomizePanel(get_unconfirmed) {
             var dcrAmount = tx.amount / 100000000;
             return  "<tr>" + 
                         "<td width='5%'><input type='checkbox' class='custom-input' name='utxo' value="+ tx.key+" data-amount='" + dcrAmount + "' /></td>" +
-                        "<td width='50%'>" + tx.key + "</td>" + 
-                        "<td width='20%'>" + dcrAmount + " DCR</td>" + 
-                        "<td width='25%'>" + receiveDateTime.toString().split(' ').slice(0,5).join(' '); + "</td>" +
+                        "<td width='40%'>" + tx.address + "</td>" +
+                        "<td width='15%'>" + dcrAmount + " DCR</td>" +
+                        "<td width='25%'>" + receiveDateTime.toString().split(' ').slice(0,5).join(' ') + "</td>" +
+                        "<td width='15%'>" + tx.confirmations + " confirmation(s)</td>" +
                     "</tr>"
         });
         $("#custom-tx-row tbody").html(utxoHtml.join('\n'));
@@ -101,7 +102,7 @@ function openCustomizePanel(get_unconfirmed) {
            calculateSelectedInputPercentage();
         });
 
-        $(".amount").on("keyup", function(){
+        $("#destinations").on("keyup", ".amount", function(){
             validateDestinationFields();
             calculateSelectedInputPercentage();
         });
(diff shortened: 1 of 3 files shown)


engprodigy:
I need an extra address I can send to to test for when sending to multiple addresses. @ademuanthony Please can you provide a receiving I can ass to the one I have to test.

web/public/js/send.js
```diff
@@ -248,6 +278,11 @@ $(function(){
         }
     });
 
+    $("#add-destination-btn").on("click", function (e) {
+        e.preventDefault();
```
itswisdomagain: If you add `type="button"` to `#add-destination-btn`, you won't need to prevent default. See https://stackoverflow.com/a/41904218

web/public/js/send.js
```diff
@@ -20,11 +21,7 @@ function validateSendForm() {
...
     }
 
-    isClean = validateAmountField();
-
-    if ($("#destination-address").val() === "") {
-        errors.push("The destination address is required");
-    }
+    isClean = validateDestinationFields();
 
     if ($("#use-custom").prop("checked") && (getSelectedInputsSum() < $("#amount").val()) ) {
```
itswisdomagain: replace `$("#amount").val()` with `getTotalSendAmount()`

ademuanthony:
> I need an extra address I can send to to test for when sending to multiple addresses. @ademuanthony Please can you provide a receiving I can ass to the one I have to test.

TsodqRbyRz2j164YcyyLJC2WuF8Z8hx6S5u

engprodigy (commented):
the additional input addresses are not able to receive dcr token, only the first input address receives dcr token. Its not yet working perfectly.

engprodigy:
@ademuanthony what is select custom input all about? I'm new to the platform and I'm trying to review this PR. I think it will help me understand the whole process occurring in the  send page. So I will be able to properly review the code. Thanks.

web/public/js/send.js
```diff
@@ -1,12 +1,13 @@
 /**==================================================================*
  *                  SEND PAGE FUNCTIONS                              *
  *===================================================================*/
-function validateAmountField() { 
-    if ($("#amount").val() === "") {
-        $(".errors").html("<div class='error'>Please enter an amount first</div>");
-        return false;
+function validateDestinationFields() {
+    for(let i = 1; i <= destinationCount; i++) {
```
itswisdomagain: Can iterate through the inputs in the `#destination` div:
```
for (const field of $('#destinations input')) {
    if ($(field).val() === "") {
        $(".errors").html("<div class='error'>The destination address and amount are required</div>");
        return false;
    }
}
```

itswisdomagain: Also a good idea to check that the amount is a positive, non-zero number as part of the validation

web/public/js/send.js
```diff
@@ -167,6 +173,30 @@ function submitSendForm() {
...
+ *                    MULTI-ADDRESS FUNCTIONS                        *
+ *===================================================================*/
+var destinationCount = 1;
+
+function newDestination() {
+    destinationCount++;
+    let html = `<div class="col-md-6 col-sm-12">
+                    <div class="form-group">
+                        <label for="destination-address-${destinationCount}">Destination Address ${destinationCount}</label>
+                        <input type="text" class="form-control" id="destination-address-${destinationCount}" name="destination-address[${destinationCount}]" />
```
itswisdomagain: Should probably use an array instead of the amounts and addresses. Using the same form name in multiple form fields will produce an array when the form data is parsed on the backend. I tested with `destination-address` and `destination-amount` as name attributes for the destination fields and was able to get the collated information as an array in the backend (more on that in a separate comment)

Using the same form name would mean we won't need the `destinationCount` variable and the id attribute can be done without as well, to prevent multiple fields having the same id, also since the id attribute isn't really needed

itswisdomagain: Put succinctly, this line should be:
```
<input type="text" class="form-control" name="destination-address" />
```

web/public/js/send.js
```diff
@@ -167,6 +173,30 @@ function submitSendForm() {
...
+                    </div>
+                </div>
+                <div class="col-md-6 col-sm-12">
+                    <div class="form-group">
+                        <label for="amount-">Amount (DCR)</label>
+                        <input type="number" class="form-control amount" id="amount-${destinationCount}" name="amount[${destinationCount}]" />
```
itswisdomagain: This line should simplify to
```
<input type="number" class="form-control amount" name="destination-amount" />
```

web/routes/handlers.go
```diff
@@ -78,31 +78,38 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
...
-	destAddress := req.FormValue("destination-address")
 	passphrase := req.FormValue("wallet-passphrase")
 	spendUnconfirmed := req.FormValue("spend-unconfirmed")
 	useCustom := req.FormValue("use-custom")
 
-	amount, err := strconv.ParseFloat(amountStr, 64)
-	if err != nil {
-		data["error"] = err.Error()
-		return
+	var sendDestinations []txhelper.TransactionDestination
```
itswisdomagain: Going with the approach of using an array for the address and amount input fields, the `sendDestinations` slice can be populated thus:
```
destinationAddresses := req.Form["destination-address"]
destinationAmounts := req.Form["destination-amount"]

sendDestinations := make([]txhelper.TransactionDestination, len(destinationAddresses))
for i := range destinationAddresses {
	amount, err := strconv.ParseFloat(destinationAmounts[i], 64)
	if err != nil {
		data["error"] = err.Error()
		return
	}
	sendDestinations[i] = txhelper.TransactionDestination{
		Address: destinationAddresses[i],
		Amount: amount,
	}
}
```

web/routes/handlers.go
```diff
@@ -78,31 +78,38 @@ func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Reques
...
+			data["error"] = err.Error()
+			return
+		}
+		sendDestinations = append(sendDestinations, txhelper.TransactionDestination{
+			Address: destAddress, Amount: amount,
+		})
+		destinationCount++
 	}
 
+
```
itswisdomagain: additional empty line space can go

itswisdomagain (changes requested):
In addition to the comments below, we need a way to remove destination fields, say a user mistakenly added 5 fields but only wants to send to 3 or 4 destinations, currently, the user will have to refresh and start over again.

Please also modify the `getWalletPassphraseAndSubmit` function in `send.js` to prevent multiple send attempts when the send button is clicked:
```
var passphraseModal = $("#passphrase-modal");
var submitPassphrase = $("#passphrase-submit");

submitPassphrase.off("click");
submitPassphrase.on("click", function(){
    if (validatePassphrase()) {
        passphraseModal.modal('hide');
        submitSendForm();
    }
});
passphraseModal.modal();
```

ademuanthony:
send-feature-multi-destination-web

web/public/js/send.js
```diff
@@ -248,6 +282,20 @@ $(function(){
         }
     });
 
+    $("#add-destination-btn").on("click", function (e) {
```
itswisdomagain: `e` is not used

web/views/send.html
```diff
@@ -31,18 +31,31 @@
...
+                                        <div class="col-md-6 col-sm-12">
+                                            <div class="form-group">
+                                                <label>Amount (DCR)</label>
+                                                <input type="number" class="form-control amount" name="destination-amount" />
+                                            </div>
                                         </div>
                                     </div>
                                 </div>
+                                <div class="form-group">
+                                    <button id="add-destination-btn" type="button" class="btn btn-sm">
```
itswisdomagain: Add `btn-outline-secondary` class to the buttons so they actually look like buttons. Also use `btn-primary` or something for the final send button as well.

itswisdomagain (changes requested):
I think this is good to merge, just some cleanup.
Also help to simplify line 221 of `web/handlers.go` to `if getUnconfirmed == "true" {`

itswisdomagain (approved):

PR #185, opened 2019-01-25
StakeInfo page (nuklear)

yaowork:
This adds a passphrase widget and stake info page to the nuklear app

### DIFF
--- a/nuklear/handlers/stakeinfo.go
+++ b/nuklear/handlers/stakeinfo.go
@@ -68,7 +68,3 @@ func (handler *StakeInfoHandler) Render(window *nucular.Window, wallet walletcor
 		pageWindow.End()
 	}
 }
-
-func (handler *StakeInfoHandler) AfterRender() {
-
-}
--- a/nuklear/handlers/stakeinfo.go
+++ b/nuklear/handlers/stakeinfo.go
@@ -1,0 +1,74 @@
+package handlers
+
+import (
+	"context"
+	"strconv"
+
+	"github.com/aarzilli/nucular"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/nuklear/helpers"
+)
+
+type StakeInfoHandler struct {
+	err         error
+	isRendering bool
+	stakeInfo   *walletcore.StakeInfo
+}
+
+func (handler *StakeInfoHandler) BeforeRender() {
+	handler.err = nil
+	handler.stakeInfo = nil
+	handler.isRendering = false
+}
+
+func (handler *StakeInfoHandler) Render(window *nucular.Window, wallet walletcore.Wallet) {
+	if !handler.isRendering {
+		handler.isRendering = true
+		handler.stakeInfo, handler.err = wallet.StakeInfo(context.Background())
+	}
+
+	// draw page
+	if pageWindow := helpers.NewWindow("StakeInfo Page", window, 0); pageWindow != nil {
+		pageWindow.DrawHeader("Stake Info")
+
+		// content window
+		if contentWindow := pageWindow.ContentWindow("StakeInfo Content"); contentWindow != nil {
+			if handler.err != nil {
+				contentWindow.SetErrorMessage(handler.err.Error())
+			} else {
+				contentWindow.Row(20).Ratio(0.07, 0.09, 0.06, 0.07, 0.07, 0.08, 0.12, 0.09, 0.09, 0.09, 0.1)
+				contentWindow.Label("Expired", "LC")
+				contentWindow.Label("Immature", "LC")
+				contentWindow.Label("Live", "LC")
+				contentWindow.Label("Revoked", "LC")
+				contentWindow.Label("Unmined", "LC")
+				contentWindow.Label("Unspent", "LC")
+				contentWindow.Label("AllmempoolTix", "LC")
+				contentWindow.Label("PoolSize", "LC")
+				contentWindow.Label("Missed", "LC")
+				contentWindow.Label("Voted", "LC")
+				contentWindow.Label("Total Subsidy", "LC")
+
+				contentWindow.Row(20).Ratio(0.07, 0.09, 0.06, 0.07, 0.07, 0.08, 0.12, 0.09, 0.09, 0.09, 0.1)
+				contentWindow.Label(strconv.Itoa(int(handler.stakeInfo.Expired)), "LC")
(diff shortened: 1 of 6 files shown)


nuklear/handlers/stakeinfo.go
```diff
@@ -0,0 +1,74 @@
...
+				contentWindow.Label(strconv.Itoa(int(handler.stakeInfo.TotalSubsidy)), "LC")
+
+			}
+			contentWindow.End()
+		}
+		pageWindow.End()
+	}
+}
+
+func (handler *StakeInfoHandler) AfterRender() {
```
itswisdomagain: `StakeInfoHandler` does not implement `Widget`, this method is unnecessary.

nuklear/handlers/transactions.go
```diff
@@ -58,15 +58,16 @@ func (handler *TransactionsHandler) Render(window *nucular.Window, wallet wallet
 					}
 				}
 			} else {
-				widgets.ShowIsFetching(contentWindow)
+				widgets.Run("loading", contentWindow.Window, func() {})
```
itswisdomagain: Allow passing nil instead of an empty function

nuklear/handlers/widgets/passphrase.go
```diff
@@ -0,0 +1,83 @@
+package widgets
```
itswisdomagain: I have some comments on this file but would prefer to not get into them until I see some code that uses this widget. Best to temporarily remove this file and references to it just for this PR

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"fmt"
+
+	"github.com/aarzilli/nucular"
+)
+
+type Widget interface {
+	BeforeRender(window *nucular.Window)
+	Render(finishHandler func())
+	AfterRender()
```
itswisdomagain: What's `AfterRender` supposed to do? None of the 2 implementations of `Widget` so far perform any operation with this function, it's always empty.

Best to remove the method until it's use becomes apparent

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+var widgets = map[string]Widget{
+	"passphrase": NewPassphraseWidget(),
+	"loading":    NewLoadingWidget(),
+}
+
+func RegisterWidget(key string, widget Widget) error {
```
itswisdomagain: This function is not used and can be safely removed

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/aarzilli/nucular"
+)
+
+type Widget interface {
+	BeforeRender(window *nucular.Window)
+	Render(finishHandler func())
+	AfterRender()
+}
+
+var widgets = map[string]Widget{
```
itswisdomagain: Should find a way to use a function, so the widgets are not instantiated until necessary.

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+	widgets[key] = widget
+	return nil
+}
+
+func GetWidget(key string) (Widget, error) {
+	if widget, ok := widgets[key]; ok {
+		return widget, nil
+	}
+
+	return nil, fmt.Errorf("Widget :%s is not registered. Forgotten import?", key)
```
itswisdomagain: Probably should be `Widget: %s` (notice the position of the colon)
The last part of this error message, `Forgotten import?` doesn't make sense. What import? Also, will this message be displayed to the user? If so, it can be better constructed.

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func Run(key string, window *nucular.Window, finishHandler func()) error {
+	widget, err := GetWidget(key) // this will panic if widget is no registered
```
itswisdomagain: no => **not**

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+	if err != nil {
+		return err
+	}
+	widget.BeforeRender(window)
```
itswisdomagain: should add a line space after the if block

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+	widget.Render(finishHandler)
+	widget.AfterRender()
```
itswisdomagain: What's the difference between `finishHandler` and `AfterRender`? Can they be merged?
Also, `widget.BeforeRender` is called sequentially with `widget.Render`, they should be merged into one function: `Render(window, finishHandler)`

nuklear/handlers/widgets/widget.go
```diff
@@ -0,0 +1,46 @@
...
+
+func GetWidget(key string) (Widget, error) {
+	if widget, ok := widgets[key]; ok {
+		return widget, nil
+	}
+
+	return nil, fmt.Errorf("Widget :%s is not registered. Forgotten import?", key)
+}
+
+func Run(key string, window *nucular.Window, finishHandler func()) error {
```
itswisdomagain: I suppose the idea of this `widgets` package is to hold controls/views that can be used from different pages. Having to pass the key for each widget from different call sites could cause issues if there's a typo in one of the call sites. Also, if the key changes for any reason, you have to manually look for all its usages.

Better to make this a private method `run` and create wrappers for each widget that can be used at the call sites. For example, `Run("loading",....)` can become:
```
func ShowLoadingWidget(window *nucular.Window, finishHandler func()) error {
    return run("loading", window, finishHandler)
}
```

nuklear/handlers/widgets/loading.go
```diff
@@ -1,10 +1,25 @@
...
+
+func NewLoadingWidget() Widget {
+	return &LoadingWidget{}
+}
+
+func (l *LoadingWidget) BeforeRender(window *nucular.Window) {
+	l.window = window
+}
+
+func (l *LoadingWidget) Render(finishHandler func()) {
```
itswisdomagain: If you merge `BeforeRender` and `Render`, there'd be no need to save a copy of `window` in `LoadingWidget. window`, you can just use it directly. There really is no need to save the copy actually, it's never used at a later time, always used immediately after it's set and that's it. So why not use it at once and keep it at that?

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #186, opened 2019-01-26
change outputs customization - web

### LINKED ISSUE #139: send feature improvement (web only) - change outputs customization
Beginning with #123, the `sendcustom` command on cli allows users to customize the change outputs for created transactions. Similar behaviour should be added to the send page on the web interface using collapsible div(s).

It is necessary to test the behaviour on the cli interface in order to replicate same behaviour in the web interface. The flow on cli is something like this:
- User is asked if to  use automatically generated change outputs or manually specify
- If manual, user is asked to enter a change amount. If the amount entered is less than the estimated change for the transaction, the user is asked to enter another amount and this is repeated until the estimated tx change is accounted for. Each amount entered is assigned to a different change address.
- If automatic, user is asked to specify how many change outputs should be used. The cli app automatically generates addresses to use for change outputs and assigns amounts to each address. Total amounts assigned is equal to the estimated change for the transaction.

Below is one way to include the change output customization behaviour on the web interface:
- Add a collapsible div for "Automatic Change Outputs", if this div is expanded, user should be asked to enter the number of change outputs to generate
- User should be able to click "Generate Outputs", which should perform an ajax call to generate change addresses with random amounts and display to the user
- Add another collapsible div for "Custom Change Outputs", this should have a button "Add Change Output". The button click should perform an ajax call to generate an address and estimate the change for the transaction.
- After the address is generated and returned with the estimated change, the address should be displayed to the user with an input field to enter amount to assign to the generated change address. The amount entered cannot be higher than the estimated change amount.
- If the user enters an amount lower than the estimated change amount, the "Add Change Output" button should be visible to allow user generate another address and assign another amount. This should continue until the total amount the user has assigned to change addresses covers the estimated change.

ademuanthony:
Resolve #139 

This PR adds change output customization to the web. 

This feature is available only when the user chose to specify a custom input
co-no-show-if-not-custom

It gives the user the ability to auto-generate random amounts for his desired number of random output
co-auto-change

 or to manually add change amount for a desired number of destinations.
co-custom-change


web/public/js/send.js
```diff
@@ -296,6 +420,38 @@ $(function(){
         }
     })
 
+
+    $("#changeOutputsCard").hide();
```
itswisdomagain: best to hide this directly in html to prevent the flicker that is visible when the page loads. At first the element is visible, then it quickly hides. Can be avoided.

Can use `class="d-none"` on those elements and when you need to make them visible, use `element.show()` or `element.removeClass('d-none')`, if the `show` doesn't work.

web/views/send.html
```diff
@@ -82,6 +82,69 @@
                                 </table>
                             </div>
                         </div>
+
+                        <div id="changeOutputsCard">
+                            <div class="card">
+                                  <div class="card-header" id="headingOne">
+                                    <h5 class="mb-0">
```
itswisdomagain: `h5` element is unnecessary, can use button directly

web/views/send.html
```diff
@@ -82,6 +82,69 @@
...
+                                      </button>
+                                    </h5>
+                                  </div>
+                              
+                                  <div id="automaticChangeOutputPnl" class="collapse" aria-labelledby="headingOne" data-parent="#changeOutputsCard">
+                                    <div class="card-body">
+                                        <div class="row">
+                                            <div class="col-md-6 col-sm-12">
+                                                <div class="form-group">
+                                                    <label for="amount">Number Of Change Outputs</label>
```
itswisdomagain: `for="numberOfChangeOutput"`

itswisdomagain (changes requested):

itswisdomagain:
Work on this is paused temporarily for PR #190 to be completed and merged in order to avoid over-complicating the already complicated js files

ademuanthony:
I have migrated the code base to stimulusjs
I also added the ability to reset the form after a successful send operation and a reset button.

Based on my discussion with @raedah we now have a single panel for the change output and a checkbox for the user to specify whether or not to use a random amount

change-output-no-random

change-output-random

web/routes/handlers.go
```diff
@@ -233,6 +205,84 @@ func (routes *Routes) getUnspentOutputs(res http.ResponseWriter, req *http.Reque
...
+
+	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, int(nChangeOutputs), int64(totalInputAmount), sourceAccount, len(utxos), destinations)
+	if err != nil {
+		data["error"] = "276" + err.Error()
+		return
+	}
+	data["message"] = changeOutputDestinations
+}
+
+func buildTxDestinations(destinationAddresses []string, destinationAmounts []string) (destinations []txhelper.TransactionDestination, err error) {
```
itswisdomagain: This should go into a helper file in this package

web/routes/routes.go
```diff
@@ -62,6 +62,7 @@ func (routes *Routes) registerRoutesRequiringWallet(router chi.Router) {
 	router.Get("/receive", routes.receivePage)
 	router.Get("/generate-address/{accountNumber}", routes.generateReceiveAddress)
 	router.Get("/unspent-outputs/{accountNumber}", routes.getUnspentOutputs)
+	router.Post("/random-change-outputs", routes.getRandomChangeOutputs)
```
itswisdomagain: This should be a "GET" request, the route function even says as much

web/src/controllers/send_controller.js
```diff
@@ -3,7 +3,7 @@ import axios from 'axios'
 
 export default class extends Controller {
   static get targets () {
-    return ['sourceAccount', 'address', 'amount', 'destinations', 'destinationTemplate', 'changeAddress', 'changeAmount', 'errors', 'customInput', 'customTxRow', 'customInputContent', 'changeOutputsTarget', 'submitButton', 'nextButton', 'removeDestinationButton', 'form', 'walletPassphrase', 'passwordError', 'useCustom', 'spendUnconfirmed', 'errorMessage', 'successMessage', 'progressBar']
+    return ['sourceAccount', 'address', 'amount', 'destinations', 'destinationTemplate', 'changeAddress', 'changeAmount', 'errors', 'customInput', 'customTxRow', 'customInputContent', 'submitButton', 'nextButton', 'removeDestinationButton', 'form', 'walletPassphrase', 'passwordError', 'useCustom', 'spendUnconfirmed', 'errorMessage', 'successMessage', 'progressBar', 'changeOutputsCard', 'changeOutputPnl', 'numberOfChangeOutputs', 'useRandomChangeOutputs', 'generateOutputsButton', 'changeOutputContent', 'changeDestinationTemplate', 'changeOutputAddress', 'changeOutputAmount']
```
itswisdomagain: Please break this into multiple lines

web/src/controllers/send_controller.js
```diff
@@ -149,6 +139,88 @@ export default class extends Controller {
...
+    }, function () {
+      _this.generateOutputsButtonTarget.removeAttribute('disabled')
+      _this.generateOutputsButtonTarget.innerHTML = 'Generate Change Outputs'
+      _this.numberOfChangeOutputsTarget.removeAttribute('disabled')
+    })
+  }
+
+  getRandomChangeOutputs (numberOfOutputs, successCallback, completeCallback) {
+    let postData = $('#send-form').serialize()
+    postData += `&totalSelectedInputAmountDcr=${this.getSelectedInputsSum()}`
```
itswisdomagain: make this into `queryParams`, I think prepending the output of `$('#send-form').serialize()` with "?" should do the trick

web/src/controllers/send_controller.js
```diff
@@ -149,6 +139,88 @@ export default class extends Controller {
...
+
+    // add source-account value to post data if source-account element is disabled
+    if (this.sourceAccountTarget.disabled) {
+      postData += `&source-account=${this.sourceAccountTarget.value}`
+    }
+
+    postData += `&nChangeOutput=${numberOfOutputs}`
+
+    let _this = this
+    $.ajax({
```
itswisdomagain: Why are we not using axios?

web/views/send.html
```diff
@@ -83,10 +86,70 @@
                                 </table>
                             </div>
                         </div>
+
+                        <div id="changeOutputsCard" data-target="send.changeOutputsCard" class="d-none">
+                            <div class="card">
+                                  <div class="card-header" id="headingOne">
+                                    <h5 class="mb-0">
```
itswisdomagain: Surrounding `h5` element is unnecessary.

web/src/controllers/send_controller.js
```diff
@@ -187,28 +258,67 @@ export default class extends Controller {
     })
   }
 
+  resetSendForm () {
```
itswisdomagain: this function should also clear error and success messages

web/views/send.html
```diff
@@ -83,10 +86,70 @@
...
+                                  </div>
+                            </div>
+
+                        </div>
+
                         <div class="card">
                             <div class="card-body">
                                 <div class="errors" data-target="send.errors"></div>
                                 <button data-target="send.nextButton" data-action="click->send#getWalletPassphraseAndSubmit" class="btn btn-default" id="submit-btn" type="button">Next</button>
+                                <button data-action="click->send#resetSendForm" class="btn btn-danger" id="reset-btn" type="button">Reset</button>
```
itswisdomagain, line 179: The reset button should be on the left

itswisdomagain: Also the "Next' button on the passphrase modal should be "Submit" please

web/views/send.html
```diff
@@ -83,10 +86,70 @@
...
+                                    <div class="card-body">
+                                        <div class="row">
+                                            <div class="col-sm-12 form-inline">
+
+                                                <label class="sr-only" for="numberOfChangeOutput">Number Of Change Outputs</label>
+                                                <input data-target="send.numberOfChangeOutputs" type="text" class="form-control mb-2 mr-sm-2" type="number" class="form-control" id="numberOfChangeOutput" name="numberOfChangeOutput"
+                                                       placeholder="Number Of Outputs">
+
+                                                <div class="form-check mb-2 mr-sm-2">
+                                                    <input data-target="send.useRandomChangeOutputs" class="form-check-input" type="checkbox" id="inlineFormCheck">
```
itswisdomagain: `send#generateChangeOutputs` should be triggered when this checkbox is checked or unchecked

itswisdomagain (changes requested):
**Some observations:**
- If I send using custom inputs without **generating** change outputs, I get a success message without an hash, the transaction does not go through and my balance is unaffected. This happens even if I enter number of change outputs

- New destination addresses should be added below the previous not to the side for easier scanning by the eyes
screenshot 2019-02-28 at 1 11 25 pm

**Some ideas:**
- The change outputs section should not be collapsible, should be always visible. By default, it should assume 1 output with random amount.
- The change outputs should be generated whenever the number value changes or the checkbox is checked/unchecked
- If the user wants to manually enter amounts for each change address, there should be an indication somewhere of the maximum change address that the user can assign. This value should reduce as the user assigns amounts to different addresses and must be equal to zero before the transaction should be submitted

web/static/app/src/controllers/send_controller.js
```diff
@@ -176,39 +283,81 @@ export default class extends Controller {
...
     }).then(() => {
-      _this.nextButtonTarget.innerHTML = 'Send'
+      _this.nextButtonTarget.innerHTML = 'Next'
       _this.nextButtonTarget.removeAttribute('disabled')
     })
   }
 
+  resetSendForm () {
+    this.resetCustomizePanel()
+    while (this.destinationCount() > 1) {
```
itswisdomagain: best to call this function only once and save the value in a variable, then use a for loop
the current pattern makes multiple calls to the dom..

itswisdomagain (changes requested):
Delete `web/public` and restore `web/static/app/yarn.lock`

itswisdomagain (changes requested):
Validate the change outputs client-side before displaying the wallet passphrase prompt. In the screenshot below, the change outputs do not have amounts but I'm able to submit the form, which leads to a `strconv.ParseFloat: parsing "": invalid syntax` error
screenshot 2019-03-05 at 11 00 37 am

Also, if the change outputs section is hidden, I think that can be assumed to mean the user does not want to customize the change outputs and any previous customization done should be disregarded. In the screenshot above, I did not set amounts for the change outputs and that gave an error. I expected that by closing the change outputs section, it'd revert to the default autogenerated change output but it still gave the error gotten earlier.

web/static/app/src/controllers/send_controller.js
```diff
@@ -173,8 +160,60 @@ export default class extends Controller {
...
+    }
+
+    _this.changeOutputAmountTargets.forEach(function (ele) {
+      if (parseInt(ele.getAttribute('data-index')) === index) {
+        let amount = percentage / 100 * _this.totalChangeAmount
+        ele.value = amount
+      }
+    })
+
+    // todo validate amount constraint, reset element
```
itswisdomagain: should this be here?

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #189, opened 2019-01-28
Create wallet nuklear

yaowork:
This adds the create wallet page to the nuklear app. The page is displayed only when dcrwallet does not exist

### DIFF
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -18,17 +18,12 @@ const (
 	contentPaneWidthPadding = 55
 )
 
-type pageHandler struct {
-	handler    Handler
-	standalone bool
-}
-
 type Desktop struct {
 	masterWindow     nucular.MasterWindow
 	walletMiddleware app.WalletMiddleware
 	currentPage      string
 	pageChanged      bool
-	handlers         map[string]pageHandler
+	handlers         map[string]handlerData
 }
 
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
@@ -40,13 +35,9 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 
 	// register handlers
 	handlers := getHandlers()
-	desktop.handlers = make(map[string]pageHandler, len(handlers))
+	desktop.handlers = make(map[string]handlerData, len(handlers))
 	for _, handler := range handlers {
-		handlerItem := pageHandler{
-			standalone: handler.standalone,
-			handler:    handler.handler,
-		}
-		desktop.handlers[handler.name] = handlerItem
+		desktop.handlers[handler.name] = handler
 	}
 
 	// open wallet and start blockchain syncing in background
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -2,15 +2,12 @@ package nuklear
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/aarzilli/nucular"
 	"github.com/aarzilli/nucular/label"
 	"github.com/aarzilli/nucular/rect"
 	"github.com/raedahgroup/godcr/app"
-	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/raedahgroup/godcr/nuklear/helpers"
-	"github.com/raedahgroup/godcr/nuklear/nuklog"
 )
 
 const (
@@ -22,27 +19,16 @@ const (
 )
 
 type Desktop struct {
-	masterWindow nucular.MasterWindow
-	wallet       walletcore.Wallet
-	currentPage  string
-	pageChanged  bool
-	handlers     map[string]Handler
+	masterWindow     nucular.MasterWindow
+	walletMiddleware app.WalletMiddleware
+	currentPage      string
+	pageChanged      bool
+	handlers         map[string]Handler
 }
 
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
 	desktop := &Desktop{
-		wallet: walletMiddleware,
-	}
-
-	// initialize master window and set style
-	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, desktop.render)
-	window.SetStyle(helpers.GetStyle())
-	desktop.masterWindow = window
-
-	// initialize fonts for later use
-	err := helpers.InitFonts()
-	if err != nil {
-		return nil
+		walletMiddleware: walletMiddleware,
 	}
 
 	// register handlers
@@ -58,11 +44,21 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 		return err
 	}
 
+	renderFn := desktop.renderWithNav
 	if !walletExists {
-		// todo add ui to create wallet
-		err = fmt.Errorf("No wallet found. Use 'godcr create' to create a wallet before launching the desktop app")
-		// fmt.Println(err.Error())
-		nuklog.LogInfo(err.Error())
+		desktop.currentPage = "createwallet"
+		renderFn = desktop.renderWithoutNav
+		desktop.pageChanged = true
+	}
+
+	// initialize master window and set style
+	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, renderFn)
+	window.SetStyle(helpers.GetStyle())
+	desktop.masterWindow = window
+
+	// initialize fonts for later use
+	err = helpers.InitFonts()
+	if err != nil {
 		return err
 	}
 
@@ -73,7 +69,23 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	return nil
 }
 
-func (desktop *Desktop) render(w *nucular.Window) {
+func (desktop *Desktop) renderWithoutNav(window *nucular.Window) {
+	window.Row(0).SpaceBeginRatio(1)
+	window.LayoutSpacePushRatio(0.1, 0.05, 0.9, 0.8)
+
+	handler := desktop.handlers[desktop.currentPage]
+	// ensure that the handler's BeforeRender function is called only once per page call
+	// as it initializes page variables
+	if desktop.pageChanged {
+		handler.BeforeRender()
+		desktop.pageChanged = false
+	}
+
+	helpers.SetNoNavWindowStyle(window.Master())
+	handler.Render(window, desktop.walletMiddleware)
+}
+
+func (desktop *Desktop) renderWithNav(w *nucular.Window) {
 	area := w.Row(0).SpaceBegin(2)
 
 	// create nav pane
@@ -125,7 +137,7 @@ func (desktop *Desktop) render(w *nucular.Window) {
 		desktop.pageChanged = false
 	}
 
-	handler.Render(w, desktop.wallet)
+	handler.Render(w, desktop.walletMiddleware)
 }
 
 func (desktop *Desktop) changePage(page string) {
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -7,7 +7,6 @@ import (
 	"github.com/aarzilli/nucular/label"
 	"github.com/aarzilli/nucular/rect"
 	"github.com/raedahgroup/godcr/app"
-	"github.com/raedahgroup/godcr/nuklear/handlers"
 	"github.com/raedahgroup/godcr/nuklear/helpers"
 )
 
@@ -20,11 +19,12 @@ const (
 )
 
 type Desktop struct {
-	masterWindow     nucular.MasterWindow
-	walletMiddleware app.WalletMiddleware
-	currentPage      string
-	pageChanged      bool
-	handlers         map[string]Handler
+	masterWindow       nucular.MasterWindow
+	walletMiddleware   app.WalletMiddleware
+	currentPage        string
+	pageChanged        bool
+	handlers           map[string]Handler
+	standaloneHandlers map[string]Handler
 }
 
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
@@ -47,10 +47,11 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 
 	if !walletExists {
 		desktop.currentPage = "createwallet"
+		desktop.pageChanged = true
 	}
 
 	// initialize master window and set style
-	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, desktop.renderWithNav)
+	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, desktop.render)
 	window.SetStyle(helpers.GetStyle())
 	desktop.masterWindow = window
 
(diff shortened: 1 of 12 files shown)


nuklear/desktop.go
```diff
@@ -71,7 +67,28 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	return nil
 }
 
-func (desktop *Desktop) render(w *nucular.Window) {
+func (desktop *Desktop) renderWithoutNav(window *nucular.Window) {
```
itswisdomagain: `renderWithoutNav` is called by `renderWithNav`, please re-order the functions

nuklear/desktop.go
```diff
@@ -71,7 +67,28 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+		desktop.pageChanged = false
+	}
+
+	helpers.SetNoNavWindowStyle(window.Master())
+	handler.Render(window, desktop.walletMiddleware)
+}
+
+func (desktop *Desktop) renderWithNav(w *nucular.Window) {
+	if desktop.currentPage == "createwallet" {
+		desktop.renderWithoutNav(w)
```
itswisdomagain: You should create a `render` function that triggers either `renderWithNav` or `renderWithoutNav` depending on the  value of currentPage. It doesn't make sense for `renderWithNav` to trigger `renderWithoutNav`

nuklear/desktop.go
```diff
@@ -71,7 +67,28 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+	if desktop.pageChanged {
+		handler.BeforeRender()
+		desktop.pageChanged = false
+	}
+
+	helpers.SetNoNavWindowStyle(window.Master())
+	handler.Render(window, desktop.walletMiddleware)
+}
+
+func (desktop *Desktop) renderWithNav(w *nucular.Window) {
```
itswisdomagain: `window` is better than `w`

nuklear/handlers/createwallet.go
```diff
@@ -0,0 +1,133 @@
...
+			handler.confirmPasswordInput.Edit(contentWindow.Window)
+
+			passwordError, ok := handler.validationErrors["password"]
+			if ok {
+				contentWindow.LabelColored(passwordError, "LC", helpers.DangerColor)
+			}
+
+			if confirmPasswordError, ok := handler.validationErrors["confirmpassword"]; ok {
+				if passwordError != "" {
+					contentWindow.Label("", "LC")
```
itswisdomagain: Is this meant to create a space between the 2 error message labels? If so, it's a hack, doesn't look good.

nuklear/handlers/createwallet.go
```diff
@@ -0,0 +1,133 @@
...
+	password := string(handler.passwordInput.Buffer)
+	confirmPassword := string(handler.confirmPasswordInput.Buffer)
+	hasStoredSeed := handler.hasStoredSeed
+
+	if password == "" {
+		handler.validationErrors["password"] = "Wallet password is required"
+	}
+
+	if confirmPassword == "" {
+		handler.validationErrors["confirmpassword"] = "Please confirm wallet password"
```
itswisdomagain: This is unnecessary. The only validation for the confirm password field is if it matches with the password field. If the password field is empty and the confirm password field is empty, it matches. The only error is that the main password field should not be empty - and that is handled already

nuklear/handlers/createwallet.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if password != "" && confirmPassword != "" && password != confirmPassword {
+		handler.validationErrors["confirmpassword"] = "Both passwords do not match"
+	}
+
+	if !hasStoredSeed {
+		handler.validationErrors["hasstoredseed"] = "Please store seed and check this box"
+	}
+
+	if len(handler.validationErrors) > 0 {
```
itswisdomagain: can simplify to `return len(handler.validationErrors) > 0`

nuklear/handlers/createwallet.go
```diff
@@ -0,0 +1,133 @@
...
+		return true
+	}
+	return false
+}
+
+func (handler *CreateWalletHandler) submitForm(wallet app.WalletMiddleware) {
```
itswisdomagain: Don't see a need for this function, can use the body directly instead of having a function, it's just one line of code, it's not reused and does not improve readability

nuklear/helpers/style.go
```diff
@@ -15,7 +15,8 @@ import (
 )
 
 var (
-	whiteColor             = color.RGBA{0xff, 0xff, 0xff, 0xff}
+	WhiteColor             = color.RGBA{0xff, 0xff, 0xff, 0xff}
```
itswisdomagain: Will `color.RGBA{0, 0, 0, 0}` work? That's simpler

itswisdomagain (changes requested):

itswisdomagain:
I'm also unable to use the Create Wallet page. It keeps re-painting every time I move the mouse or try to type or check the checkbox. Does it work on your end?

nuklear/handlers/createwallet.go
yaowork: The space is needed. No alternatives

nuklear/desktop.go
```diff
@@ -71,8 +68,33 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+func (desktop *Desktop) render(window *nucular.Window) {
+	if desktop.currentPage == "createwallet" {
+		desktop.renderStandalone(window)
+		return
+	}
+
+	desktop.renderWithNav(window)
+}
+
+func (desktop *Desktop) renderStandalone(window *nucular.Window) {
```
itswisdomagain: `renderStandalonePage` sounds more descriptive

nuklear/desktop.go
```diff
@@ -21,27 +19,17 @@ const (
...
-	wallet       walletcore.Wallet
-	currentPage  string
-	pageChanged  bool
-	handlers     map[string]Handler
+	masterWindow       nucular.MasterWindow
+	walletMiddleware   app.WalletMiddleware
+	currentPage        string
+	pageChanged        bool
+	handlers           map[string]Handler
+	standaloneHandlers map[string]Handler
```
itswisdomagain: - `standaloneHandlers` does not contrast well with `handlers`, could use `standalonePageHandlers` and `navPageHandlers`
- when looping through all the handlers in lines 38-40 of this file, check if the `standalone` property is true and if so, don't add the handler to `handlers`, add it instead to `standalonePageHandlers`

nuklear/desktop.go
```diff
@@ -58,9 +46,18 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	}
 
 	if !walletExists {
-		// todo add ui to create wallet
-		err = fmt.Errorf("No wallet found. Use 'godcr create' to create a wallet before launching the desktop app")
-		fmt.Println(err.Error())
+		desktop.currentPage = "createwallet"
+		desktop.pageChanged = true
```
itswisdomagain: I think when the app launches, `desktop.pageChanged` should always be true; this would be best set in line 32 above when initializing the `desktop` object.

Also, `desktop.currentPage` should always be set when the app is launched. If there's no wallet, then `desktop.currentPage = "createwallet"` else `desktop.currentPage = "balance"`

nuklear/desktop.go
```diff
@@ -71,8 +68,33 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	return nil
 }
 
-func (desktop *Desktop) render(w *nucular.Window) {
-	area := w.Row(0).SpaceBegin(2)
+func (desktop *Desktop) render(window *nucular.Window) {
+	if desktop.currentPage == "createwallet" {
```
itswisdomagain: Here, could do something like
if `handler, ok := desktop.standalonePageHandlers[desktop.currentPage]`, call the `renderStandalonePage` function and pass `handler` as a parameter to the function

then else if `handler, ok := desktop.navPageHandlers[desktop.currentPage]`, call the `renderNavPage` function and pass `handler` in as a parameter as well

finally, if current page does not exist as a key in `desktop.standalonePageHandlers` and `desktop.navPageHandlers`, show an error message in `window`. Of course, this last case should not happen but good to write code like this to catch it in case it does happen. Also, this prepares the ground to support other possible ways of rendering a window.

nuklear/desktop.go
```diff
@@ -71,8 +68,33 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+	if desktop.pageChanged {
+		handler.BeforeRender()
+		desktop.pageChanged = false
+	}
+
+	helpers.SetStandaloneWindowStyle(window.Master())
+	handler.Render(window, desktop.walletMiddleware)
+}
+
+func (desktop *Desktop) renderWithNav(window *nucular.Window) {
```
itswisdomagain: `renderNavPage` here, to par with the alternate function `renderStandalonePage`

nuklear/desktop.go
```diff
@@ -108,7 +132,7 @@ func (desktop *Desktop) render(w *nucular.Window) {
 	// style content area
 	helpers.SetPageStyle(desktop.masterWindow)
 
-	w.LayoutSpacePushScaled(contentRect)
+	window.LayoutSpacePushScaled(contentRect)
 	if desktop.currentPage == "" { // ideally, this should only be false once in the lifetime of an instance
```
itswisdomagain: There's a comment above about always setting a value for `desktop.currentPage` when the app is launched, doing that would render this check unnecessary.

nuklear/handlers/createwallet.go
itswisdomagain: From your form validation code below, this should never happen, so we can gladly take this off. Best to keep it that way too, one error message on the password region is okay.

nuklear/helpers/style.go
```diff
@@ -15,7 +15,8 @@ import (
 )
 
 var (
-	whiteColor             = color.RGBA{0xff, 0xff, 0xff, 0xff}
+	WhiteColor             = color.RGBA{255, 255, 255, 255}
```
itswisdomagain: `WhiteColor` is only used in this package, good to keep it private `whiteColor`. Also confirm that other variables declared publicly here are actually needed outside the package. Would be good to group public variables separate from private variables using an empty line to separate them. Public variables would ideally be declared first

itswisdomagain: Also add a `colors.go` file and put all these there, especially since some of these variables are used outside this file but in this same package

itswisdomagain (changes requested):

nuklear/handlers/createwallet.go
```diff
@@ -0,0 +1,122 @@
...
+			contentWindow.Row(30).Dynamic(2)
+			contentWindow.CheckboxText("I've stored the seed in a safe and secure location", &handler.hasStoredSeed)
+			if hasStoredSeedError, ok := handler.validationErrors["hasstoredseed"]; ok {
+				contentWindow.LabelColored("("+hasStoredSeedError+")", "LC", helpers.DangerColor)
+			}
+
+			contentWindow.Row(30).Static(200)
+			if contentWindow.Button(label.T("Create Wallet"), false) {
+				if !handler.hasErrors() {
+					handler.err = wallet.CreateWallet(string(handler.passwordInput.Buffer), handler.seed)
```
itswisdomagain: - after creating the wallet, a blockchain sync should be performed (could just add a comment for now and implement it later since similar functionality is required when the nuklear app is launched)
- should also change the page if the wallet creation succeeds, it doesn't currently, user has to close and re-launch

itswisdomagain (changes requested):

nuklear/desktop.go
```diff
@@ -20,35 +18,35 @@ const (
 	contentPaneWidthPadding = 55
 )
 
+type pageHandler struct {
+	handler    Handler
+	standalone bool
```
itswisdomagain: This is a big no, why duplicating this property in 2 structs? There was a recommendation to maintain 2 handler maps without duplication, anything wrong with that approach? Will continue review after this is attended to.

nuklear/handler.go
```diff
@@ -2,52 +2,64 @@ package nuklear
...
 	"github.com/aarzilli/nucular"
-	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/app"
 	"github.com/raedahgroup/godcr/nuklear/handlers"
 )
 
 type Handler interface {
 	BeforeRender()
-	Render(*nucular.Window, walletcore.Wallet)
+	Render(*nucular.Window, app.WalletMiddleware, func(page string))
```
itswisdomagain, line 11: The changes to this method now require a different interface. All handlers except the create wallet handler do not need
- `app.WalletMiddleware`, a minimal version `walletcore.Wallet` is all that's needed
- `func(page string)` is not at all needed in other handlers

Instead of having all handlers declare parameters that they'll never use, better to have difference interfaces with the ability to provide just the dependencies required by each handler to perform its operation. Look into the cli interface to see how similar objective is achieved.

nuklear/handler.go
```diff
@@ -2,52 +2,64 @@ package nuklear
...
 }
 
-type handlersData struct {
-	name     string
-	navLabel string
-	handler  Handler
+type handlerData struct {
```
itswisdomagain: The data defined in this `handlerData` struct best describes a page. It makes sense for a page to have a `name`, a `label`, to be `standalone` or not, to have a `handler`. Other related comments brings the need for this change to life.

nuklear/handler.go
```diff
@@ -2,52 +2,64 @@ package nuklear
...
+	name       string
+	navLabel   string
+	standalone bool
+	handler    Handler
 }
 
-func getHandlers() []handlersData {
-	return []handlersData{
+func getHandlers() []handlerData {
```
itswisdomagain: This would be `getPages()` based on the previous comment

nuklear/desktop.go
```diff
@@ -73,8 +68,31 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
-func (desktop *Desktop) render(w *nucular.Window) {
-	area := w.Row(0).SpaceBegin(2)
+func (desktop *Desktop) render(window *nucular.Window) {
+	handler := desktop.handlers[desktop.currentPage]
+	if handler.standalone {
+		desktop.renderStandalonePage(window, handler.handler)
+		return
+	}
+
+	desktop.renderNavPage(window, handler.handler)
```
itswisdomagain: `handler.handler` doesn't make much sense. Here's what makes more sense.
- line 72 reads better as `page := desktop.pages[desktop.currentPage]`
- line 73 reads better as `if page.standalone`
- line 74 reads better as `desktop.renderStandalonePage(window, page.handler)`, same goes for this line 78

nuklear/helpers/style.go
```diff
@@ -37,7 +34,7 @@ var colorTable = nstyle.ColorTable{
 	ColorText:                  color.RGBA{106, 106, 106, 255},
```
itswisdomagain: These color constants should also go to the `colors.go` file

itswisdomagain (changes requested):

itswisdomagain:
Also run `go mod tidy` to update `go.mod`

itswisdomagain (approved):

PR #190, opened 2019-01-28
Introduced stimulusjs on web

### LINKED ISSUE #170: http package cleanup - remove jQuery and introduce stimulus js
The javascript code existing in the http package should be simplified by using https://github.com/stimulusjs/stimulus

jQuery should be removed as well

ademuanthony:
Resolves #170 

This PR introduces stimulus on the web and reduces the use of jquery
Jquery is not entirely removed as the frontend template uses bootstrap which depends on jquery.
This PR also uses npm for dependency management

Follow this instruction to run the app in http mode

1. `cd web`
2. `npm install`
3. `npm run build`
4. You can now run `cd ../ && ./godcr --mode=http`


.gitignore
```diff
@@ -6,6 +6,9 @@ vendor/
 godcr
 godcr.conf
 godcr.exe
+web/node_modules
+web/package-lock.json
```
itswisdomagain: You shouldn't ignore `package-lock.json`, it's important to keep track of the versions of different packages used to prevent issues that could arise when breaking changes are introduced in later versions

web/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "watch": "webpack --watch --config webpack/webpack.dev.js",
+    "start": "webpack-dev-server --open --config webpack/webpack.dev.js",
+    "analyze": "webpack --config webpack/webpack.analyze.js",
+    "build": "webpack --config webpack/webpack.prod.js"
+  },
+  "dependencies": {
+    "axios": "^0.18.0",
+    "webpack-merge": "^4.2.1"
```
itswisdomagain: `webpack-merge` should be a dev dependency - it's only used during build by the webpack config files

web/webpack/webpack.common.js
```diff
@@ -0,0 +1,30 @@
...
+    splitChunks: {
+      chunks: 'all'
+    }
+  },
+  module: {
+    
+  },
+  plugins: [
+    new CleanWebpackPlugin(['dist']),
+    new StyleLintPlugin()
```
itswisdomagain, line 49: This is not properly configured, css errors are not caught currently when `npm run build` is run

web/package.json
```diff
@@ -0,0 +1,57 @@
...
+  },
+  "devDependencies": {
```
itswisdomagain, line 17: - `@babel/plugin-syntax-dynamic-import` isn't registered as a plugin in `.babelrc`, not necessary
- `babel-polyfill` is not properly setup. See https://babeljs.io/docs/en/next/babel-polyfill.html
- `clean-webpack-plugin` doesn't have any real use, can go
- `stimulus` is not a dev dependency
- `style-loader` isn't used anymore, can go
- `stylelint` isn't properly configured, might consider taking off related deps

itswisdomagain (changes requested):
I've only reviewed the usage of babel, webpack, eslint and stylelint so far. Will review the usage of stimulus and axios next

itswisdomagain:
Should consider bundling all js files to `public/js` rather than having js files in `public/js` and `public/dist/js`
This would mean adding bootstrap and jquery to webpack bundles. [This](https://stevenwestmoreland.com/2018/01/how-to-include-bootstrap-in-your-project-with-webpack.html) may help.

web/webpack/webpack.common.js
itswisdomagain: To properly set this up, would need to pass in some options to `StyleLintPlugin()` and add a `.stylelintrc` file that `"extends": "stylelint-config-standard"`

web/package.json
```diff
@@ -11,26 +11,32 @@
     "build": "webpack --config webpack/webpack.prod.js"
   },
   "dependencies": {
+    "autoprefixer": "^9.4.7",
```
itswisdomagain: `autoprefixer` and `babel-polyfill` should be devDependencies I believe

web/webpack/webpack.common.js
```diff
@@ -1,26 +1,66 @@
...
   },
   optimization: {
     splitChunks: {
       chunks: 'all'
     }
   },
   module: {
-    
+    rules: [
+      {
```
itswisdomagain, line 18: Do we need these things at this time? [See here](https://medium.com/a-beginners-guide-for-webpack-2/webpack-loaders-css-and-sass-2cc0079b5b3a) for how style-loader work. For brevity, using style-loader, there would be no need to link to css files, instead css files would need to simply be imported or required in js code and the css file content will be injected into the html dom structure

Not sure we need this for now, we can keep linking to css files from html files and use a later PR to change this behaviour if necessary.

web/webpack/webpack.common.js
```diff
@@ -1,26 +1,66 @@
...
+            loader: 'sass-loader'
+          }
+        ]
+      }
+    ]
   },
   plugins: [
-    new CleanWebpackPlugin(['dist']),
-    new StyleLintPlugin()
+    new MiniCssExtractPlugin({
```
itswisdomagain: Also don't see a need for this

web/yarn-error.log
```diff
@@ -0,0 +1,7615 @@
+Arguments:
```
itswisdomagain: This and other log files have no real use in shared repo, should add `*.log` to gitignore.

web/package.json
```diff
@@ -11,26 +11,32 @@
...
     "axios": "^0.18.0",
-    "webpack-merge": "^4.2.1"
+    "babel-polyfill": "^6.26.0",
+    "bootstrap": "^4.2.1",
+    "jquery": "^3.3.1",
+    "popper.js": "^1.14.7",
```
itswisdomagain: Are these `bootstrap `, `jquery` and `popper.js` deps used? I still see the downloaded, minified js files for bootstrap and jquery in `/web/public/js`

If they're not use, they should be removed from this file along with references to them...

itswisdomagain (changes requested):
With regards to the dependencies, let's aim to keep things as simple as possible. We can add support for other stuffs such as sass/scss later when they become more necessary.

.gitignore
```diff
@@ -6,3 +6,12 @@ vendor/
 godcr
 godcr.conf
 godcr.exe
+web/node_modules
+web/public/dist
+
+# files generated by qt
```
itswisdomagain: These qt files need no longer be git ignored, a recent PR removed the qt interface and along with it, any qt generated files. If you still have such files in your working directory, manually delete them to ensure they don't get added to future commits.

web/.eslintrc.json
```diff
@@ -0,0 +1,7 @@
+{
+    "env": {
+        "jquery": true
```
itswisdomagain, line 3: What's this do? Useful?

web/views/layout.html
```diff
@@ -7,10 +7,14 @@
     <meta name="description" content="godcr - decred wallet">
     <meta name="author" content="The Decred developers">
     <title>GoDCR</title>
-    <link rel="stylesheet" href="/static/css/bootstrap.min.css">
-    <link rel="stylesheet" href="/static/css/style.css">
+
+    <link rel="stylesheet" href="/static/css/bootstrap.min.css?v={{timestamp}}"/>
```
itswisdomagain: `bootstrap.min.css` shouldn't need versioning, since it shouldn't be modified

web/views/layout.html
```diff
@@ -74,7 +78,7 @@ <h3>GoDCR</h3>
 
 {{ define "footer" }}
 <script>
-    $(function(){
+    window.$(function(){
```
itswisdomagain: we can revert to simply using `$`

itswisdomagain (changes requested):
Looking better, will review again after these changes

web/.eslintrc.json
ademuanthony: Yes, otherwise eslint will flag `$` as undefined

PR #191, opened 2019-01-29
Send page nuklear

yaowork:
This adds the send page to the nuklear app


nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
 package handlers
 
 import (
+	"fmt"
+	"strconv"
+	"time"
+
+	"github.com/decred/dcrd/dcrutil"
+
```
itswisdomagain: can do without this empty line

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	"github.com/raedahgroup/godcr/nuklear/helpers"
 )
 
+type utxoSelection struct {
+	selected bool
+	utxo     *walletcore.UnspentOutput
+}
+
+type input struct {
+	addressErrStr      string
```
itswisdomagain: the type already shows this is a string, the variable name need not repeat it, `addressErr` and `amountErr` is just fine

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	"github.com/raedahgroup/godcr/nuklear/handlers/widgets"
+	"github.com/raedahgroup/godcr/nuklear/helpers"
 )
 
+type utxoSelection struct {
+	selected bool
+	utxo     *walletcore.UnspentOutput
+}
+
+type input struct {
```
itswisdomagain: The data defined in this struct does not represent a generic input but rather a recipient / destination set. Variable name can be better

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	fetchUTXOError error
+	isRendering    bool
+
+	utxos           []*utxoSelection
+	isFetchingUTXOS bool
+
+	accountNumbers   []uint32
+	accountOverviews []string
+
+	inputs []*input
```
itswisdomagain: Similar to the comments above, this variable `inputs` does not represent generic inputs, it's actually an input set and the variable should be better named to show what it holds

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+			if contentWindow.ButtonText(submitButtonText) {
+				handler.validateAndSubmit(window, wallet)
+			}
+			contentWindow.End()
+		}
+		pageWindow.End()
+	}
+}
+
+// fetch accounts for select source account field
```
itswisdomagain: `// fetch accounts to display in select source account dropdown`

ademuanthony (changes requested):
I am unable to paste content from the clipboard into the address field

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	}
+
+	accountNumber := handler.accountNumbers[handler.selectedAccountIndex]
+	var requiredConfirmations int32 = walletcore.DefaultRequiredConfirmations
+	if handler.spendUnconfirmed {
+		requiredConfirmations = 0
+	}
+
+	if handler.selectCustomInputs {
+		utxos, totalInputAmount := handler.getUTXOSAndSelectedAmount()
```
ademuanthony, line 407: If a user is selecting custom input, we have to verify that the total amount selected is enough to cover the transaction amount

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+		changeDestinations := []txhelper.TransactionDestination{{
+			Amount:  dcrutil.Amount(changeAmount).ToCoin(),
+			Address: changeAddress,
+		}}
+
+		handler.successHash, handler.err = wallet.SendFromUTXOs(accountNumber, requiredConfirmations, utxos, sendDestinations, changeDestinations, passphrase)
+	} else {
+		handler.successHash, handler.err = wallet.SendFromAccount(accountNumber, requiredConfirmations, sendDestinations, passphrase)
+	}
```
ademuanthony, line 429: It will be nice to reset the form after a successful transaction

ademuanthony (changes requested):

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+	// draw page
+	if pageWindow := helpers.NewWindow("Send Page", window, 0); pageWindow != nil {
+		pageWindow.DrawHeader("Send")
+
+		// content window
+		if contentWindow := pageWindow.ContentWindow("Send Form"); contentWindow != nil {
+			helpers.SetFont(window, helpers.PageContentFont)
+
+			if handler.err != nil {
```
itswisdomagain, line 84: Have you noticed that lines 76 to 81 above are repeated across different pages? Could refactor into a function. Could modify the `helpers.NewWindow` function to do the following
- create the page window
- draw the header
- create the content window
- set the font
- return the page window and content window

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+	utxos           []*utxoSelection
+	isFetchingUTXOS bool
+
+	accountNumbers   []uint32
+	accountOverviews []string
+
+	inputs []*input
+
+	selectedAccountIndex int
```
itswisdomagain: should group related variables, `selectedAccountIndex` should be with `accountNumbers` and `accountOverviews`

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+			contentWindow.Row(25).Dynamic(1)
+			handler.selectedAccountIndex = contentWindow.ComboSimple(handler.accountOverviews, handler.selectedAccountIndex, 25)
+
+			contentWindow.Row(15).Dynamic(2)
+			if contentWindow.CheckboxText("Spend Unconfirmed", &handler.spendUnconfirmed) {
+				handler.fetchCustomInputsCheck(wallet, masterWindow)
+			}
+
+			for i := 0; i < len(handler.inputs); i++ {
```
itswisdomagain: Can simplify to `for _, destInputPair := range handler.inputs`

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+				if handler.inputs[i].addressErrStr != "" || handler.inputs[i].amountErrStr != "" {
+					contentWindow.Row(10).Dynamic(2)
+					contentWindow.LabelColored(handler.inputs[i].addressErrStr, "LC", helpers.DangerColor)
+					contentWindow.LabelColored(handler.inputs[i].amountErrStr, "LC", helpers.DangerColor)
+				}
+			}
+
+			contentWindow.Row(25).Dynamic(2)
+			if contentWindow.ButtonText("Add anohter address") {
```
itswisdomagain: `another`

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+				handler.addSendInputPair(true, window.Master())
+			}
+
+			if len(handler.inputs) > 1 {
+				if contentWindow.ButtonText("Remove last address") {
```
itswisdomagain: this can be simplified to `if len(handler.inputs) > 1 && contentWindow.ButtonText("Remove last address")`
the button will not be drawn if the first if evaluates to false

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+				handler.fetchCustomInputsCheck(wallet, masterWindow)
+			}
+
+			if handler.isFetchingUTXOS {
+				widgets.ShowIsFetching(contentWindow)
+			} else if handler.fetchUTXOError != nil {
+				contentWindow.Row(10).Dynamic(1)
+				contentWindow.LabelColored(handler.fetchUTXOError.Error(), "LC", helpers.DangerColor)
+			} else if handler.utxos != nil {
+				contentWindow.Row(20).Ratio(0.1, 0.3, 0.2, 0.2, 0.2)
```
itswisdomagain, line 145: I've also noticed this form of creating a table in at least one other page. Can this be refactored into a `drawTable` function? Perhaps taking, `columnWidthRatios []float64, rowHeader []string, rows [][]string`

itswisdomagain: This could also be something for later anyways

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+		handler.accountOverviews[index] = fmt.Sprintf("%s - Total %s (Spendable %s)", account.Name, account.Balance.Total.String(), account.Balance.Spendable.String())
+		handler.accountNumbers[index] = account.Number
+	}
+	handler.selectedAccountIndex = 0
+}
+
+// addSendInputPair adds a destinationAddress and amount input field pair on user click of
+// the 'add another address' button. This function is called at least once in the lifetime of
+// the send page
+func (handler *SendHandler) addSendInputPair(updateWindow bool, window nucular.MasterWindow) {
```
itswisdomagain: `updateWindow` should probably be the second parameter to this function

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	if updateWindow {
+		window.Changed()
+	}
+}
+
+// removeLastSendInputPair removes the last destinationAddress and amount input pair
+// when the 'remove last address' button is clicked. The button is hidden when only one
+// input pair exists on form
+func (handler *SendHandler) removeLastSendInputPair(window nucular.MasterWindow) {
+	handler.inputs = handler.inputs[:len(handler.inputs)-1]
```
itswisdomagain: should set `handler.inputs[len(handler.inputs)-1] = nil` first

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+	}
+	handler.isFetchingUTXOS = false
+	masterWindow.Changed()
+}
+
+func (handler *SendHandler) validateAndSubmit(window *nucular.Window, wallet walletcore.Wallet) {
+	isClean := true
+
+	for _, input := range handler.inputs {
+		if string(input.amount.Buffer) == "" {
```
itswisdomagain: should also check to ensure that the amount is a valid amount (> 0)

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+			input.amountErrStr = "This amount field is required"
+			isClean = false
+		} else {
+			input.amountErrStr = ""
+		}
+
+		if string(input.destinationAddress.Buffer) == "" {
```
itswisdomagain: Should also check that the address is a valid address (as is done in the cli interface)

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+			}
+
+			submitButtonText := "Submit"
+			if handler.isSubmitting {
+				submitButtonText = "Submitting"
+			}
+
+			contentWindow.Row(25).Dynamic(2)
+			if contentWindow.ButtonText(submitButtonText) {
+				handler.validateAndSubmit(window, wallet)
```
itswisdomagain: should not recall this function if `handler.isSubmitting == true`, or the function could check if `handler.isSubmitting == true` and return without performing any op

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+}
+
+func (handler *SendHandler) submit(passphrase string, window *nucular.Window, wallet walletcore.Wallet) {
+	handler.isSubmitting = true
+	window.Master().Changed()
+
+	defer window.Master().Changed()
+
+	sendDestinations := make([]txhelper.TransactionDestination, len(handler.inputs))
+	for index := range handler.inputs {
```
itswisdomagain: Is there a reason you're using `index` and not `_, input`?

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+		handler.successHash, handler.err = wallet.SendFromUTXOs(accountNumber, requiredConfirmations, utxos, sendDestinations, changeDestinations, passphrase)
+	} else {
+		handler.successHash, handler.err = wallet.SendFromAccount(accountNumber, requiredConfirmations, sendDestinations, passphrase)
+	}
+}
+
+func (handler *SendHandler) getUTXOSAndSelectedAmount() ([]string, dcrutil.Amount) {
+	var totalInputAmount dcrutil.Amount
+	var utxos []string
```
itswisdomagain: `totalInputAmount` and `utxos` could be named return variables

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+	for i := range handler.utxos {
```
itswisdomagain: `for _, utxo` instead

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+// removeLastSendInputPair removes the last destinationAddress and amount input pair
+// when the 'remove last address' button is clicked. The button is hidden when only one
+// input pair exists on form
+func (handler *SendHandler) removeLastSendInputPair(window nucular.MasterWindow) {
+	handler.inputs = handler.inputs[:len(handler.inputs)-1]
+	window.Changed()
+}
+
+// fetchCustomInputsCheck is called everytime the 'select custom inputs' checbox is checked or unchecked.
```
itswisdomagain: `checbox` => `checkbox`

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+	var requiredConfirmations int32 = walletcore.DefaultRequiredConfirmations
+	if handler.spendUnconfirmed {
+		requiredConfirmations = 0
+	}
+
+	accountNumber := handler.accountNumbers[handler.selectedAccountIndex]
+	utxos, err := wallet.UnspentOutputs(accountNumber, 0, requiredConfirmations)
+	if err != nil {
+		handler.fetchUTXOError = err
```
itswisdomagain, line 275: should also add `handler.isFetchingUTXOS = false` so that the render function shows the error message instead of showing the loading indicator

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+		return
+	}
+	window.Master().Changed()
+}
+
+func (handler *SendHandler) submit(passphrase string, window *nucular.Window, wallet walletcore.Wallet) {
+	handler.isSubmitting = true
+	window.Master().Changed()
+
+	defer window.Master().Changed()
```
itswisdomagain: should also reset `handler.isSubmitting` to false when this function returns to update the UI accordingly
```
defer func() {
	window.Master().Changed()
	handler.isSubmitting = false
}()
```

itswisdomagain (changes requested):
- Need to add sync functionality in a separate PR to make this PR work. Trying to send funds says `decred wallet backend is unreachable`
- Need to remove the [WIP] text from the Send button in the navigation window as well

nuklear/handlers/send.go
```diff
@@ -1,20 +1,365 @@
...
+
+func (handler *SendHandler) getUTXOSAndSelectedAmount() ([]string, dcrutil.Amount) {
+	var totalInputAmount dcrutil.Amount
+	var utxos []string
+
+	for i := range handler.utxos {
+		if handler.utxos[i].selected {
+			totalInputAmount += handler.utxos[i].utxo.Amount
+		}
+		utxos = append(utxos, handler.utxos[i].utxo.Address)
```
itswisdomagain: - you should only append to the slice if the utxo is selected
- you should be appending `utxo.OutputKey` not `Address`

itswisdomagain (changes requested):
Missed this

itswisdomagain:
Need to separate the UI into sections with some space between sections to look better. Everything currently looks jumbled together

screenshot 2019-01-31 at 8 50 37 am

nuklear/handlers/send.go
yaowork: i've reserved this for a later PR

nuklear/handlers/send.go
yaowork: also reserved this for a later PR

nuklear/handlers/send.go
yaowork: done

nuklear/handlers/send.go
```diff
@@ -1,20 +1,427 @@
...
+
+func (handler *SendHandler) validate(window *nucular.Window, wallet walletcore.Wallet) bool {
+	isClean := true
+
+	for _, input := range handler.sendDetailInputPairs {
+		amountStr := string(input.amount.Buffer)
+		if amountStr == "" {
+			input.amountErr = "This amount field is required"
+		} else {
+			amountInt, err := strconv.ParseFloat(amountStr, 64)
```
itswisdomagain: `amountInt`? You're parsing float...

nuklear/handlers/send.go
```diff
@@ -1,20 +1,427 @@
...
+			if len(handler.sendDetailInputPairs) > 1 && contentWindow.ButtonText("Remove last address") {
+				handler.removeLastSendInputPair(window.Master())
+			}
+
+			contentWindow.Spacing(2)
+
+			contentWindow.Row(15).Dynamic(2)
+			if contentWindow.CheckboxText("Select custom inputs", &handler.selectCustomInputs) {
+				if handler.validate(window, wallet) {
+					handler.fetchCustomInputsCheck(wallet, masterWindow)
```
itswisdomagain: No need to validate the form before displaying custom inputs, an invalid form would only affect form submission. User should be able to view and select custom inputs even if the form has invalid fields. Doesn't affect listing/selecting custom inputs

nuklear/handlers/send.go
```diff
@@ -1,20 +1,427 @@
...
+			contentWindow.Spacing(2)
+
+			submitButtonText := "Submit"
+			if handler.isSubmitting {
+				submitButtonText = "Submitting"
+			}
+
+			contentWindow.Row(25).Dynamic(2)
+			if contentWindow.ButtonText(submitButtonText) {
+				if handler.validate(window, wallet) {
```
itswisdomagain: validation should not be performed if a form submission process is already ongoing. Also, need to clear `handler.err` before validation - assuming of course, that the form is not already being submitted.

itswisdomagain (changes requested):
Still getting `Decred network is unreachable` error, send feature is not working

Also, the error message should wrap so it can be completely displayed
screenshot 2019-02-03 at 4 37 47 am

nuklear/handlers/send.go
yaowork: validation rules concerning this have been added

nuklear/handlers/send.go
yaowork: done

itswisdomagain:
Will revisit this PR after #207 is merged. @codemaestro64, you'll need to rebase, may wait till #207 is merged

nuklear/handler.go
```diff
@@ -54,6 +49,11 @@ func getNavPages() []navPage {
 			label:   "Stake Info",
 			handler: &handlers.StakeInfoHandler{},
 		},
+		{
```
itswisdomagain: The nav order should be maintained. Balance, receive, send... Compare the http interface

terminal/terminal.go
```diff
@@ -84,8 +84,10 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
 		AddItem("Stakeinfo", "", 'k', func() {
 			changePageColumn(pages.StakeinfoPage(walletMiddleware, setFocus, clearFocus))
 		}).
-		AddItem("Purchase Tickets", "", 't', nil).
-		AddItem("Quit", "", 'q', func() {
+		AddItem("Purchase Tickets", "", 't', func() {
```
itswisdomagain: Why does this PR include these changes?

test/config_test.go
```diff
@@ -0,0 +1,15 @@
+package test
```
itswisdomagain: This file belongs in a different PR, please remove

web/routes/walletloader.go
```diff
@@ -54,6 +54,7 @@ func (routes *Routes) walletLoaderFn(next http.Handler) http.Handler {
 
 		// wallet is open, check if blockchain is synced
 		blockchainSyncStatus := routes.blockchain.status()
+		blockchainSyncStatus = walletcore.SyncStatusSuccess
```
itswisdomagain: This was meant to be a temporary hack I believe. Please remove

itswisdomagain (changes requested):

terminal/terminal.go
yaowork: Did a rebase that caused conflicts.

itswisdomagain (changes requested):
After successful sending, only 1 set of destination address fields should be left on the page; and the utxo list should be cleared

screenshot 2019-03-05 at 1 49 29 pm

ademuanthony (approved):

nuklear/handlers/send.go
```diff
@@ -1,21 +1,455 @@
...
+	return
+}
+
+func (handler *SendHandler) resetForm() {
+	if len(handler.accountNumbers) > 1 {
+		handler.selectedAccountIndex = 0
+	}
+
+	handler.sendDetailInputPairs = handler.sendDetailInputPairs[1 : len(handler.sendDetailInputPairs)-1]
+	handler.sendDetailInputPairs[0].amount.Buffer = []rune("")
```
itswisdomagain, line 454: This throws `panic: runtime error: index out of range`

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #192, opened 2019-01-29
add purchase tickets page

yaowork:
This adds the purchase tickets page to the nuklear app interface


ademuanthony (changes requested):
Neat job.

I think it will make sense if the form is reset after a successful transaction

ademuanthony (approved):

PR #193, opened 2019-01-29
Writing output/error log to log file using slog

### LINKED ISSUE #84: store dcrcli logs in log file
This (https://github.com/decred/slog) should be used for logging throughout the dcrcli app to achieve the objective described in [this comment](https://github.com/raedahgroup/dcrcli/pull/64#issuecomment-450188278):

> I notice we're simply logging errors to the console. I suggest we log them to a file as well, and let the users know about the file. That way, the history of errors does not get lost, and we can get easily introduce telemetry if we want to. The main benefit is being able to go look at detailed error output (including stack traces if necessary) in the log file, and knowing where to start looking for problems. If a user reports an error, we can request the log file for richer context.

Sirmorrison:
Fixes #84 

Implemented output logging to file in appdata directory using slog library.


app/config/config.go
```diff
@@ -38,6 +42,9 @@ type ConfFileOptions struct {
 	NoWalletRPCTLS  bool   `long:"nowalletrpctls" description:"Disable TLS when connecting to dcrwallet daemon via RPC"`
 	HTTPHost        string `long:"httphost" description:"HTTP server host address or IP"`
 	HTTPPort        string `long:"httpport" description:"HTTP server port"`
+	DebugLevel      string `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
```
ademuanthony: The `-d` here conflicts with the `-d` in `./godcr balance -d` where the `-d` flag is the short form of `--detailed`
Maybe we should keep only the long version `--debuglevel` here

cli/cli.go
```diff
@@ -47,7 +47,7 @@ func Run(ctx context.Context, walletMiddleware app.WalletMiddleware, appConfig c
 	} else if noCommandPassed {
 		listCommands()
 	} else if err != nil {
-		fmt.Println(err)
```
ademuanthony, line 50: We shouldn't depend on log output for communicating critical information to the user. The output here gives suggestions to the user on what to do. 
If the user set `debug=critical` and type in a non-existing command
```./godcr nothing --debuglevel=critical```

File: cli/commands/createaccount.go [hunk 3]
```diff
@@ -25,7 +25,7 @@ func (c CreateAccountCommand) Run(ctx context.Context, wallet walletcore.Wallet)
 	if err != nil {
 		return err
 	}
-
+	log.Infof("Account created successfully")
```
ademuanthony: this will print both the log output and the app notification below.
see comment below for recommendation

cli/commands/send.go
```diff
@@ -155,10 +155,12 @@ func completeNormalSend(wallet walletcore.Wallet, sourceAccount uint32, sendDest
 	}
 
 	if len(sendDestinations) == 1 {
+		log.Warnf("You are about to send %f DCR to %s", sendDestinations[0].Amount, sendDestinations[0].Address)
```
ademuanthony: Duplication notification

log.go
```diff
@@ -0,0 +1,203 @@
...
+	"github.com/raedahgroup/godcr/nuklear"
+	"github.com/raedahgroup/godcr/qt"
+	"github.com/raedahgroup/godcr/web"
+	"github.com/decred/slog"
+	"github.com/jrick/logrotate/rotator"
+)
+
+// logWriter implements an io.Writer that outputs to both standard output and
+// the write-end pipe of an initialized log rotator.
+type logWriter struct{}
```
ademuanthony, line 25: We can add a field to indicate whether or not this writer writes to `Stdout`

log.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (logWriter) Write(p []byte) (n int, err error) {
+	os.Stdout.Write(p)
```
ademuanthony: writing to `Stdout` should be conditional as recommended in the comment above

log.go
```diff
@@ -0,0 +1,203 @@
...
+
+	// logRotator is one of the logging outputs.  It should be closed on
+	// application shutdown.
+	logRotator *rotator.Rotator
+
+	log          = backendLog.Logger("GODC")
+	webLog    	 = backendLog.Logger("WEB")
+	nuklearLog   = backendLog.Logger("NUK")
+	qtLog    	 = backendLog.Logger("QTL")
+	cliLog    	 = backendLog.Logger("CLI")
```
ademuanthony, line 52: `cliLog` should be configured not to write to `Stdout`

ademuanthony (changes requested):

log.go
```diff
@@ -27,7 +26,6 @@ import (
 type logWriter struct{}
 
 func (logWriter) Write(p []byte) (n int, err error) {
-	os.Stdout.Write(p)
 	logRotator.Write(p)
```
ademuanthony: You shouldn't ignore the returned value here
instead `return logRotator.Write(p)`

cli/commands/createaccount.go
```diff
@@ -25,7 +26,6 @@ func (c CreateAccountCommand) Run(ctx context.Context, wallet walletcore.Wallet)
 	if err != nil {
 		return err
 	}
-	log.Infof("Account created successfully")
-	fmt.Println("Account created successfully")
+	clilog.LogInfo("Account created successfully")
```
ademuanthony, line 28: printing out this information should be explicit.
I thinking it will be better if logging activities can just handle writing to file for `cli` and `fmt.Print*` for the app interaction
We should still maintain `fmt.Println("Account created successfully")`

cli/clilog/helper.go
```diff
@@ -0,0 +1,23 @@
...
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package clilog
+
+import "fmt"
+
+func LogInfo(message string) {
+ log.Info(message)
+ fmt.Println(message)
```
ademuanthony, line 12: The `clilog` should not write to the standard output

cli/clilog/helper.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func LogWarn(message string) {
+ log.Warn(message)
+ fmt.Println(message)
```
ademuanthony, line 17: The `clilog` should not write to the standard output

cli/clilog/helper.go
```diff
@@ -0,0 +1,23 @@
...
+}
+
+func LogError(message error) {
+ log.Error(message)
+ fmt.Println(message)
```
ademuanthony, line 22: The `clilog` should not write to the standard output

cli/cli.go
```diff
@@ -47,7 +48,7 @@ func Run(ctx context.Context, walletMiddleware app.WalletMiddleware, appConfig c
 	} else if noCommandPassed {
 		listCommands()
 	} else if err != nil {
-		log.Error(err)
```
ademuanthony: The user should be interactively  notified of this error

ademuanthony (changes requested):

ademuanthony (approved):

PR #194, opened 2019-01-30
add detailed and simple views to balance page on nuklear

### LINKED ISSUE #181: nuklear balance page improvements
As in other interfaces, the nuklear balance page should show a simple balance by default but also provide a means for a user to view the detailed balance report.

yaowork:
This adds the option of displaying simple and detailed balance to the balance page of the nuklear app interface. 

Fixes #181


nuklear/handlers/balance.go
```diff
@@ -10,12 +12,15 @@ type BalanceHandler struct {
...
 	accounts    []*walletcore.Account
+	detailed    bool
 }
 
 func (handler *BalanceHandler) BeforeRender() {
 	handler.err = nil
 	handler.accounts = nil
 	handler.isRendering = false
+	handler.detailed = false
+
```
ademuanthony: empty space should be remove

ademuanthony (changes requested):

nuklear/handlers/balance.go
```diff
@@ -33,28 +38,67 @@ func (handler *BalanceHandler) Render(w *nucular.Window, wallet walletcore.Walle
...
+					handler.showDetailedView(content.Window)
 				}
 			}
 			content.End()
 		}
 		page.End()
 	}
 }
+
+func (handler *BalanceHandler) showDetailedView(window *nucular.Window) {
```
itswisdomagain: `showSimpleView` should come before this method

nuklear/handlers/balance.go
```diff
@@ -33,28 +38,67 @@ func (handler *BalanceHandler) Render(w *nucular.Window, wallet walletcore.Walle
...
+					detailsCheckboxText = "Hide details"
+				}
+
+				content.Row(20).Dynamic(1)
+				if content.CheckboxText(detailsCheckboxText, &handler.detailed) {
+					content.Master().Changed()
+				}
+
+				if !handler.detailed && len(handler.accounts) == 1 {
+					handler.showSimpleView(content.Window)
```
itswisdomagain, line 49: Don't understand what's going on here. Please compare the http interface for what's expected. 1 anomaly I've noticed is mentioned in the next comment...

nuklear/handlers/balance.go
```diff
@@ -33,28 +38,67 @@ func (handler *BalanceHandler) Render(w *nucular.Window, wallet walletcore.Walle
...
+
+	// rows
+	helpers.SetFont(window, helpers.PageContentFont)
+	for _, account := range handler.accounts {
+		window.Row(20).Ratio(0.16, 0.14, 0.15, 0.15, 0.2, 0.20)
+		window.Label(account.Name, "LC")
+		window.Label(handler.simpleBalance(account.Balance, handler.detailed), "LC")
+
+		if handler.detailed {
+			window.Label(account.Balance.Spendable.String(), "LC")
```
itswisdomagain, line 87: `if handler.detailed == true`, line 81 would be showing the wrong value for the `Balance` column. Detailed view should not combine the total balance and spendable balance in the `Balance` column

itswisdomagain (changes requested):
After building, I'm unable to run the desktop app with `godcr --mode=nuklear`. It just prints a "launching" message on my terminal and exits without even showing the window. Anyone else having this issue?

Also, preliminary review below shows that this code does not yet produce the expected behaviour. Will test more when that is fixed and I'm able to run the app in nuklear mode.

app/shared.go
```diff
@@ -0,0 +1,15 @@
+package app
+
+import (
+	"fmt"
+
+	"github.com/raedahgroup/godcr/app/walletcore"
+)
+
+func SimpleBalance(balance *walletcore.Balance, detailed bool) string {
```
itswisdomagain: There is an existing file where this function could be better placed: `app/walletcore/helper.go`

nuklear/handlers/balance.go
```diff
@@ -33,28 +36,59 @@ func (handler *BalanceHandler) Render(w *nucular.Window, wallet walletcore.Walle
...
+
+				content.Row(20).Dynamic(1)
+				if content.CheckboxText(detailsCheckboxText, &handler.detailed) {
+					content.Master().Changed()
+				}
+
+				if !handler.detailed && len(handler.accounts) == 1 {
+					handler.showSimpleView(content.Window)
+				} else {
+					handler.showDetailedView(content.Window)
```
itswisdomagain: This function does not only cover scenarios where `handler.detailed == true`, it also covers scenario where the `handler.detailed == false`. Better to rename this to `showTabularView` or something that more closely represent what the function actually does

itswisdomagain (changes requested):
The second column here is trimmed off, what can be done about it?
screenshot 2019-02-04 at 9 41 08 pm

itswisdomagain (approved):

ademuanthony (approved):

PR #200, opened 2019-02-06
implement the terminal UI history page

Sirmorrison:
made the terminal history demo page to into a full functional history page by fetching the wallet history from wallet core

image


terminal/pages/history.go
```diff
@@ -1,16 +1,40 @@
...
 
-func HistoryPage() tview.Primitive {
+func HistoryPage(wallet walletcore.Wallet) tview.Primitive {
 	body := tview.NewTable().SetBorders(true)
-	body.SetCell(0, 0, tview.NewTableCell("Date").SetAlign(tview.AlignCenter))
+	transactions, err := wallet.TransactionHistory()
+	if err != nil {
+		panic(err) 
+	}
+		color := tcell.ColorYellow
```
itswisdomagain: `go fmt ./...` please

terminal/terminal.go
```diff
@@ -6,15 +6,22 @@ import (
 	"strings"
 
 	"github.com/raedahgroup/godcr/app"
-	"github.com/raedahgroup/godcr/cli/walletloader"
+		"github.com/raedahgroup/godcr/app/walletcore"
+
+	// "github.com/raedahgroup/godcr/cli/walletloader"
```
itswisdomagain: commented code should not make their way into shared repo, also don't see the need for the empty line space in line 10 above

terminal/terminal.go
```diff
@@ -25,17 +32,20 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 		})
 	list.SetBorder(true).SetTitle(fmt.Sprintf("%s Terminal", strings.ToUpper(app.Name)))
 
-	err := syncBlockChain(ctx, walletMiddleware)
-	if err != nil {
-		fmt.Println(err)
-	}
+	// err := syncBlockChain(ctx, walletMiddleware)
```
itswisdomagain: commented code should not be committed into repo
also, while it may be necessary to remove this code for testing, it's important to retain it in the project source code
what should ideally be done about this is to stop displaying sync progress as messages on the terminal and instead use a progress bar on the terminal UI app, but that's for a later PR

terminal/terminal.go
```diff
@@ -59,7 +69,7 @@ func terminalLayout(tviewApp *tview.Application) tview.Primitive {
 			changePageColumn(pages.SendPage(setFocus, clearFocus))
 		}).
 		AddItem("History", "", 'h', func() {
-			changePageColumn(pages.HistoryPage())
+			changePageColumn(pages.HistoryPage(terminal.wallet))
```
itswisdomagain: You could just pass `walletMiddleware` here directly, don't see no need for the `Terminal` struct type or the `terminal` variable declared in line 46.

terminal/terminal.go
```diff
@@ -6,15 +6,22 @@ import (
...
 	"github.com/raedahgroup/godcr/terminal/pages"
 	"github.com/rivo/tview"
 )
 
+type Terminal struct {
+	wallet   walletcore.Wallet
+}
+
 func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
+
```
itswisdomagain: this space is unnecessary

terminal/pages/history.go
itswisdomagain: Getting an `undefined: tcell` error from this line, PR is not building

terminal/pages/history.go
```diff
@@ -1,16 +1,40 @@
...
+	body.SetCell(0, 0, tview.NewTableCell("Date").SetAlign(tview.AlignCenter).SetBackgroundColor(color))
```
itswisdomagain: Not sure what this background color does, can't notice it in the screenshot attached to this PR

terminal/pages/history.go
```diff
@@ -1,16 +1,40 @@
...
 	body.SetCell(0, 1, tview.NewTableCell("Amount").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 2, tview.NewTableCell("Fee").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 3, tview.NewTableCell("Direction").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 4, tview.NewTableCell("Type").SetAlign(tview.AlignCenter))
+	body.SetCell(0, 5, tview.NewTableCell("Hash").SetAlign(tview.AlignCenter))
 
+	for i , tx := range transactions {
+		i = i + 1
```
itswisdomagain: no, no, no,
don't re-assign (change the value of) a loop variable, it automatically increases and that behaviour should be maintained
Instead create another variable to use as appropriate: `row := i + 1`

terminal/pages/history.go
```diff
@@ -1,16 +1,40 @@
...
+		body.SetCell(i, 3, tview.NewTableCell(tx.Direction.String()).SetAlign(tview.AlignCenter))
+		body.SetCell(i, 4, tview.NewTableCell(tx.Type).SetAlign(tview.AlignCenter))
+		body.SetCell(i, 5, tview.NewTableCell(tx.Hash).SetAlign(tview.AlignCenter))
+	}
 	return body
+}
+
+func AmountToString(amount float64) string {
+	amount = math.Round(amount)
+	return fmt.Sprintf("%d DCR", int(amount))
```
itswisdomagain: Why rounding the amount to remove the decimal part? Space?
Assuming this is necessary, there is another function of this nature in the nuklear interface, to avoid code repetition, put this function in `app/walletcore/helpers.go` and have both this interface and the nuklear interface call the function from there.

itswisdomagain (changes requested):

main.go
```diff
@@ -49,15 +49,15 @@ func main() {
...
+	// Special show command to list supported subsystems and exit.
 	if appConfig.DebugLevel == "show" {
 		fmt.Println("Supported subsystems", supportedSubsystems())
 		os.Exit(0)
 	}
 
 	// Parse, validate, and set debug log level(s).
 	if err := parseAndSetDebugLevels(appConfig.DebugLevel); err != nil {
-		err :=fmt.Errorf("%s: %v", "loadConfig", err.Error())
+		err := fmt.Errorf("%s: %v", "loadConfig", err.Error())
```
itswisdomagain: not part of this PR, but this code could be better written: `fmt.Errorf("loadConfig: %s", err.Error())`
- `err.Error()` is of type string
- string formatting should be for variables not string literals

terminal/pages/history.go
```diff
@@ -1,40 +1,33 @@
...
-	"math"
-
-	"github.com/rivo/tview"
+	"github.com/raedahgroup/godcr/app"
 	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/rivo/tview"
 )
 
-func HistoryPage(wallet walletcore.Wallet) tview.Primitive {
+func HistoryPage(walletMiddleware app.WalletMiddleware) tview.Primitive {
```
itswisdomagain: this can be, and should be `wallet walletcore.Wallet`, you can still call `HistoryPage` with `app.WalletMiddleware ` because `app.WalletMiddleware` is a super type of `walletcore.Wallet`

web/server.go
```diff
@@ -55,7 +55,7 @@ func StartServer(ctx context.Context, walletMiddleware app.WalletMiddleware, hos
 	<-ctx.Done()
 	// log.Info("web server stopped")
 	// fmt.Println("Web server stopped")
```
itswisdomagain: Commented code in repo? No please

itswisdomagain (changes requested):

terminal/terminal.go
```diff
@@ -24,16 +23,10 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 			tviewApp.Stop()
 		})
 	list.SetBorder(true).SetTitle(fmt.Sprintf("%s Terminal", strings.ToUpper(app.Name)))
-
-	err := syncBlockChain(ctx, walletMiddleware)
```
itswisdomagain: Restore this sync functionality. The `syncBlockChain` also opens the wallet without which trying to use the wallet to perform operations like load history causes the app to panic with the following error: `panic: runtime error: invalid memory address or nil pointer dereference`

itswisdomagain (changes requested):

terminal/pages/history.go
```diff
@@ -1,16 +1,32 @@
...
+	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/rivo/tview"
 )
 
-func HistoryPage() tview.Primitive {
+func HistoryPage(wallet walletcore.Wallet) tview.Primitive {
 	body := tview.NewTable().SetBorders(true)
+	transactions, err := wallet.TransactionHistory()
+	if err != nil {
+		panic(err)
```
itswisdomagain: This is a big no, `panic` causes the app to fail ungracefully. Should instead use a view to display the error message and return

terminal/terminal.go
```diff
@@ -23,8 +23,8 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 		AddItem("Exit", "", 'q', func() {
 			tviewApp.Stop()
 		})
-	list.SetBorder(true).SetTitle(fmt.Sprintf("%s Terminal", strings.ToUpper(app.Name)))
 
+	list.SetBorder(true).SetTitle(fmt.Sprintf("%s Terminal", strings.ToUpper(app.Name)))
```
itswisdomagain: There should be a space between this line 27 and the next line 28, to show separation of concerns

itswisdomagain (changes requested):
In addition to the changes below, also make the history table scrollable. That might mean, focusing the table instead of the nav menu, which would also require a way to move focus back to the menu in case the user wants to select another menu item. Compare how similar stuff is done in `SendPage`

terminal/pages/history.go
```diff
@@ -1,16 +1,34 @@
...
+	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/rivo/tview"
 )
 
-func HistoryPage() tview.Primitive {
+func HistoryPage(wallet walletcore.Wallet) tview.Primitive {
 	body := tview.NewTable().SetBorders(true)
+	transactions, err := wallet.TransactionHistory()
+	if err != nil {
+		fmt.Println(err)
```
itswisdomagain: - seeing as this is a GUI mode, errors should not simply be printed to terminal, they should be displayed on the GUI page
- should return if there's an error, there's no need to show table headers when an error occurs, just show the error message

Here's something that could work (with some tweaking if possible):
```
if err != nil {
    return tview.NewTextView().SetTextAlign(tview.AlignCenter).SetText(err.Error())
}
```

Going with this approach, the declaration `body := tview.NewTable().SetBorders(true)` should be moved to after the if statement, since it will never be use before that point

itswisdomagain (dismissed):
There was a previous comment about making the history table scrollable. Is that done yet? I'll test after this last change mentioned below is attended to

ademuanthony:
You need to `go fmt ./...`

terminal/pages/history.go
```diff
@@ -1,16 +1,43 @@
...
+		}
+	})
 	body.SetCell(0, 0, tview.NewTableCell("Date").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 1, tview.NewTableCell("Amount").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 2, tview.NewTableCell("Fee").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 3, tview.NewTableCell("Direction").SetAlign(tview.AlignCenter))
 	body.SetCell(0, 4, tview.NewTableCell("Type").SetAlign(tview.AlignCenter))
+	body.SetCell(0, 5, tview.NewTableCell("Hash").SetAlign(tview.AlignCenter))
+
+	for i, tx := range transactions {
```
ademuanthony, line 30: Can we have the history paginated?
This result will keep building up and it will become a lot to handle at one call

ademuanthony (changes requested):

nuklear/handlers/transactions.go
```diff
@@ -50,8 +50,8 @@ func (handler *TransactionsHandler) Render(window *nucular.Window, wallet wallet
 						helpers.SetFont(window, helpers.PageContentFont)
 						contentWindow.Row(20).Ratio(0.18, 0.08, 0.1, 0.15, 0.15, 0.7)
 						contentWindow.Label(tx.FormattedTime, "LC")
-						contentWindow.Label(helpers.AmountToString(tx.Amount.ToCoin()), "LC")
-						contentWindow.Label(helpers.AmountToString(tx.Fee.ToCoin()), "LC")
+						contentWindow.Label(walletcore.AmountToString(tx.Amount.ToCoin()), "LC")
```
ademuanthony: Use `tx.Amount.String()` instead

nuklear/handlers/transactions.go
```diff
@@ -50,8 +50,8 @@ func (handler *TransactionsHandler) Render(window *nucular.Window, wallet wallet
...
+						contentWindow.Label(walletcore.AmountToString(tx.Fee.ToCoin()), "LC")
```
ademuanthony: Use `tx.Amount.String()` instead

app/walletcore/helper.go
```diff
@@ -34,3 +35,8 @@ func SimpleBalance(balance *Balance, detailed bool) string {
 		return fmt.Sprintf("Total %s (Spendable %s)", balance.Total.String(), balance.Spendable.String())
 	}
 }
+
+func AmountToString(amount float64) string {
```
ademuanthony: You can delete this function as it's same as `amount.String()`
update the consumers both and `terminal/history.go` and in `nuklear/handlers/transactions.go` line 53 and 54

ademuanthony (changes requested):

app/walletcore/helper.go
```diff
@@ -35,8 +35,3 @@ func SimpleBalance(balance *Balance, detailed bool) string {
 		return fmt.Sprintf("Total %s (Spendable %s)", balance.Total.String(), balance.Spendable.String())
 	}
 }
-
```
ademuanthony: app\walletcore\helper.go:6:2: imported and not used: "math"
Please also test your code locally before giving out for review

ademuanthony (changes requested):

ademuanthony (approved):

PR #203, opened 2019-02-11
implemented terminal stake info demo page to live

Sirmorrison:
image

### DIFF
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -80,7 +80,7 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
 			changePageColumn(pages.HistoryPage())
 		}).
 		AddItem("Stakeinfo", "", 'k', func() {
-			changePageColumn(pages.ReceivePage())
+			changePageColumn(pages.StakeinfoPage(walletMiddleware, setFocus, clearFocus))
 		}).
 		AddItem("Purchase Tickets", "", 't', func() {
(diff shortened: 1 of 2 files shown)


ademuanthony (changes requested):
I am experiencing a panic whenever I try to navigate to the `Stake info` page.
I am using decrlibwallet, the `walletrpcserver` is not set in the config

Sirmorrison:
> 
> 
> I am experiencing a panic whenever I try to navigate to the `Stake info` page.
> I am using decrlibwallet, the `walletrpcserver` is not set in the config

terminal/terminal.go
```diff
@@ -81,7 +80,7 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
 			changePageColumn(pages.HistoryPage())
 		}).
 		AddItem("Stakeinfo", "", 'k', func() {
-			changePageColumn(pages.ReceivePage())
+			changePageColumn(pages.StakeinfoPage(walletMiddleware, setFocus, clearFocus))
```
ademuanthony, line 83: We need a way to return the focus back to the main menu from this page. I have to restart the application each time I want to go to the main menu

ademuanthony: I see that the Esc key works as we discussed.
But we will need to have a help-tip at the footer of every page that explains the control keys

ademuanthony (approved):

PR #205, opened 2019-02-12
added walletloader to terminal package to prevent panic

### LINKED ISSUE #179: error/crash when running godcr if wallet doesn't exist
### Cli
Running a godcr command e.g. `godcr stakeinfo` without having a wallet, prompts user to create a wallet. If user choses not to create a wallet at the moment, godcr crashes.
screenshot 2019-01-25 at 2 04 57 pm

### Nuklear
The UI launch is canceled and the following error is printed after the app quits.`fatal error: unexpected signal during runtime execution`

### qt
UI launch is canceled without any message.

### terminal
If wallet doesn't exist, user is prompted to create wallet. If user does not accept to create wallet, the terminal UI is displayed even though the features in it cannot be used without a wallet. When the app is quit, the following message is displayed:
```
Blockchain sync failed to start. wallet_not_loaded
wallet_not_loaded
```

If user accepts to create a wallet, wallet is created. After wallet creation, user is asked if to sync blockchain. **Error is** blockchain sync happens whether user agrees to sync or not.

Sirmorrison:
fixes #179 for terminal

> I am experiencing a panic whenever I try to navigate to the `Stake info` page.
> I am using decrlibwallet, the `walletrpcserver` is not set in the config

implemented the terminal create wallet page and also added walletloader that fixes panic when using dcrlibwallet

### DIFF
--- a/terminal/terminal.go
+++ b/terminal/terminal.go
@@ -21,24 +21,24 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 	}
 
 	if walletExists {
-	// `Run` blocks until app.Stop() is called before returning
+		// `Run` blocks until app.Stop() is called before returning
 		return tviewApp.SetRoot(layout, true).SetFocus(layout).Run()
 	}
 
 	var password, confPassword string
 	form := tview.NewForm().
-		AddPasswordField("Password", "", 10, '*', func(text string){
+		AddPasswordField("Password", "", 10, '*', func(text string) {
 			if len(text) == 0 {
 				fmt.Println("password cannot be less than 4")
 				return
 			}
 			password = text
 		}).
-		AddPasswordField("Password", "", 10, '*', func(text string){
+		AddPasswordField("Password", "", 10, '*', func(text string) {
 			confPassword = text
 		}).
 		AddButton("Create", func() {
-			if password != confPassword{
+			if password != confPassword {
 				fmt.Println("password does not match")
 				return
 			}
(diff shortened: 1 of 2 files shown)


ademuanthony (changes requested):
Just a few changes

terminal/terminal.go
```diff
@@ -6,24 +6,54 @@ import (
...
+				}
+				CreateWallet(ctx, password, walletMiddleware)
+			}).
+			AddButton("Quit", func() {
+				tviewApp.Stop()
+			})
+		form.SetBorder(true).SetTitle("Enter some data").SetTitleAlign(tview.AlignCenter).SetRect(30,10,40,10)
+		return tviewApp.SetRoot(form, false).Run()
+	
+	}else{
```
ademuanthony: you can invert the condition to do away with the else
```
 if walletExists {
    return tviewApp.SetRoot(layout, true).SetFocus(layout).Run()
}
````

File: terminal/walletloader.go [hunk 2]
```diff
@@ -0,0 +1,83 @@
...
+			fmt.Fprintln(os.Stderr, err.Error())
+		}
+		return
+
+	case <-ctx.Done():
+		return false, ctx.Err()
+	}
+}
+
+func CreateWallet(ctx context.Context, password string, walletMiddleware app.WalletMiddleware) (err error) {
```
ademuanthony, line 52: the `ctx` param is not used, we can do away with it

terminal/walletloader.go
```diff
@@ -0,0 +1,83 @@
...
+	// first check if wallet already exists
+	//work in progress
```
ademuanthony, line 54: what is this for, I think it can go

Sirmorrison: this is a future implementation for creating wallet will create a different pr to complete the implementation is only half implemented

Sirmorrison: ctx will be used for blockchain sync function which will be implemented in future

ademuanthony (approved):

PR #207, opened 2019-02-14
add sync feature

yaowork:
This adds sync blockchain feature/page to the nuklear app. It runs automatically on running godcr in nuklear mode

### DIFF
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -2,12 +2,14 @@ package nuklear
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/aarzilli/nucular"
 	"github.com/aarzilli/nucular/label"
 	"github.com/aarzilli/nucular/rect"
 	"github.com/raedahgroup/godcr/app"
 	"github.com/raedahgroup/godcr/nuklear/helpers"
+	"github.com/raedahgroup/godcr/nuklear/nuklog"
 )
 
 const (
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -29,6 +29,7 @@ type Desktop struct {
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
 	desktop := &Desktop{
 		walletMiddleware: walletMiddleware,
+		pageChanged:      true,
 		currentPage:      "sync",
 	}
 
--- a/nuklear/desktop.go
+++ b/nuklear/desktop.go
@@ -11,19 +11,21 @@ import (
 )
 
 const (
-	navWidth = 260
-	homePage = "balance"
+	navWidth  = 260
+	homePage  = "balance"
+	startPage = "sync"
 
 	contentPaneXOffset      = 45
 	contentPaneWidthPadding = 55
 )
 
 type Desktop struct {
-	masterWindow     nucular.MasterWindow
-	walletMiddleware app.WalletMiddleware
-	currentPage      string
-	pageChanged      bool
-	pages            map[string]page
+	masterWindow    nucular.MasterWindow
+	wallet          app.WalletMiddleware
+	currentPage     string
+	pageChanged     bool
+	navPages        map[string]NavPageHandler
+	standalonePages map[string]StandalonePageHandler
 }
 
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
@@ -33,11 +35,29 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 		currentPage:      homePage,
 	}
 
-	// register pages
-	pages := getPages()
-	desktop.pages = make(map[string]page, len(pages))
-	for _, page := range pages {
-		desktop.pages[page.name] = page
+	// initialize master window and set style
+	window := nucular.NewMasterWindow(nucular.WindowNoScrollbar, app.Name, desktop.render)
+	window.SetStyle(helpers.GetStyle())
+	desktop.masterWindow = window
+
+	// initialize fonts for later use
+	err := helpers.InitFonts()
+	if err != nil {
+		return nil
+	}
+
+	// register nav page handlers
+	navPages := getNavPagesData()
+	desktop.navPages = make(map[string]NavPageHandler, len(navPages))
+	for _, page := range navPages {
+		desktop.navPages[page.name] = page.handler
+	}
+
+	// register standalone page handlers
+	standalonePages := getStandalonePageData()
+	desktop.standalonePages = make(map[string]StandalonePageHandler, len(standalonePages))
+	for _, page := range standalonePages {
+		desktop.standalonePages[page.name] = page.handler
 	}
 
 	// open wallet and start blockchain syncing in background
@@ -61,7 +81,7 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 		return err
 	}
 
-	// todo run sync and show progress
+	desktop.currentPage = startPage
 
 	// draw window
 	desktop.masterWindow.Main()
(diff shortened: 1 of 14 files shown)


nuklear/handlers/sync.go
```diff
@@ -0,0 +1,115 @@
...
+
+			if s.isShowingPercentageProgress {
+				contentWindow.Row(30).Dynamic(1)
+				contentWindow.Progress(&s.percentageProgress, 100, false)
+			}
+
+			contentWindow.End()
+		}
+	}
+
```
Sirmorrison: this space is not needed

Sirmorrison (changes requested):

nuklear/desktop.go
```diff
@@ -66,15 +74,25 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 		return err
```
Sirmorrison: remove the commented line of code on line 72

Sirmorrison (changes requested):

nuklear/desktop.go
itswisdomagain: Good catch @Sirmorrison. @codemaestro64, I understand that line wasn't introduced in this PR but shouldn't even be in master ideally. Please delete it...

nuklear/desktop.go
```diff
@@ -8,25 +8,26 @@ import (
...
 	"github.com/raedahgroup/godcr/nuklear/helpers"
 	"github.com/raedahgroup/godcr/nuklear/nuklog"
 )
 
 const (
-	navWidth = 260
-	homePage = "balance"
+	navWidth  = 260
+	homePage  = "balance"
+	startPage = "sync"
```
itswisdomagain: Can do without these `homePage` and `startPage` constants. `homePage` is only  referenced in line 133 **if desktop.currentPage == ""**, with  the  introduction  of the sync page, there should no longer be a case where currentPage is empty. Thus, the  if statement  can go.

`startPage` is only used in line 77. That line can be deleted and the declaration of `desktop` in line 34  modified to include the `currentPage: "sync"` property assignment.

nuklear/desktop.go
```diff
@@ -61,20 +69,29 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
 	if !walletExists {
 		// todo add ui to create wallet
 		err = fmt.Errorf("No wallet found. Use 'godcr create' to create a wallet before launching the desktop app")
-		// fmt.Println(err.Error())
 		nuklog.LogInfo(err.Error())
 		return err
 	}
 
-	// todo run sync and show progress
+	desktop.currentPage = startPage
```
itswisdomagain: Add `currentPage: "sync"`  to line 35 and delete this line and the `startPage` constant declaration

nuklear/desktop.go
```diff
@@ -61,20 +69,29 @@ func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error
...
+	if handler, ok := desktop.standalonePages[desktop.currentPage]; ok {
+		desktop.renderStandalonePage(window, handler)
+		return
+	}
+
+	handler := desktop.navPages[desktop.currentPage]
+	desktop.renderNavPage(window, handler)
+}
+
+func (desktop *Desktop) renderNavPage(window *nucular.Window, handler NavPageHandler) {
```
itswisdomagain: `renderStandalonePage` should come before this function

nuklear/handler.go
```diff
@@ -2,52 +2,72 @@ package nuklear
...
 	Render(*nucular.Window, walletcore.Wallet)
 }
 
-type handlersData struct {
-	name     string
-	navLabel string
-	handler  Handler
+type StandalonePageHandler interface {
+	BeforeRender()
+	Render(*nucular.Window, app.WalletMiddleware, func(string))
```
itswisdomagain, line 16: Introduce an alias for this function so it's expected usage is immediately obvious:
`type  pageChangeFunc func(string)`

nuklear/handlers/sync.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+func (s *SyncHandler) BeforeRender() {
+	s.isRendering = false
+	s.report = ""
+	s.isShowingPercentageProgress = false
+	s.percentageProgress = 0
+}
+
+func (s *SyncHandler) Render(window *nucular.Window, wallet app.WalletMiddleware, pageChangeFunc func(string)) {
```
itswisdomagain: `pageChangeFunc` should be `changePage`  so line 45 makes more sense  as `changePage("balance")`

nuklear/handler.go
```diff
@@ -2,52 +2,72 @@ package nuklear
 
 import (
 	"github.com/aarzilli/nucular"
+	"github.com/raedahgroup/godcr/app"
 	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/raedahgroup/godcr/nuklear/handlers"
 )
 
-type Handler interface {
+type NavPageHandler interface {
```
itswisdomagain: `NavPageHandler` and `StandalonePageHandler` do not need to be exported

nuklear/handlers/sync.go
```diff
@@ -0,0 +1,113 @@
...
+		s.syncBlockchain(window, wallet)
+	}
+
+	// change page onSyncStatusSuccess
+	if s.status == syncStatusSuccess {
+		pageChangeFunc("balance")
+		return
+	}
+
+	if contentWindow := helpers.NewWindow("dd", window, 0); contentWindow != nil {
```
itswisdomagain: Why is the title of this window  "dd"?

nuklear/handlers/sync.go
```diff
@@ -0,0 +1,113 @@
...
+type SyncHandler struct {
+	err                         error
+	isRendering                 bool
+	isShowingPercentageProgress bool
+	percentageProgress          int
+	report                      string
+	status                      syncStatus
+}
+
+type syncStatus uint8
```
itswisdomagain: this  exists in  the web interface  as well, abstract  shared code and  avoid repetition/duplication

itswisdomagain: on further thought, seems to me that all you really need is to determine if the sync operation is completed, the other status are irrelevant. You can replace `SyncHandler.status` with `SyncHandler.syncCompleted (bool)` instead

itswisdomagain (changes requested):

nuklear/desktop.go
```diff
@@ -2,83 +2,92 @@ package nuklear
...
+	// open wallet and start blockchain syncing in background
+	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
 	if err != nil {
 		return err
 	}
 
-	// todo run sync and show progress
+	if !walletExists {
+		// todo add ui to create wallet
+		err = fmt.Errorf("No wallet found. Use 'godcr create' to create a wallet before launching the desktop app")
```
itswisdomagain: This is already implemented, guess it was mistakenly taken out due to rebase. See https://github.com/raedahgroup/godcr/blob/master/nuklear/desktop.go#L50.

Also remember that after wallet is created, the next page should be sync page not balance...

nuklear/desktop.go
```diff
@@ -2,83 +2,92 @@ package nuklear
...
+	navPages        map[string]navPageHandler
+	standalonePages map[string]standalonePageHandler
 }
 
 func LaunchApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
 	desktop := &Desktop{
-		walletMiddleware: walletMiddleware,
-		pageChanged:      true,
-		currentPage:      homePage,
+		wallet:      walletMiddleware,
```
itswisdomagain: Why rename `walletMiddleware` to `wallet`. Remember there is a `walletcore.Wallet` type. The variable name `wallet` should only be associated with the `walletcore.Wallet` type to maintain consistency.

Always `wallet walletcore.Wallet` and `walletMiddleware app.WalletMiddleware`

nuklear/desktop.go
```diff
@@ -2,83 +2,92 @@ package nuklear
...
+	window.SetStyle(helpers.GetStyle())
+	desktop.masterWindow = window
 
-	// open wallet and start blockchain syncing in background
-	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
+	// initialize fonts for later use
+	err := helpers.InitFonts()
 	if err != nil {
-		return err
+		return nil
```
itswisdomagain: Should return `err` please

nuklear/desktop.go
```diff
@@ -2,83 +2,92 @@ package nuklear
...
 	}
 
-	if !walletExists {
-		desktop.currentPage = "createwallet"
+	// register nav page handlers
+	navPages := getNavPagesData()
```
itswisdomagain: `getNavPages` is better, same for `getStandalonePages`

nuklear/handler.go
```diff
@@ -6,60 +6,67 @@ import (
...
 }
 
-type page struct {
-	name       string
-	navLabel   string
-	standalone bool
-	handler    Handler
+type standalonePageHandler interface {
+	BeforeRender()
+	Render(*nucular.Window, app.WalletMiddleware, func(string))
```
itswisdomagain, line 16: Introduce an alias for this function so it's expected usage is immediately obvious:
`type pageChangeFunc func(string)`

nuklear/handler.go
```diff
@@ -6,60 +6,67 @@ import (
...
+}
+
+type navPageData struct {
```
itswisdomagain: Can this be just `navPage` rather than `navPageData`? It's obvious structs hold data for an object/entity, no need t append data to the object/entity name. It's like saying `WalletData`, just `Wallet` is fine. Same for `standalonePageData`, rename to `standalonePage`

itswisdomagain (changes requested):

itswisdomagain (changes requested):
Create wallet -> Passphrase inputs are not masked
screenshot 2019-03-04 at 7 58 51 am

Wallet seed doesn't display completely (the last word in the screenshot below is "k" which is not a word). Also, individual words should not be broken (aimless is broken in the screenshot below)
screenshot 2019-03-04 at 7 59 15 am

itswisdomagain (approved):

PR #213, opened 2019-03-01
Terminal ui tickets

Sirmorrison:
image


terminal/pages/tickets.go
```diff
@@ -0,0 +1,112 @@
+package pages
```
itswisdomagain: This file should be `purchase-tickets.go`. Let's make it easier for another dev to find his way around with less stress

terminal/pages/tickets.go
```diff
@@ -0,0 +1,112 @@
...
+	accountNumbers := make([]uint32, len(accounts))
+	accountOverviews := make([]string, len(accounts))
+	for index, account := range accounts {
+		accountOverviews[index] = fmt.Sprintf("%s - Total %s (Spendable %s)", account.Name, account.Balance.Total.String(), account.Balance.Spendable.String())
+		accountNumbers[index] = account.Number
+
+		var accountNum uint32
+		var numTickets, passphrase string
+		var spendUnconfirmed bool
+		body.AddItem(form.AddDropDown("Source Account", []string{accountOverviews[index]}, 0, func(option string, optionIndex int) {
```
itswisdomagain: You're adding views/controls/widgets to the form and body multiple times, that's terrible!
Your loop should only be to append elements to the `accountNumbers` and `accountOverviews` slices.

```
accountNumbers := make([]uint32, len(accounts))
accountOverviews := make([]string, len(accounts))
for index, account := range accounts {
	accountOverviews[index] = fmt.Sprintf("%s - Total %s (Spendable %s)", account.Name, account.Balance.Total.String(), account.Balance.Spendable.String())
	accountNumbers[index] = account.Number
}
```

terminal/pages/tickets.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/gdamore/tcell"
+	"github.com/raedahgroup/dcrlibwallet"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/rivo/tview"
+)
+
+func PurchaseTicketPage(wallet walletcore.Wallet, setFocus func(p tview.Primitive) *tview.Application, clearFocus func()) tview.Primitive {
+	body := tview.NewFlex().SetDirection(tview.FlexRow)
+	form := tview.NewForm()
+	textView := tview.NewTextView()
```
itswisdomagain: Variables should ideally be declared just before they're used. Don't declare variables in Jericho only to use them 10 years later in Samaria.

Where is `textView` and `form` used? Move the declarations to the line just before they're used

terminal/pages/tickets.go
itswisdomagain: Too much nesting done here, makes the code harder to read. Break it down. Notice also that the following example code follows the one above. Also notice the spacing for easier readability and the placement of variables (right before they're used)

```
accountNumbers := make([]uint32, len(accounts))
accountOverviews := make([]string, len(accounts))
for index, account := range accounts {
	accountOverviews[index] = fmt.Sprintf("%s - Total %s (Spendable %s)", account.Name, account.Balance.Total.String(), account.Balance.Spendable.String())
	accountNumbers[index] = account.Number
}

var accountNum uint32
form.AddDropDown("Source Account", accountOverviews, 0, func(option string, optionIndex int) {
	accountNum = accountNumbers[optionIndex]
})

var numTickets string
form.AddInputField("Number of tickets", "", 20, nil, func(text string) {
	numTickets = text
})
```

itswisdomagain (changes requested):
Please attend to the issues raised below before I can continue review

terminal/pages/tickets.go
Sirmorrison: declaring variables between the forms fields will not work as form is nested inside of a flex box with input parameters
`flex.Additem(form, int, int, int)`

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	"strconv"
+	"strings"
+
+	"github.com/gdamore/tcell"
+	"github.com/raedahgroup/dcrlibwallet"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/rivo/tview"
+)
+
+func PurchaseTicketPage(wallet walletcore.Wallet, setFocus func(p tview.Primitive) *tview.Application, clearFocus func()) tview.Primitive {
```
itswisdomagain: `PurchaseTicketsPage`

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	body := tview.NewFlex().SetDirection(tview.FlexRow)
+	textView := tview.NewTextView().SetTextAlign(tview.AlignCenter)
```
itswisdomagain: Why is `textView` declared here? Should be declared just above `form.AddButton` and the variable name should be more descriptive of its function. `errorTextView` perhaps

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	})
+
+	var numTickets string
+	form.AddInputField("Number of tickets", "", 20, nil, func(text string) {
+		numTickets = text
+	})
+
+	var spendUnconfirmed bool
+	form.AddCheckbox("Spend Unconfirmed", false, func(checked bool) {
+		if checked {
```
itswisdomagain: 2 problems here
- If it is checked, `spendUnconfirmed` becomes true, if it is unchecked, `spendUnconfirmed` _remains_ true
- This can be simplified to `spendUnconfirmed = checked`

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+			spendUnconfirmed = true
+		}
+	})
+
+	var passphrase string
+	form.AddPasswordField("wallet Passphrase", "", 20, '*', func(text string) {
+		passphrase = text
+	})
+	form.AddButton("Submit", func() {
```
itswisdomagain, line 59: Give a single line space before this

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+
+	var spendUnconfirmed bool
+	form.AddCheckbox("Spend Unconfirmed", false, func(checked bool) {
+		if checked {
+			spendUnconfirmed = true
+		}
+	})
+
+	var passphrase string
+	form.AddPasswordField("wallet Passphrase", "", 20, '*', func(text string) {
```
itswisdomagain: "W" in "Wallet Passphrase" should be capitalized. Also, the expression is now "Spending Passphrase"

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+		passphrase = text
+	})
+	form.AddButton("Submit", func() {
+		if len(numTickets) == 0 {
+			body.RemoveItem(textView)
+			body.AddItem(textView.SetText(fmt.Sprintf("Error: %s", "Please specify the number of tickets to purchase")).SetDoneFunc(func(key tcell.Key) {
```
itswisdomagain: Why do we have `SetDoneFunc` in each of the lines of code where the error message is displayed? Line 76 below is how I expect it to be...

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+			return
+		}
+
+		ticketHashes, err := submit(passphrase, numTickets, accountNum, spendUnconfirmed, wallet)
+		if err != nil {
+			body.RemoveItem(textView)
+			body.AddItem(textView.SetText(fmt.Sprintf("Error: %s", err.Error())), 0, 1, true)
+			return
+		}
+		body.RemoveItem(textView)
```
itswisdomagain: A single line space before this to show separation of concerns please

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	form.AddButton("Submit", func() {
+		if len(numTickets) == 0 {
+			body.RemoveItem(textView)
+			body.AddItem(textView.SetText(fmt.Sprintf("Error: %s", "Please specify the number of tickets to purchase")).SetDoneFunc(func(key tcell.Key) {
+			}), 0, 1, true)
+			return
+		}
+		if len(passphrase) == 0 {
+			body.RemoveItem(textView)
+			body.AddItem(textView.SetText(fmt.Sprintf("Error: %s", "please enter your wallet passphrase")).SetDoneFunc(func(key tcell.Key) {
```
itswisdomagain: Only use string formatting when combining string literal with variable(s). You can simply use `"Error: please enter your wallet passphrase"` directly

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+		ticketHashes, err := submit(passphrase, numTickets, accountNum, spendUnconfirmed, wallet)
+		if err != nil {
+			body.RemoveItem(textView)
+			body.AddItem(textView.SetText(fmt.Sprintf("Error: %s", err.Error())), 0, 1, true)
+			return
+		}
+		body.RemoveItem(textView)
+		body.AddItem(textView.SetText(fmt.Sprintf("You have purchased %d ticket(s)\n%s", len(ticketHashes), strings.Join(ticketHashes, "\n"))), 0, 1, true)
+	})
+	form.SetCancelFunc(func() {
```
itswisdomagain: This can simplify to `form.SetCancelFunc(clearFocus)`

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	requiredConfirmations := walletcore.DefaultRequiredConfirmations
+	if spendUnconfirmed {
+		requiredConfirmations = 0
+	}
+
+	request := dcrlibwallet.PurchaseTicketsRequest{
+		RequiredConfirmations: uint32(requiredConfirmations),
+		Passphrase:            []byte(passphrase),
+		NumTickets:            uint32(nTickets),
+		Account:               uint32(accountNum),
```
itswisdomagain: `accountNum` is already `uint32`, conversion is unnecessary

itswisdomagain (changes requested):

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/rivo/tview"
+)
+
+func PurchaseTicketsPage(wallet walletcore.Wallet, setFocus func(p tview.Primitive) *tview.Application, clearFocus func()) tview.Primitive {
+	body := tview.NewFlex().SetDirection(tview.FlexRow)
+
+	body.AddItem(tview.NewTextView().SetTextAlign(tview.AlignCenter).SetText("Purchase Tickets"), 4, 1, false)
+
+	outputTextView := tview.NewTextView().SetTextAlign(tview.AlignCenter)
```
itswisdomagain: You can move this declaration lower to just before `form.AddButton` and use `tview.NewTextView().SetTextAlign(tview.AlignCenter)` below in line 23. The variable will not be reused in this case and it's pointless to keep a reference to it.

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,114 @@
...
+	})
+
+	var numTickets string
+	form.AddInputField("Number of tickets", "", 20, nil, func(text string) {
+		numTickets = text
+	})
+
+	var spendUnconfirmed bool
+	form.AddCheckbox("Spend Unconfirmed", false, func(checked bool) {
+			spendUnconfirmed = checked
```
itswisdomagain: go fmt ./... please

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,114 @@
...
+
+	var passphrase string
+	form.AddPasswordField("Spending Passphrase", "", 20, '*', func(text string) {
+		passphrase = text
+	})
+
+	form.AddButton("Submit", func() {
+		if len(numTickets) == 0 {
+			body.RemoveItem(outputTextView)
+			body.AddItem(outputTextView.SetText(fmt.Sprintf("Error: Please specify the number of tickets to purchase")), 0, 1, true)
```
itswisdomagain: P in "Please" should not be capitalized

itswisdomagain: Instead of having this set of code in multiple places, you can have a function `dislayError` that removes the outputTextView and re-adds it when called

itswisdomagain (changes requested):

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+			body.AddItem(outputTextView.SetText(fmt.Sprintf(err.Error())), 0, 1, true)
+		} else {
+			body.AddItem(outputTextView.SetText(fmt.Sprintf("You have purchased %d ticket(s)\n%s", len(ticketHashes), strings.Join(ticketHashes, "\n"))), 0, 1, true)
+		}
+	}
+
+	form.AddButton("Submit", func() {
+		if len(numTickets) == 0 {
+			err := errors.New("Error: please specify the number of tickets to purchase")
+			outputMessage(nil, err)
```
itswisdomagain: Make the `outputMessage` function take a single string parameter. Then send the error message directly rather than creating an error object

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+	var passphrase string
+	form.AddPasswordField("Spending Passphrase", "", 20, '*', func(text string) {
+		passphrase = text
+	})
+
+	outputTextView := tview.NewTextView().SetTextAlign(tview.AlignCenter)
+	outputMessage := func(ticketHashes []string, err error) {
+		body.RemoveItem(outputTextView)
+		if err != nil {
+			body.AddItem(outputTextView.SetText(fmt.Sprintf(err.Error())), 0, 1, true)
```
itswisdomagain: `fmt.Sprintf(err.Error())` is the same as `err.Error()`, only use fmt when COMBINING literal text with variable

terminal/pages/purchase-tickets.go
```diff
@@ -0,0 +1,121 @@
...
+			return
+		}
+
+		ticketHashes, err := submit(passphrase, numTickets, accountNum, spendUnconfirmed, wallet)
+		if err != nil {
+			outputMessage(nil, err)
+			return
+		}
+
+		outputMessage(ticketHashes, nil)
```
itswisdomagain: Once you make the `outputMessage` function take a single string parameter, you can no longer pass the ticketHashes slice to the function.  Convert to string using `fmt.Sprintf("You have purchased %d ticket(s)\n%s", len(ticketHashes), strings.Join(ticketHashes, "\n"))` and pass the string message to be displayed

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #214, opened 2019-03-04
Fixed the bug seen when wallet does not exist and user chose not to create it

### LINKED ISSUE #179: error/crash when running godcr if wallet doesn't exist

ademuanthony:
Resolves #179

### DIFF
--- a/cli/runner/helpers.go
+++ b/cli/runner/helpers.go
@@ -24,19 +24,19 @@ func CommandRequiresWallet(command flags.Commander) bool {
 
 // prepareWallet gets a wallet ready for use by opening the wallet using the provided walletMiddleware
 // and performing sync operations if requested
-func prepareWallet(ctx context.Context, middleware app.WalletMiddleware, options config.CliOptions) error {
+func prepareWallet(ctx context.Context, middleware app.WalletMiddleware, options config.CliOptions) (bool, error) {
 	walletExists, err := walletloader.OpenWallet(ctx, middleware)
 	if err != nil || !walletExists {
-		return err
+		return walletExists, err
 	}
 
 	if options.SyncBlockchain {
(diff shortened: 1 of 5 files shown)


cli/runner/helpers.go
```diff
@@ -24,19 +24,19 @@ func CommandRequiresWallet(command flags.Commander) bool {
 
 // prepareWallet gets a wallet ready for use by opening the wallet using the provided walletMiddleware
 // and performing sync operations if requested
-func prepareWallet(ctx context.Context, middleware app.WalletMiddleware, options config.CliOptions) error {
+func prepareWallet(ctx context.Context, middleware app.WalletMiddleware, options config.CliOptions) (bool, error) {
```
itswisdomagain: Using named return variables will make this function easier to understand. `(walletExists bool, err 
 error)`. You can then simply `return` in lines 30, 36 and 39

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #221, opened 2019-03-08
Terminal create wallet

Sirmorrison:
image


terminal/terminal.go
```diff
@@ -15,40 +15,19 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 	tviewApp := tview.NewApplication()
 
 	// open wallet and start blockchain syncing in background
+	var layout tview.Primitive
 	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
 	if err != nil {
 		return err
 	}
 	if walletExists {
-		// `Run` blocks until app.Stop() is called before returning
```
itswisdomagain, line 30: Restore this comment, it's important

itswisdomagain (changes requested):

itswisdomagain (changes requested):
Make use of the `createwallet.go` file, and re-order the UI elements: Display wallet seed before the important notice and before the buttons

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"context"
+
+	"github.com/raedahgroup/godcr/app"
+	"github.com/raedahgroup/godcr/terminal/helpers"
+	"github.com/rivo/tview"
+)
+
+func pageLoader(ctx context.Context, tviewApp *tview.Application, walletMiddleware app.WalletMiddleware) tview.Primitive {
+	newPrimitive := func(text string) tview.Primitive {
```
itswisdomagain: `newPrimitive` is not a good name, `primitive` is generic, can mean button or image or any other number of things really. Also not good enough to just say `newLabel`. `newWordWrappedTextView` perhaps

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+		return tview.NewTextView().
+			SetTextAlign(tview.AlignCenter).
+			SetText(text).SetRegions(true).
+			SetWordWrap(true)
+	}
+
+	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
```
itswisdomagain: This doesn't belong in this file.
- This file should only deal with creating wallets, not opening one if it exists
- Very soon, we'd support creating new wallets even if a wallet already exists, which makes the previous point even more important

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+			layout.AddItem(outputTextView.SetText(text), 0, 1, true)
+		}
+
+		form := tview.NewForm()
+		var password string
+		form.AddPasswordField("Passphrase", "", 20, '*', func(text string) {
+			password = text
+		})
+
+		var confPassword string
```
itswisdomagain: `confirmPassword` is better
Or even `passphrase` and `confirmPassphrase` respectively

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+		form.AddButton("Create", func() {
+			if len(password) == 0 {
+				outputMessage("Passphrase cannot empty")
+				return
+			}
+			if password != confPassword {
+				outputMessage("passphrase does not match")
+				return
+			}
+			if hasStoredSeed == false {
```
itswisdomagain: `if !hasStoredSeed`

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+				outputMessage("please Store seed in a szfe location and check the box")
```
itswisdomagain: **safe**
"please" is not capitalised, but "Store" is?

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+				return
+			}
+
+			err := CreateWallet(tviewApp, seed, password, walletMiddleware)
```
itswisdomagain: The `CreateWallet` function isn't very big, you can just bring the code over here directly and do away with the function; unless the function is called from more than 1 place

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+			if err != nil {
+				outputMessage(err.Error())
+				return
+			}
+
+			pages.AddAndSwitchToPage("main", terminalLayout(ctx, tviewApp, walletMiddleware), true)
```
itswisdomagain: Should also remove the "create" page.

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+		layout.AddItem(form, 10, 1, true)
+
+		seedView := tview.NewTextView().SetTextAlign(tview.AlignCenter).SetText(seed).SetRegions(true).SetWordWrap(true)
+		seedView.SetBorder(true).SetTitle("Wallet Seed").SetTitleColor(helpers.PrimaryColor)
+		layout.AddItem(seedView, 7, 1, false)
+
+		seedInfo := tview.NewTextView().SetTextAlign(tview.AlignCenter).SetText(`Keep the seed in a safe place as you will NOT be able to restore your wallet without it. Please keep in mind that anyone who has access to the seed can also restore your wallet thereby giving them access to all your funds, so it is imperative that you keep it in a secure location.`).
+		SetRegions(true).SetWordWrap(true)
+		seedInfo.SetBorder(true).SetTitle("IMPORTANT NOTICE:").SetTitleColor(helpers.WarnColor)
+		layout.AddItem(seedInfo, 7, 1, false)
```
itswisdomagain: The has stored seed checkbox should come below the seed and notice boxes, you should aim to replicate the http version

terminal/createwallet.go
```diff
@@ -0,0 +1,108 @@
...
+
+		layout.SetFullScreen(true).SetBorderPadding(3, 1, 6, 4)
+
+		layout.SetFullScreen(true)
+
```
itswisdomagain: How many times do you have to call `layout.SetFullScreen(true)`?

terminal/helpers/colors.go
```diff
@@ -5,4 +5,6 @@ import "github.com/gdamore/tcell"
 var (
 	DecredLightColor = tcell.NewRGBColor(112, 203, 255)
 	DecredColor      = tcell.NewRGBColor(41, 112, 255)
+	WarnColor        = tcell.NewRGBColor(255, 0, 0)
+	PrimaryColor     = tcell.NewRGBColor(0, 255, 255)
```
itswisdomagain: What does `PrimaryColor` mean?

terminal/layout.go
```diff
@@ -0,0 +1,118 @@
...
+
+func terminalLayout(ctx context.Context, tviewApp *tview.Application, walletMiddleware app.WalletMiddleware) tview.Primitive {
+	newPrimitive := func(text string) tview.Primitive {
+		return tview.NewTextView().
+			SetTextAlign(tview.AlignCenter).
+			SetText(text)
+	}
+
+	syncBlockchain(walletMiddleware)
+	if Status == walletcore.SyncStatusError {
```
itswisdomagain: Line 22 above starts a background process - a subroutine. By the time line 23 is executed, the sync status may not have changed and the if statements below become pretty much invalid.

The background process started by line 22 has callback functions, similar to nodejs callback functions; the `SyncStarted`, `SyncEnded`, `OnHeadersFetched`, `OnDiscoveredAddress` and `OnRescanningBlocks` functions are called at different times while the background process is running.

You should add code in those `SyncStarted`, `OnHeadersFetched`, `OnDiscoveredAddress` and `OnRescanningBlocks` functions to display the report **whenever** those functions are called instead of using the 2 if statements below.

You should also add the code in the 3rd if statement below to the `SyncEnded` function so that those code only actually run when the sync has ended.

itswisdomagain (dismissed):

terminal/helpers/walletloader.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/raedahgroup/godcr/app"
+)
+
+// this method may stall until previous godcr instances are closed (especially in cases of multiple dcrlibwallet instances)
```
ademuanthony: comments should start with the function name `OpenWalletIfExist`

terminal/helpers/walletloader.go
```diff
@@ -0,0 +1,49 @@
...
+// hence the need for ctx, so user can cancel the operation if it's taking too long
+func OpenWalletIfExist(ctx context.Context, walletMiddleware app.WalletMiddleware) (walletExists bool, err error) {
```
ademuanthony: This is a duplicate of `walletloader.OpenWallet`
We should remove this function and use that instead

terminal/terminal.go
```diff
@@ -14,109 +14,72 @@ import (
 func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
 	tviewApp := tview.NewApplication()
 
-	// open wallet and start blockchain syncing in background
-	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
+	walletExists, err := helpers.OpenWalletIfExist(ctx, walletMiddleware)
```
ademuanthony: WE should use  `walletloader.OpenWallet` instead

terminal/terminal.go
```diff
@@ -14,109 +14,72 @@ import (
...
-	menuColumn.AddItem("Purchase Tickets", "", 't', func() {
-		changePageColumn(pages.PurchaseTicketsPage(walletMiddleware, setFocus, clearFocus))
-	})
+	return nil
+}
 
-	menuColumn.AddItem("Exit", "", 'q', func() {
-		tviewApp.Stop()
-	})
+var Report string
```
ademuanthony: this field is not used outside this package, it should be `report` instead

terminal/terminal.go
```diff
@@ -14,109 +14,72 @@ import (
...
+var Status walletcore.SyncStatus
```
ademuanthony: this field is not used outside this package, it should be `status` instead

ademuanthony (changes requested):
You have done such a neat job. 
I am impressed with how you maintained the line of sight. 
Here are few change requests
Please work on them and also rebase and `go fmt ./...` so that we can get this merged

ademuanthony (approved):

ademuanthony:
I think this PR is now ready

terminal/createwallet.go
```diff
@@ -0,0 +1,163 @@
...
+	seedInfo.SetBorder(true).SetTitle("IMPORTANT NOTICE:").SetTitleColor(helpers.WarnColor)
+	layout.AddItem(seedInfo, 7, 3, true)
+
+	var hasStoredSeed bool
+	checkbox := tview.NewCheckbox().SetLabel("I have stored wallet seed  ").SetChecked(false).SetChangedFunc(func(checked bool) {
+		hasStoredSeed = checked
+	})
+	layout.AddItem(checkbox, 2, 1, true)
+
+	createButton := tview.NewForm().
```
itswisdomagain: No need to create a form just to display a button. Instead...
```
createWalletButton := tview. NewButton("Create Wallet").SetSelectedFunc(func() {
...
})
```

- Also notice I opted for `createWalletButton` instead of `createButton`

terminal/createwallet.go
```diff
@@ -0,0 +1,163 @@
...
+		AddButton("Create Wallet", func() {
+			if len(passphrase) == 0 {
+				outputMessage("Passphrase cannot empty")
+				return
+			}
+			if passphrase != confirmPassphrase {
+				outputMessage("passphrase does not match")
```
itswisdomagain: Output message should begin with capital letter

terminal/createwallet.go
```diff
@@ -0,0 +1,163 @@
...
+				outputMessage("Please store seed in a safe location and check the box")
+				return
+			}
+
+			err = walletMiddleware.CreateWallet(passphrase, seed)
+			if err != nil {
+				outputMessage(err.Error())
+			}
+
+			res := pageLoader(tviewApp, walletMiddleware)
```
itswisdomagain: - What does `res` mean? Response? Resource? Need to give better variable names.
- In trying to understand what `res` represents, I discovered that `pageLoader` has 2 issues
  - The `pageLoader` function returns a nil primitive - doesn't make sense. Why does a PAGE **LOADER** return no page?
  - The `pageLoader` function seems to have nothing to do with loading pages, more to do with listening for sync status updates. Function needs to be renamed appropriately

terminal/terminal.go
```diff
@@ -3,120 +3,92 @@ package terminal
...
 
-func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMiddleware) tview.Primitive {
-	gridLayout := tview.NewGrid().
-		SetRows(3, 0).
-		SetColumns(30, 0, 1).
-		SetGap(0, 2).
-		SetBordersColor(helpers.DecredLightColor)
-
-	gridLayout.SetBackgroundColor(tcell.ColorBlack)
+func pageLoader(tviewApp *tview.Application, walletMiddleware app.WalletMiddleware) tview.Primitive {
```
itswisdomagain: - This `pageLoader` function returns a nil primitive - doesn't make sense. Why does a PAGE **LOADER** return no page?
- Also, this function seems to have nothing to do with loading pages, more to do with listening for sync status updates. Function needs to be renamed appropriately

terminal/createwallet.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		if event.Key() == tcell.KeyTAB {
+			setFocus(confirmPassphraseField)
+			return nil
+		}
+
+		return event
+	})
+
+	// listen to escape and left key press events on all form items and buttons
```
itswisdomagain: Correct this comment, and add similar comment to each section of code where you set input capture. The comment should accurately tell what the input capture function does

terminal/helpers/walletloader.go
```diff
@@ -0,0 +1,49 @@
+package helpers
```
itswisdomagain: Is the content of this file any different from similar code found in the `cli` package? We should avoid code duplication and opt for shared code/functions.

itswisdomagain (changes requested):
I notice some improvements with how sync updates are handled but the desired solution isn't what we have here yet. Primarily, sync update messages should not be displayed on (logged to) the terminal/command prompt window directly. Such informative messages should be displayed on a primitive - perhaps a text view.

See other required changes below:

terminal/terminal.go
```diff
@@ -3,120 +3,90 @@ package terminal
...
-	"github.com/raedahgroup/godcr/terminal/primitives"
 	"github.com/rivo/tview"
 )
 
 func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware) error {
 	tviewApp := tview.NewApplication()
 
-	// open wallet and start blockchain syncing in background
-	walletExists, err := openWalletIfExist(ctx, walletMiddleware)
+	walletExists, err := nuklear.OpenWalletIfExist(ctx, walletMiddleware)
```
ademuanthony: This cross-reference is not ideal. I think we should move `nuklear.OpenWalletIfExist` to walletcore so that other interfaces can call it from there.

terminal/terminal.go
```diff
@@ -3,120 +3,90 @@ package terminal
...
-		tviewApp.Draw()
-		tviewApp.SetFocus(menuColumn)
+	for status := range syncStatus {
+		if status == walletcore.SyncStatusError {
+			tviewApp.Stop()
+			return nil
+		}
+		if status == walletcore.SyncStatusInProgress {
+			msgOutput := fmt.Sprintf(syncMessage)
+			fmt.Println(msgOutput)
```
ademuanthony: There is a review comment that recommended that this message is written to a Primitive, maybe, a textview

ademuanthony (changes requested):
There is a panic during shutdown

itswisdomagain (approved):

ademuanthony (approved):

PR #222, opened 2019-03-11
fix terminal UI content mismatch when navigating menu list

Sirmorrison:
fix issues with page  content mismatch when navigating from one menu item to another on the terminal interface another


terminal/terminal.go
```diff
@@ -68,32 +67,49 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
 	}
 
 	setFocus := tviewApp.SetFocus
+
+	menuColumn := tview.NewList()
+	var page tview.Primitive
```
itswisdomagain: Rename to `activePage`

terminal/terminal.go
```diff
@@ -68,32 +67,49 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
...
 	clearFocus := func() {
+		gridLayout.RemoveItem(page)
```
itswisdomagain: Also do this in the `changePageColumn` function. Remove the active page before adding the new one.

terminal/terminal.go
```diff
@@ -68,32 +67,49 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
...
-			changePageColumn(pages.StakeinfoPage(walletMiddleware, setFocus, clearFocus))
-		}).
-		AddItem("Purchase Tickets", "", 't', func() {
-			changePageColumn(pages.PurchaseTicketsPage(walletMiddleware, setFocus, clearFocus))
-		}).
-		AddItem("Exit", "", 'q', func() {
-			tviewApp.Stop()
-		})
+	menuColumn.AddItem("Balance", "", 'b', func() {
+		page = pages.BalancePage(walletMiddleware, setFocus, clearFocus)
```
itswisdomagain: Do not assign `page` here. Simply call `changePageColumn` and in that `changePageColumn` function, remove the currently active page, assign the new page and add the new page.

terminal/terminal.go
```diff
@@ -68,32 +67,49 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
...
+		changePageColumn(page)
+	}).SetSelectedFocusOnly(true)
```
itswisdomagain: Don't see a need for this `SetSelectedFocusOnly` function call

itswisdomagain (changes requested):

terminal/terminal.go
```diff
@@ -24,6 +24,7 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 		if err != nil {
 			fmt.Println(err)
```
itswisdomagain, line 25: If there's a sync error, the error should be returned thereby canceling the app launch

terminal/terminal.go
```diff
@@ -24,6 +24,7 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
...
 		}
+
 		// `Run` blocks until app.Stop() is called before returning
 		layout := terminalLayout(tviewApp, walletMiddleware)
 		return tviewApp.SetRoot(layout, true).SetFocus(layout).Run()
```
itswisdomagain: Rewrite lines 28-30 to aid readability and understanding

```
layout := terminalLayout(tviewApp, walletMiddleware)
tviewApp.SetRoot(layout, true)

// `Run` blocks until app.Stop() is called before returning
return tviewApp.Run()
```

Also note that `SetFocus` is not necessary after calling `SetRoot`. Always make good use of the documentation: https://godoc.org/github.com/rivo/tview#Application.SetRoot
Screenshot 2019-03-12 at 12 35 16 PM

terminal/terminal.go
```diff
@@ -56,44 +57,55 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 }
 
 func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMiddleware) tview.Primitive {
-	var menuColumn *tview.List
-
 	header := tview.NewTextView().SetTextAlign(tview.AlignCenter).SetText(fmt.Sprintf("\n%s Terminal", strings.ToUpper(app.Name)))
```
itswisdomagain: I see a lot of this `tview.NewTextView().SetTextAlign(tview.AlignCenter)` in the terminal code base, worth refactoring into a helper function. Create a `primitives.go` file in the `terminal/helpers` package and add the following function:

```
func CenterAlignedTextView(text string)  *tview.TextView {
	return NewTextView(text, tview.AlignCenter)
}

func NewTextView(text string, alignment int) *tview.TextView {
	return tview.NewTextView().SetTextAlign(alignment).SetText(text)
}
```
Then replace line 60 with 
```
header := helpers.CenterAlignedTextView(fmt.Sprintf("\n%s Terminal", strings.ToUpper(app.Name)))
```

Also update the rest of the code to use either of these functions where appropriate rather than always writing `tview.NewTextView().SetTextAlign(tview.AlignCenter)`

terminal/terminal.go
```diff
@@ -56,44 +57,55 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
...
+
+	menuColumn.AddItem("Send", "", 's', func() {
+		changePageColumn(pages.SendPage(setFocus, clearFocus))
+	})
+
+	menuColumn.AddItem("History", "", 'h', func() {
+		changePageColumn(pages.HistoryPage(walletMiddleware, setFocus, clearFocus))
+	})
+
+	menuColumn.AddItem("Stakeinfo", "", 'k', func() {
```
itswisdomagain: "Stake Info"

terminal/terminal.go
```diff
@@ -56,44 +57,55 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
...
+	changePageColumn := func(page tview.Primitive) {
+		gridLayout.RemoveItem(activePage)
+		activePage = page
+		gridLayout.AddItem(activePage, 1, 1, 1, 1, 0, 0, true)
 	}
 
 	setFocus := tviewApp.SetFocus
+
+	menuColumn := tview.NewList()
 	clearFocus := func() {
```
itswisdomagain, line 76: `clearFocus` should also do `gridLayout.RemoveItem(activePage)`

itswisdomagain (changes requested):

terminal/terminal.go
```diff
@@ -102,7 +114,6 @@ func terminalLayout(tviewApp *tview.Application, walletMiddleware app.WalletMidd
 	gridLayout.AddItem(header, 0, 0, 1, 2, 0, 0, false)
 	// Layout for screens with two column
 	gridLayout.AddItem(menuColumn, 1, 0, 1, 1, 0, 0, true)
-	gridLayout.AddItem(pages.BalancePage(walletMiddleware, setFocus, clearFocus), 1, 1, 1, 1, 0, 0, true)
```
itswisdomagain, line 105: In all other godcr GUIs, the balance page is loaded by default. Follow same pattern as well, use `changePageColumn(pages.BalancePage(walletMiddleware, setFocus, clearFocus))`

itswisdomagain (changes requested):

terminal/terminal.go
```diff
@@ -20,14 +20,14 @@ func StartTerminalApp(ctx context.Context, walletMiddleware app.WalletMiddleware
 		return err
 	}
 	if walletExists {
-		err := SyncBlockChain(ctx, walletMiddleware)
-		if err != nil {
-			fmt.Println(err)
-		}
+		// err := SyncBlockChain(ctx, walletMiddleware)
```
itswisdomagain: You really need to stop doing this!

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #231, opened 2019-03-21
generate address directly on receive page

### LINKED ISSUE #182: http receive page - generate address directly if wallet has only 1 account
If wallet has only 1 account, generate receive address for that account and display instead of showing the account selection widget. This is already done in the cli interface. Same needs to be done for the http interface.

ademuanthony:
Resolve [Issue 182](https://github.com/raedahgroup/godcr/issues/182)

This PR generate address automatically in the receive on the web and only show the generate button if the user has more than 1 account

The size of the account control is also reduced while the embedded jquery script for the address generation is rewritten and moved to a new stimulusjs controller


web/routes/handlers.go
```diff
@@ -160,7 +160,14 @@ func (routes *Routes) receivePage(res http.ResponseWriter, req *http.Request) {
 
 	data := map[string]interface{}{
 		"accounts": accounts,
+		"accountsCount": len(accounts),
```
itswisdomagain: This value can be evaluated in the template, no need to include it in this map

web/views/receive.html
```diff
@@ -11,29 +11,38 @@
                        <h5 class="card-title">Generate Receive Address</h5>
                        <form id="generate-form">
                             <div class="row">
-                                <div class="col-md-9 col-sm-12">
+                                <div class="col-md-6 col-sm-12">
                                   <div class="form-group">
                                        <label for="account">Account</label>
-                                       <select class="form-control" name="account" id="account">
+                                        {{ if gt .accountsCount 1 }}
```
itswisdomagain: The send page template has a simpler, more straightforward way of doing this:
```
<select data-target="send.sourceAccount" class="form-control" id="source-account" name="source-account" {{ if eq (len .accounts) 1 }}disabled{{ end }}>
    <option value="{{ $account.Number }}">{{ accountString $account }}</option>
{{ end }}	                                              
</select>
```

Notice also that `.accountsCount` is not required

web/views/receive.html
```diff
@@ -11,29 +11,38 @@
...
+                                        {{ end  }}
                                   </div>
                                 </div>
-                                <div class="col-md-3 col-sm-12">
-                                    <div class="form-group btn-form-group">
-                                      <button type="submit" id="submit" class="btn btn-success form-control">Generate</button>
+                                {{ if gt .accountsCount 1 }}
+                                    <div class="col-md-3 col-sm-12">
+                                        <div class="form-group btn-form-group">
+                                            <button data-action="click->receive#generate" type="submit" id="submit" class="btn btn-success form-control">Generate</button>
```
itswisdomagain: In keeping feature-par with the mobile wallet apps, let's retire this generate button and auto-generate an address when a different account is selected from the dropdown

web/static/app/src/controllers/receive_controller.js
```diff
@@ -0,0 +1,29 @@
...
+import axios from 'axios'
+
+export default class extends Controller {
+  static get targets () {
+    return [
+      'account', 'address', 'image'
+    ]
+  }
+
+  generate (evn) {
```
itswisdomagain: `e` or `event`

web/static/app/src/controllers/receive_controller.js
```diff
@@ -0,0 +1,29 @@
...
+    evn.preventDefault()
+    let _this = this
+    axios.get('/generate-address/' + this.accountTarget.value)
+      .then((response) => {
+        let result = response.data
+        if (result.success) {
+          _this.addressTarget.textContent = result.address
+          _this.imageTarget.setAttribute('src', result.imageData)
+        } else {
+          window.alert(result.message)
```
itswisdomagain: Alert dialogs = terrible user experience; display error messages on the page instead.

web/static/app/src/controllers/receive_controller.js
```diff
@@ -0,0 +1,29 @@
...
+        }
+      })
+      .catch((error) => {
+        console.log(error)
```
itswisdomagain: This shouldn't be in production app. Also, don't use `window.alert`.

itswisdomagain (changes requested):
Let's aim to make the receive page look a lot more similar to what we have on the Android/iOS apps:


Required changes/additions would include:
- Remove the generate the button and generate address when the account dropdown selection is changed
- Add "Click to copy" button
- If a previously generated address exists that has not been used, invoking `wallet.CurrentAddress` will return that address. Should include a means to force generating a different (next) address. The mobile apps have an option for this on the top right

web/views/receive.html
itswisdomagain: Just noticed you refactored this code on another PR; should make this into a reusable template in `utils.html` and prevent duplicating code

ademuanthony: moving this to a template may not work for other use cases where change event is not used the way it will now be used here to generate new address

ademuanthony:
I have worked on all the changes requested in the review

itswisdomagain (changes requested):
Remove auto generated font awesome files from dist, probably add `dist/font-awesome*` to gitignore for now. Will test/review again after this

.gitignore
```diff
@@ -7,5 +7,5 @@ godcr
 godcr.conf
 godcr.exe
 web/static/app/node_modules
-*.log
-*.bat
+web/static/dist/font-awesome/*
+*.log*.bat
```
itswisdomagain: This should be separate lines

web/routes/templates.go
```diff
@@ -46,5 +46,11 @@ func templateFuncMap() template.FuncMap {
 		"timestamp": func() int64 {
 			return time.Now().Unix()
 		},
+		"gt": func(a, b int) bool {
```
itswisdomagain: This isn't necessary, there's a default `gt` function for go templates

web/static/app/package.json
```diff
@@ -16,10 +16,14 @@
...
   },
   "devDependencies": {
-    "autoprefixer": "^9.4.7",
     "@babel/core": "^7.1.6",
     "@babel/plugin-proposal-class-properties": "^7.3.0",
     "@babel/preset-env": "^7.1.6",
+    "@fortawesome/fontawesome-svg-core": "^1.2.17",
+    "@fortawesome/free-brands-svg-icons": "^5.8.1",
+    "@fortawesome/free-regular-svg-icons": "^5.8.1",
+    "@fortawesome/free-solid-svg-icons": "^5.8.1",
```
itswisdomagain, line 23: Why do we have several icon-font-related libraries? won't font-awesome alone do the job? I'd prefer we use only font-awesome

web/static/app/src/controllers/receive_controller.js
```diff
@@ -0,0 +1,34 @@
...
+
+export default class extends Controller {
+  static get targets () {
+    return [
+      'account', 'address', 'image'
+    ]
+  }
+
+  copyToClipboard () {
+    copyToClipboard(this.addressTarget.textContent)
```
itswisdomagain, line 13: Does this work? Even if it does, there's need to rename one of these functions to aid readability

web/static/app/package.json
```diff
@@ -31,6 +35,7 @@
     "eslint-plugin-node": "^8.0.0",
     "eslint-plugin-promise": "^4.0.1",
     "eslint-plugin-standard": "^4.0.0",
+    "file-loader": "^3.0.1",
```
itswisdomagain: Is this needed or used?

itswisdomagain (changes requested):

web/static/app/package.json
```diff
@@ -16,10 +16,12 @@
     "stimulus": "^1.1.0"
   },
   "devDependencies": {
-    "autoprefixer": "^9.4.7",
     "@babel/core": "^7.1.6",
     "@babel/plugin-proposal-class-properties": "^7.3.0",
     "@babel/preset-env": "^7.1.6",
+    "@fortawesome/fontawesome-svg-core": "^1.2.17",
+    "@fortawesome/free-solid-svg-icons": "^5.8.1",
+    "autoprefixer": "^9.4.7",
```
itswisdomagain: Need to update webpack config for this to become relevant. See https://github.com/postcss/autoprefixer#webpack

web/static/app/package.json
```diff
@@ -16,10 +16,12 @@
     "stimulus": "^1.1.0"
   },
   "devDependencies": {
-    "autoprefixer": "^9.4.7",
     "@babel/core": "^7.1.6",
     "@babel/plugin-proposal-class-properties": "^7.3.0",
     "@babel/preset-env": "^7.1.6",
+    "@fortawesome/fontawesome-svg-core": "^1.2.17",
+    "@fortawesome/free-solid-svg-icons": "^5.8.1",
```
itswisdomagain, line 23: I believe there's another PR that introduced font-awesome, let's stick to one library for fonts, preferably font-awesome. These 2 libraries are only added because of the copy icon. If we can't get that icon in font-awesome, let's use a link that looks like [(copy)](url) instead

(UPDATE)
Seems that PR uses this same library now. If so, we can keep this.

itswisdomagain (changes requested):
Will build, run and test after these changes.

web/static/app/package.json
ademuanthony: we are using the same lib

.gitignore
```diff
@@ -7,5 +7,6 @@ godcr
 godcr.conf
 godcr.exe
 web/static/app/node_modules
+web/static/dist/font-awesome/*
```
itswisdomagain: Is this still necessary as we're no longer using font-awesome?

web/static/app/src/utils.js
```diff
@@ -0,0 +1,46 @@
...
+  el.select()
+  document.execCommand('copy')
+  document.body.removeChild(el)
+  if (selected) {
+    document.getSelection().removeAllRanges()
+    document.getSelection().addRange(selected)
+  }
+}
+
+export const setErrorMessage = (ctl, message) => {
```
itswisdomagain: - Rename `ctl` to `controller`
- Refactor all controllers to use these utility methods

web/static/app/src/css/style.scss
```diff
@@ -243,3 +243,18 @@ select#source-account[disabled] {
     transform: translate(19px, 0);
   }
 }
+
+.lead-text {
+  font-size: 30px;
```
itswisdomagain: This text doesn't need to be this big, it's just a hint sort of text. We're not looking to do a perfect replica of the android app, but take what is there and make it work for the web.

web/views/receive.html
```diff
@@ -2,38 +2,43 @@
...
-                                </div>
-                                <div class="col-md-3 col-sm-12">
-                                    <div class="form-group btn-form-group">
-                                      <button type="submit" id="submit" class="btn btn-success form-control">Generate</button>
-                                    </div>
-                                </div>
-                            </div>
-                        </form>
+                       <h5 class="card-title">Receiving Decred</h5>
+                       <p class="lead-text">Each time you request payment, a new <br/>address is created to protect your privacy</p>
```
itswisdomagain: This doesn't need to be in 2 lines, should preferably be in 1 line, in smaller fonts. It's just a notice, that users will get accustomed to after a few times opening the page, doesn't need to be the focus of attention

web/views/receive.html
```diff
@@ -2,38 +2,43 @@
...
+                               {{ end  }}
+                               </div>
+                           </div>
+                       </div>
                         <div class="result-card">
                             <div id="result-text">
                                 <strong>Address:</strong>
-                                <span id="address"></span>
+                                <span data-target="receive.address" id="address">{{ .address }}</span>
+                                <button data-action="click->receive#copyAddressToClipboard" class="btn btn-xs btnCopy" title="Copy to clipboard"><i class="fas fa-copy"></i></button>
```
itswisdomagain: This button icon doesn't look conspicuous at all, ran this PR for the first time yesterday and "almost" missed the copy button. Please, let's go with `<button class="btn btn-sm btn-link">(copy)</button>`. And then after the address is successfully copied, the button should say `(copied)` for a few seconds before reverting back to `(copy)`

itswisdomagain (changes requested):

itswisdomagain:
Also need to have a "Generate New Address" button to ignore the current address and generate a new address using `wallet.GenerateNewAddress`

itswisdomagain:
Screen Shot 2019-04-06 at 3 48 55 AM

itswisdomagain (approved):

PR #232, opened 2019-03-21
http send page cleanup

### LINKED ISSUE #228: clean up http send page
Send page, when I try to copy the balance, the balance amounts are not highlightable. This should be fixed. Also, I see a missing ) bracket. Pretty obvious thing.

Trying to click the  select custom inputs button, but it doesnt work, shows errors at the bottom instead "Destination address should not be empty,  Amount must be a non-zero positive number"

ademuanthony:
Resolves #228

This PR makes the account balance on the http send page selectable and remove the restricting validation in selecting custom input

### DIFF
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -123,6 +123,7 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 		return confirmations, UnconfirmedStatus
 	}
 }
+
 func GetAllUtoxs(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
 	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
 	if err != nil {
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -124,15 +124,23 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 	}
 }
 
-func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) (utxos []string, total float64, err error) {
+func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) (utxos []string, total int64, err error) {
 	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
 	if err != nil {
 		return nil, 0, err
 	}
 
+	var totalInputAmountDcr float64
 	for _, utxo := range allUtxos {
 		utxos = append(utxos, utxo.OutputKey)
-		total += utxo.Amount.ToCoin()
+		totalInputAmountDcr += utxo.Amount.ToCoin()
 	}
+
+	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
+	if err != nil {
+		return
+	}
+
+	total = int64(totalInputAmount)
 	return
 }
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -123,3 +123,17 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 		return confirmations, UnconfirmedStatus
 	}
 }
+func GetAllUtoxs(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
+	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
+	if err != nil {
+		return nil, 0, err
+	}
+
+	var total float64
+	var utxos []string
+	for _, utxo := range allUtxos {
+		utxos = append(utxos, utxo.OutputKey)
+		total += utxo.Amount.ToCoin()
+	}
+	return utxos, total, nil
+}
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -124,17 +124,15 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 	}
 }
 
-func GetAllUtoxs(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
+func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) (utxos []string, total float64, err error) {
 	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
 	if err != nil {
 		return nil, 0, err
 	}
 
-	var total float64
-	var utxos []string
 	for _, utxo := range allUtxos {
 		utxos = append(utxos, utxo.OutputKey)
 		total += utxo.Amount.ToCoin()
 	}
-	return utxos, total, nil
+	return
 }
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -124,8 +124,8 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 	}
 }
 
-func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) (utxos []string, total int64, err error) {
-	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
+func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int32) (utxos []string, total int64, err error) {
+	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, requiredConfirmations)
 	if err != nil {
 		return nil, 0, err
 	}
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -55,7 +55,7 @@ func WalletBalance(accounts []*Account) string {
 	}
 }
 
-// GetChangeDestinationsWithRandomAmounts generates change destination(s) based on the number of change address the user want
+// GetChangeDestinationsWithRandomAmounts generates change destination(s) based on the number of change addresses the user wants.
 func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, amountInAtom int64, sourceAccount uint32,
 	nUtxoSelection int, sendDestinations []txhelper.TransactionDestination) (changeOutputDestinations []txhelper.TransactionDestination, err error) {
 
@@ -97,20 +97,36 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
 	return
 }
 
-func BuildTxDestinations(destinationAddresses []string, destinationAmounts []string) (destinations []txhelper.TransactionDestination, err error) {
+func BuildTxDestinations(destinationAddresses, destinationAmounts, sendMaxAmountValues []string) (
+	destinations []txhelper.TransactionDestination, totalAmount dcrutil.Amount, err error) {
+
 	for i := range destinationAddresses {
-		var destination []txhelper.TransactionDestination
-		amount, err := strconv.ParseFloat(destinationAmounts[i], 64)
-		if err != nil {
-			return destinations, err
-		}
-		if amount == 0 {
-			continue
-		}
-		destinations = append(destination, txhelper.TransactionDestination{
+		destination := txhelper.TransactionDestination{
 			Address: destinationAddresses[i],
-			Amount:  amount,
-		})
+			// only set SendMax to true if `sendMaxAmountValues` is not nil and this particular sendMaxAmountValue is "true"
+			SendMax: sendMaxAmountValues != nil && sendMaxAmountValues[i] == "true",
+		}
+
+		if !destination.SendMax {
+			var dcrSendAmount dcrutil.Amount
+			destination.Amount, err = strconv.ParseFloat(destinationAmounts[i], 64)
+			if err == nil {
+				dcrSendAmount, err = dcrutil.NewAmount(destination.Amount)
+			}
+
+			if err != nil {
+				err = fmt.Errorf("invalid destination amount: %s", destinationAmounts[i])
+				return
+			}
+			totalAmount += dcrSendAmount
+		}
+
+		if destination.Amount == 0 && !destination.SendMax {
+			err = fmt.Errorf("invalid request, cannot send 0 amount to %s", destination.Address)
+			return
+		}
+
+		destinations = append(destinations, destination)
 	}
 	return
 }
@@ -127,8 +143,8 @@ func TxStatus(txBlockHeight, bestBlockHeight int32) (int32, string) {
 	}
 }
 
-func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int32) (utxos []string, total int64, err error) {
-	allUtxos, err := walletMiddleware.UnspentOutputs(accountNumber, 0, requiredConfirmations)
+func SumUtxosInAccount(wallet Wallet, accountNumber uint32, requiredConfirmations int32) (utxos []string, total dcrutil.Amount, err error) {
+	allUtxos, err := wallet.UnspentOutputs(accountNumber, 0, requiredConfirmations)
 	if err != nil {
 		return nil, 0, err
 	}
@@ -139,11 +155,6 @@ func GetAllUtxos(walletMiddleware Wallet, accountNumber uint32, requiredConfirma
 		totalInputAmountDcr += utxo.Amount.ToCoin()
 	}
 
-	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
-	if err != nil {
-		return
-	}
-
-	total = int64(totalInputAmount)
+	total, err = dcrutil.NewAmount(totalInputAmountDcr)
 	return
 }
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -98,16 +98,19 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
 }
 
 func BuildTxDestinations(destinationAddresses []string, destinationAmounts []string) (destinations []txhelper.TransactionDestination, err error) {
-	destinations = make([]txhelper.TransactionDestination, len(destinationAddresses))
 	for i := range destinationAddresses {
+		var destination []txhelper.TransactionDestination
 		amount, err := strconv.ParseFloat(destinationAmounts[i], 64)
 		if err != nil {
 			return destinations, err
 		}
-		destinations[i] = txhelper.TransactionDestination{
+		if amount == 0 {
+			continue
+		}
+		destinations = append(destination, txhelper.TransactionDestination{
 			Address: destinationAddresses[i],
 			Amount:  amount,
-		}
+		})
 	}
 	return
 }
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -259,6 +259,7 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 
 	req.ParseForm()
 	utxos := req.Form["utxo"]
+
 	totalSelectedInputAmountDcr := req.FormValue("totalSelectedInputAmountDcr")
 	selectedAccount := req.FormValue("source-account")
 	nChangeOutputsStr := req.FormValue("nChangeOutput")
@@ -302,6 +303,22 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 
+	// if no input is selected, then use all inputs.
+	// This is so as to make getting max amount possible for normal sending from the UI
+	if len(utxos) < 1 {
+		requiredConfirmations := walletcore.DefaultRequiredConfirmations
+
+		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
+		if getUnconfirmed == "true" {
+			requiredConfirmations = 0
+		}
+		utxos, totalInputAmountDcr, err = routes.getAllUtoxs(uint32(account), requiredConfirmations)
+		if err != nil {
+			data["error"] = err.Error()
+			return
+		}
+	}
+
 	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
 	if err != nil {
 		data["error"] = err.Error()
@@ -314,6 +331,21 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 	data["message"] = changeOutputDestinations
+}
+
+func (routes Routes) getAllUtoxs(accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
+	allUtxos, err := routes.walletMiddleware.UnspentOutputs(accountNumber, 0, int32(requiredConfirmations))
+	if err != nil {
+		return nil, 0, err
+	}
+
+	var total float64
+	var utxos []string
+	for _, utxo := range allUtxos {
+		utxos = append(utxos, utxo.OutputKey)
+		total += utxo.Amount.ToCoin()
+	}
+	return utxos, total, nil
 }
 
 func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"strconv"
+	"strings"
 
 	"github.com/decred/dcrd/dcrutil"
 	"github.com/go-chi/chi"
@@ -116,7 +117,11 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) 
 	changeAmount, err := txhelper.EstimateChange(len(payload.Utxos), payload.TotalInputAmount, payload.SendDestinations, []string{selectedAddress})
 
 	if err != nil {
-		data["error"] = fmt.Sprintf("Error in getting change amount: %s", err.Error())
+		if strings.Contains(err.Error(), "total input amount not enough to cover transaction") {
+			data["error"] = "Total send amount is already at maximum"
+			return
+		}
+		data["error"] = fmt.Sprintf("Error in getting max send amount: %s", err.Error())
 		return
 	}
 	data["amount"] = dcrutil.Amount(changeAmount).ToCoin()
@@ -229,8 +234,8 @@ func (routes *Routes) getUnspentOutputs(res http.ResponseWriter, req *http.Reque
 
 	requiredConfirmations := walletcore.DefaultRequiredConfirmations
 
-	getUnconfirmed := req.URL.Query().Get("get-unconfirmed")
-	if getUnconfirmed == "true" {
+	spendUnconfirmed := req.URL.Query().Get("spend-unconfirmed")
+	if spendUnconfirmed == "true" {
 		requiredConfirmations = 0
 	}
 
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -125,15 +125,15 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) 
 	}
 
 	// if no input is selected, then use all inputs.
-	// This is so as to make getting max amount possible for normal sending from the UI
+	// This is so as to make change amount possible for situation where custom inputs are not sent
 	if len(utxos) == 0 {
 		requiredConfirmations := walletcore.DefaultRequiredConfirmations
 
-		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
+		getUnconfirmed := req.URL.Query().Get("get-unconfirmed")
 		if getUnconfirmed == "true" {
 			requiredConfirmations = 0
 		}
-		utxos, totalInputAmountDcr, err = walletcore.GetAllUtoxs(routes.walletMiddleware, uint32(account), requiredConfirmations)
+		utxos, totalInputAmountDcr, err = walletcore.GetAllUtxos(routes.walletMiddleware, uint32(account), requiredConfirmations)
 		if err != nil {
 			data["error"] = fmt.Errorf("error fetching unspent outputs for new tx: %s", err.Error())
 			return
@@ -146,6 +146,7 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) 
 		return
 	}
 
+	// the max send amount is the change to one output for the payload
 	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, 1, int64(totalInputAmount), sourceAccount, len(utxos), destinations)
 	if err != nil {
 		data["error"] = err.Error()
@@ -310,7 +311,7 @@ func (routes *Routes) getUnspentOutputs(res http.ResponseWriter, req *http.Reque
 
 	requiredConfirmations := walletcore.DefaultRequiredConfirmations
 
-	getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
+	getUnconfirmed := req.URL.Query().Get("get-unconfirmed")
 	if getUnconfirmed == "true" {
 		requiredConfirmations = 0
 	}
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -81,6 +81,79 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
 	routes.renderPage("send.html", data, res)
 }
 
+func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) {
+	data := map[string]interface{}{}
+	defer renderJSON(data, res)
+
+	req.ParseForm()
+	utxos := req.Form["utxo"]
+
+	totalSelectedInputAmountDcr := req.FormValue("totalSelectedInputAmountDcr")
+	selectedAccount := req.FormValue("source-account")
+
+	account, err := strconv.ParseUint(selectedAccount, 10, 32)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+	sourceAccount := uint32(account)
+
+	destinationAddresses := req.Form["destination-address"]
+	destinationAmounts := req.Form["destination-amount"]
+
+	destinations, err := walletcore.BuildTxDestinations(destinationAddresses, destinationAmounts)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+
+	changeOutputAddreses := req.Form["change-output-address"]
+	changeOutputAmounts := req.Form["change-output-amount"]
+
+	existingChangeDestinations, err := walletcore.BuildTxDestinations(changeOutputAddreses, changeOutputAmounts)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+
+	destinations = append(destinations, existingChangeDestinations...)
+
+	totalInputAmountDcr, err := strconv.ParseFloat(totalSelectedInputAmountDcr, 64)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+
+	// if no input is selected, then use all inputs.
+	// This is so as to make getting max amount possible for normal sending from the UI
+	if len(utxos) == 0 {
+		requiredConfirmations := walletcore.DefaultRequiredConfirmations
+
+		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
+		if getUnconfirmed == "true" {
+			requiredConfirmations = 0
+		}
+		utxos, totalInputAmountDcr, err = walletcore.GetAllUtoxs(routes.walletMiddleware, uint32(account), requiredConfirmations)
+		if err != nil {
+			data["error"] = fmt.Errorf("error fetching unspent outputs for new tx: %s", err.Error())
+			return
+		}
+	}
+
+	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+
+	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, 1, int64(totalInputAmount), sourceAccount, len(utxos), destinations)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+	data["message"] = changeOutputDestinations
+}
+
 func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Request) {
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
@@ -303,22 +376,6 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 
-	// if no input is selected, then use all inputs.
-	// This is so as to make getting max amount possible for normal sending from the UI
-	if len(utxos) == 0 {
-		requiredConfirmations := walletcore.DefaultRequiredConfirmations
-
-		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
-		if getUnconfirmed == "true" {
-			requiredConfirmations = 0
-		}
-		utxos, totalInputAmountDcr, err = walletcore.GetAllUtoxs(routes.walletMiddleware, uint32(account), requiredConfirmations)
-		if err != nil {
-			data["error"] = fmt.Errorf("error fetching unspent outputs for new tx: %s", err.Error())
-			return
-		}
-	}
-
 	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
 	if err != nil {
 		data["error"] = err.Error()
@@ -327,7 +384,6 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 
 	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, int(nChangeOutputs), int64(totalInputAmount), sourceAccount, len(utxos), destinations)
 	if err != nil {
-		// return error key for the frontend to process
 		data["error"] = err.Error()
 		return
 	}
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -105,7 +105,7 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) 
 	selectedAddress := req.FormValue("selected-address")
 
 	// add selected address to destination so the fee for the address will be consider in the estimated change
-	payload.sendDestinations = append(payload.sendDestinations, txhelper.TransactionDestination{Address:selectedAddress})
+	payload.sendDestinations = append(payload.sendDestinations, txhelper.TransactionDestination{Address: selectedAddress})
 
 	if payload.totalSendAmount >= payload.totalInputAmount {
 		data["error"] = "Total send amount is already at maximum"
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -105,11 +105,6 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) 
 		}
 	}
 
-	if payload.totalSendAmount >= payload.totalInputAmount {
-		data["error"] = "Total send amount is already at maximum"
-		return
-	}
-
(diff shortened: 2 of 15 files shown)


web/views/send.html
```diff
@@ -18,12 +18,20 @@
...
                                     <div class="col-sm-12 col-md-6">
                                         <div class="form-group">
-                                            <label for="source-account">Source Account</label>	                                            
-                                            <select data-target="send.sourceAccount" class="form-control" id="source-account" name="source-account" {{ if eq (len .accounts) 1 }}disabled{{ end }}>
-                                            {{ range $account := .accounts }}	                                                
+                                            <label for="source-account">Source Account</label>
+                                        {{ if eq (len .accounts) 1 }}
+                                            {{ range $account := .accounts }}
+                                                <input data-target="send.sourceAccount" value="{{ $account.Number }}" type="hidden" name="source-account" id="source-account" >
+                                                <input type="text" readonly class="form-control" value="{{ accountString $account }}">
```
itswisdomagain: Why do we need 2 inputs?

web/views/send.html
```diff
@@ -18,12 +18,20 @@
                                 <div class="row">
                                     <div class="col-sm-12 col-md-6">
                                         <div class="form-group">
-                                            <label for="source-account">Source Account</label>	                                            
-                                            <select data-target="send.sourceAccount" class="form-control" id="source-account" name="source-account" {{ if eq (len .accounts) 1 }}disabled{{ end }}>
-                                            {{ range $account := .accounts }}	                                                
+                                            <label for="source-account">Source Account</label>
+                                        {{ if eq (len .accounts) 1 }}
```
itswisdomagain: Need some formatting.

itswisdomagain (changes requested):
I'm rethinking the requirement to make the source account amount selectable... If there's just 1 account, we can make this requirement work. But where we have more than 1 account, user can't highlight the text in the dropdown.

I'll recommend sticking to the dropdown pattern for both cases of single and multiple accounts - using `disabled` for single account.

To enable copying amount, we could and should follow the pattern in the mobile apps - have a "Send Max" or "Send All" button. This will also have the added advantage of calculating the fee required for the transaction and subtracting it from the total balance before displaying it on the amount field.

Alternatively, we could have the account balance info displayed in small italics below the account selection widget.

Will need to confer with @raedah for the desired use case and UI presentation.

UPDATE:
- Let's add a Send Max button to each destination address/amount pair (assigns the remaining spendable balance to the destination amount field after subtracting amount set for other destination fields, should be smart enough to update the amount value as other destination amount values are changed)
- No need to display balance separately, we can stick to making the balance text selectable when there's just one account

web/views/send.html
ademuanthony: so as to make the balance highlightable when there is just one account

ademuanthony:
I have implemented the requested changes. The destinations now have a send max button


send-page-send-max

web/routes/handlers.go
```diff
@@ -278,6 +279,22 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 
+	// if no input is selected, then use all inputs.
+	// This is so as to make getting max amount possible for normal sending from the UI
+	if len(utxos) < 1 {
```
itswisdomagain: `len(utxos) == 0` makes more sense

web/routes/handlers.go
```diff
@@ -278,6 +279,22 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
...
+		requiredConfirmations := walletcore.DefaultRequiredConfirmations
+
+		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
+		if getUnconfirmed == "true" {
+			requiredConfirmations = 0
+		}
+		utxos, totalInputAmountDcr, err = routes.getAllUtoxs(uint32(account), requiredConfirmations)
+		if err != nil {
+			data["error"] = err.Error()
```
itswisdomagain: `fmt.Errorf("error fetching unspent outputs for new tx: %s", err.Error()`

web/routes/handlers.go
```diff
@@ -292,6 +309,21 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 	data["message"] = changeOutputDestinations
 }
 
+func (routes Routes) getAllUtoxs(accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
```
itswisdomagain: - This should be in a helper.go file
- The `getUnspentOutputs` may benefit from this function too

itswisdomagain (changes requested):
**Observed error 1**
There's an anomaly with the add and remove destination address buttons:
Only 1 address, yet remove destination address button shows
Screenshot 2019-03-30 at 9 07 03 PM
2 addresses, no remove destination address button
Screenshot 2019-03-30 at 9 07 34 PM

**Observed error 2**
If I use the Send Max button on a destination field and try to use it on another destination field, the following error is presented. The error message does not even appear to be related to the attempted action:
Screenshot 2019-03-30 at 9 07 44 PM

app/walletcore/helper.go
```diff
@@ -110,3 +110,19 @@ func BuildTxDestinations(destinationAddresses []string, destinationAmounts []str
 	}
 	return
 }
+
+
+func GetAllUtoxs(walletMiddleware Wallet, accountNumber uint32, requiredConfirmations int) ([]string, float64, error) {
```
itswisdomagain: - `GetAllUtxos` (typo)
- `wallet Wallet`
- Use named return variables so you don't declare in the function body and the return values will be have more meaning (`utxos []string, total float64, err error`)

web/routes/handlers.go
```diff
@@ -278,6 +279,22 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 
+	// if no input is selected, then use all inputs.
+	// This is so as to make getting max amount possible for normal sending from the UI
```
itswisdomagain: Since this method doesn't just cover send max button click scenarios, this comment will need to be updated.

web/routes/handlers.go
```diff
@@ -286,6 +303,7 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 
 	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, int(nChangeOutputs), int64(totalInputAmount), sourceAccount, len(utxos), destinations)
 	if err != nil {
+		// return error key for the frontend to process
```
itswisdomagain: error key?

web/routes/handlers.go
```diff
@@ -278,6 +279,22 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 		return
 	}
 
+	// if no input is selected, then use all inputs.
+	// This is so as to make getting max amount possible for normal sending from the UI
+	if len(utxos) == 0 {
+		requiredConfirmations := walletcore.DefaultRequiredConfirmations
+
+		getUnconfirmed := req.URL.Query().Get("getUnconfirmed")
```
itswisdomagain: url params should not be camel cased
getUnconfirmed => get-unconfirmed

itswisdomagain (changes requested):
Will review some more after these changes...
Also, 
- create a separate route for the send max button (`/max-send-amount`) that behaves just like the getChangeOutputs route
- refactor the current getChangeOutputs handler into a shared function
- make both routes point to different handlers, then the handlers can call the shared function

web/routes/handlers.go
```diff
@@ -71,6 +71,80 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
 	routes.render("send.html", data, res)
 }
 
+func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request) {
```
itswisdomagain, line 85: Most of the code in this function is repeated in the `getRandomChangeOutputs` handler function. With proper refactoring, the repeated code can be extracted into a helper function (in a helpers.go) file.

web/routes/handlers.go
```diff
@@ -71,6 +71,80 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
+		}
+	}
+
+	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+
+	// the max send amount is the change to one output for the payload
```
itswisdomagain: This is close but not perfect yet. The change amount returned is "tied" to an address. If the address changes, the fee could change and therefore the change amount.

There's need to get the address to which the user wants to send max amount; and then use that address as a "change address" to generate the change amount. The function you'll need is:
```
changeAmount, err := txhelper.EstimateChange(nUtxoSelection, amountInAtom, sendDestinations, changeAddresses)
```
where
- the `sendDestinations` should not include the address for which the max amount is to be calculated and
- the `changeAddresses` is a single slice of the address for which the max amount is to be calculated

This function is used in the `walletcore.GetChangeDestinationsWithRandomAmounts`, more insight may be found there.

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+    if (!this.destinationFieldsValid()) {
+      amountField.value = currentAmount
+      return
+    }
+    this.resetChangeOutput()
+    let _this = this
+    // set the destination amount to zero and get the server calculated value
+    amountField.value = 0
+    this.getMaxSendAmount(changeOutputs => {
+      if (!changeOutputs || changeOutputs.length < 1) {
```
itswisdomagain: Under what conditions could this happen? We should show the user a message explaining why we could not calculate the max amount that can be sent to the address.

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+    const currentAmount = amountField.value
+    // temporarily set the destination amount field (if empty) to make destination validation pass
+    if (!(parseInt(currentAmount) > 0)) {
+      amountField.value = 1
+    }
+    if (!this.destinationFieldsValid()) {
+      amountField.value = currentAmount
+      return
+    }
+    this.resetChangeOutput()
```
itswisdomagain: Shouldn't this be done **_after_** the max amount is retrieved from the server and set to the appropriate amount field? If the amount is not changed, any previously generated change outputs would remain valid.

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+    amountField.value = 0
+    this.getMaxSendAmount(changeOutputs => {
+      if (!changeOutputs || changeOutputs.length < 1) {
+        amountField.value = currentAmount
+        return
+      }
+      amountField.value = changeOutputs[0].Amount
+      _this.calculateCustomInputsPercentage()
+    }, (errMsg) => {
+      if (errMsg.indexOf('total input amount not enough to cover transaction') > -1) {
```
itswisdomagain: 2 things here
1) If the total input amount is not enough to generate another amount for another address (the "change address"), we should show the user a proper message; perhaps: `Total send amount is already at maximum`
2) This if statement feels like a hack... what if the error message returned is different? There should be a better way to determine this scenario on the server and the server should return the proper error message described above; and the error message should always be shown to the user

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+
+  getMaxSendAmount (successCallback, errorCallback) {
+    this.maxSendAmountButtonTargets.forEach(el => {
+      el.setAttribute('disabled', 'disabled')
+    })
+
+    let queryParams = $('#send-form').serialize()
+    queryParams += `&totalSelectedInputAmountDcr=${this.getSelectedInputsSum()}`
+    if (this.spendUnconfirmedTarget.checked) {
+      queryParams += '&get-unconfirmed=true'
```
itswisdomagain: `spend-unconfirmed` is better I believe

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+
+    // add source-account value to post data if source-account element is disabled
+    if (this.sourceAccountTarget.disabled) {
+      queryParams += `&source-account=${this.sourceAccountTarget.value}`
+    }
+
+    let _this = this
+    axios.get('/max-send-amount?' + queryParams)
+      .then((response) => {
+        let result = response.data
```
itswisdomagain, line 203: This block can be simplified:

```
if (!result.error) {
  successCallback(result.message)
} else if (errorCallback) {
  errorCallback(result.error)
} else {
  _this.setErrorMessage(result.error)
}
```

web/static/app/src/controllers/send_controller.js
```diff
@@ -220,7 +307,7 @@ export default class extends Controller {
 
     let _this = this
     this.getRandomChangeOutputs(numberOfChangeOutput, function (changeOutputs) {
-      if (!_this.useCustomChangeOutput) {
+      if (!_this.useCustomChangeOutput || !changeOutputs) {
```
itswisdomagain: Why would there be no change outputs? If that happens, an explanatory error message should be displayed

web/static/app/src/controllers/send_controller.js
```diff
@@ -265,6 +352,9 @@ export default class extends Controller {
 
     let queryParams = $('#send-form').serialize()
     queryParams += `&totalSelectedInputAmountDcr=${this.getSelectedInputsSum()}`
+    if (this.spendUnconfirmedTarget.checked) {
+      queryParams += '&get-unconfirmed=true'
```
itswisdomagain: This query param isn't needed for this endpoint.

web/static/app/src/controllers/send_controller.js
```diff
@@ -278,13 +368,16 @@ export default class extends Controller {
       .then((response) => {
         let result = response.data
         if (result.error !== undefined) {
+          if (errorCallback) {
```
itswisdomagain: This code block can also be simplified. See similar comment above.

web/views/send.html
```diff
@@ -2,195 +2,236 @@
...
+                <div class="collapsible">
+                    <div class="card">
+                        <div class="card-body no-btm-pad">
+                            <!-- from account -->
+                            <div class="row">
+                                <div class="col-sm-12 col-md-6">
+                                    <div class="form-group">
+                                        <label for="source-account">Source Account</label>
+                                    {{ if eq (len .accounts) 1 }}
+                                    {{ range $account := .accounts }}
```
itswisdomagain: Instead of `{{ range }} ... {{ end }}`, can do `{{ $account := index .accounts 0 }}`. Compare balance.html

web/static/app/src/controllers/send_controller.js
```diff
@@ -265,6 +352,9 @@ export default class extends Controller {
 
     let queryParams = $('#send-form').serialize()
     queryParams += `&totalSelectedInputAmountDcr=${this.getSelectedInputsSum()}`
+    if (this.spendUnconfirmedTarget.checked) {
+      queryParams += '&get-unconfirmed=true'
+    }
 
     // add source-account value to post data if source-account element is disabled
     if (this.sourceAccountTarget.disabled) {
```
itswisdomagain: Don't think this is necessary anymore. There is never an instance of a disabled form control. What we have instead is a hidden input and that is serialized, I believe.

web/static/app/src/controllers/send_controller.js
```diff
@@ -29,13 +30,102 @@ export default class extends Controller {
...
+
+    let queryParams = $('#send-form').serialize()
+    queryParams += `&totalSelectedInputAmountDcr=${this.getSelectedInputsSum()}`
+    if (this.spendUnconfirmedTarget.checked) {
+      queryParams += '&get-unconfirmed=true'
+    }
+
+    // add source-account value to post data if source-account element is disabled
+    if (this.sourceAccountTarget.disabled) {
+      queryParams += `&source-account=${this.sourceAccountTarget.value}`
```
itswisdomagain: Don't think this is necessary anymore. There is never an instance of a disabled form control. What we have instead is a hidden input and that is serialized, I believe.

web/views/send.html
```diff
@@ -2,195 +2,236 @@
...
+                                               value="{{ accountString $account }}">
+                                    {{ end }}
+                                    {{ else }}
+                                        <select data-target="send.sourceAccount" class="form-control"
+                                                id="source-account" name="source-account">
+                                        {{ range $account := .accounts }}
+                                            <option value="{{ $account.Number }}">{{ accountString $account }}</option>
+                                        {{ end }}
+                                        </select>
+                                    {{ end }}
```
itswisdomagain, line 37: Line space should be after the closing </div>

web/views/send.html
```diff
@@ -2,195 +2,236 @@
...
-                                            <input data-action="click->send#toggleSpendUnconfirmed" data-target="send.spendUnconfirmed" type="checkbox" class="form-check-input" name="spend-unconfirmed" id="spend-unconfirmed">
-                                            <label class="form-check-label" for="spend-unconfirmed">Spend unconfirmed</label>
-                                        </div>
+                                    </div>
+                                    <div class="form-group form-check">
+                                        <input data-action="click->send#toggleSpendUnconfirmed"
+                                               data-target="send.spendUnconfirmed" type="checkbox"
+                                               class="form-check-input" name="spend-unconfirmed" id="spend-unconfirmed">
+                                        <label class="form-check-label" for="spend-unconfirmed">Spend
+                                            unconfirmed</label>
```
itswisdomagain: "Spend" and "unconfirmed" can be in one line.

web/views/send.html
```diff
@@ -2,195 +2,236 @@
...
-                                    <button id="add-destinatvion-btn" type="button" data-action="click->send#newDestination" class="btn btn-sm btn-outline-secondary btn-primary">
-                                        Add another address
-                                    </button> &nbsp;&nbsp; 
-                                    <button data-target="send.removeDestinationButton" data-action="click->send#removeDestination" id="remove-destination-btn" type="button" class="btn btn-sm btn-outline-secondary btn-primary d-none">
-                                        Remove Last address
-                                    </button>
                                 </div>
+                            </template>
+                            <div class="form-group">
+                                <button id="add-destinatvion-btn" type="button" data-action="click->send#newDestination"
```
itswisdomagain: button id is unnecessary

web/views/send.html
```diff
@@ -2,195 +2,236 @@
...
+                                        class="btn btn-sm btn-outline-secondary btn-primary">
+                                    Add another address
+                                </button> &nbsp;&nbsp;
+                                <button data-target="send.removeDestinationButton"
+                                        data-action="click->send#removeDestination" id="remove-destination-btn"
```
itswisdomagain: button id is unnecessary

itswisdomagain (changes requested):

web/routes/handlers.go
```diff
@@ -71,6 +71,80 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
+	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, 1, int64(totalInputAmount), sourceAccount, len(utxos), destinations)
+	if err != nil {
+		data["error"] = err.Error()
+		return
+	}
+	data["message"] = changeOutputDestinations
```
itswisdomagain: - message?
- also, we only need the amount...

itswisdomagain (changes requested):

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	SourceAccount uint32
+	Passphrase string
+	RequiredConfirmations int32
+	UseCustom bool
+	SendDestinations []txhelper.TransactionDestination
+	ChangeDestinations []txhelper.TransactionDestination
+	TotalInputAmount int64
+
+}
+
```
itswisdomagain, line 25: Can remove one empty line

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+type sendPagePayload struct {
+	Utxos []string
+	SourceAccount uint32
+	Passphrase string
+	RequiredConfirmations int32
+	UseCustom bool
+	SendDestinations []txhelper.TransactionDestination
+	ChangeDestinations []txhelper.TransactionDestination
+	TotalInputAmount int64
+
```
itswisdomagain: Empty line can go

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	"net/http"
+	"strconv"
+
+	"github.com/decred/dcrd/dcrutil"
+	"github.com/raedahgroup/dcrlibwallet/txhelper"
+	"github.com/raedahgroup/godcr/app/walletcore"
+)
+
+type sendPagePayload struct {
+	Utxos []string
```
itswisdomagain: These properties are only used within this package, I believe. No need to make them public

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	payload.Passphrase = req.FormValue("wallet-passphrase")
+	spendUnconfirmed := req.FormValue("spend-unconfirmed")
+	payload.UseCustom = req.FormValue("use-custom") != ""
+
+	destinationAddresses := req.Form["destination-address"]
+	destinationAmounts := req.Form["destination-amount"]
+
+	sendDestinations, err := walletcore.BuildTxDestinations(destinationAddresses, destinationAmounts)
+	if err != nil {
+		return nil, err
```
itswisdomagain: Can just `return`, same for line 46 below

web/routes/handlers.go
```diff
@@ -71,78 +73,66 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
 
-	req.ParseForm()
-	utxos := req.Form["utxo"]
-	selectedAccount := req.FormValue("source-account")
-	passphrase := req.FormValue("wallet-passphrase")
-	spendUnconfirmed := req.FormValue("spend-unconfirmed")
-	useCustom := req.FormValue("use-custom")
+	payload, err := retrieveSendPagePayload(req)
```
itswisdomagain, line 89: `payload` may be nil and `err` may need to be displayed

web/routes/handlers.go
```diff
@@ -71,78 +73,66 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
 
-	destinationAddresses := req.Form["destination-address"]
-	destinationAmounts := req.Form["destination-amount"]
+	// if no input is selected, then use all inputs.
+	// This is so as to make set max amount possible for situations where custom inputs are not sent
+	if len(payload.Utxos) == 0 {
+		payload.Utxos, payload.TotalInputAmount, err = walletcore.GetAllUtxos(routes.walletMiddleware, payload.SourceAccount, payload.RequiredConfirmations)
+		if err != nil {
+			data["error"] = fmt.Errorf("error fetching unspent outputs for new tx: %s", err.Error())
```
itswisdomagain: This error isn't helpful. User clicks "Send Max" and sees error fetching unspent outputs;  that would be confusing.

web/routes/handlers.go
```diff
@@ -71,78 +73,66 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
+			return
+		}
+	}
+
+	selectedAddress := req.FormValue("selected-address")
+	for i := 0; i < len(payload.SendDestinations); i++ {
+		if payload.SendDestinations[i].Address == selectedAddress {
```
itswisdomagain: Would need another way to determine which address is selected because there's a new requirement to allow addresses to be repeated in the send form.

web/routes/handlers.go
```diff
@@ -71,78 +73,66 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
+			break
+		}
+	}
+
+	changeAmount, err := txhelper.EstimateChange(len(payload.Utxos), payload.TotalInputAmount, payload.SendDestinations, []string{selectedAddress})
 
-	sendDestinations, err := walletcore.BuildTxDestinations(destinationAddresses, destinationAmounts)
 	if err != nil {
-		data["error"] = err.Error()
+		if strings.Contains(err.Error(), "total input amount not enough to cover transaction") {
```
itswisdomagain: If you look at the implementation of `txhelper.EstimateChange`, this error occurs when `totalInputAmount < totalSendAmount`. You can sum the total send amount and check if the input amount can cover it before calling `txhelper.EstimateChange`. Then you won't need to do a string comparison on the error returned from that function.

web/routes/handlers.go
```diff
@@ -71,78 +73,66 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
+}
+
+func (routes *Routes) submitSendTxForm(res http.ResponseWriter, req *http.Request) {
+	data := map[string]interface{}{}
+	defer renderJSON(data, res)
 
-	account, err := strconv.ParseUint(selectedAccount, 10, 32)
+	payload, err := retrieveSendPagePayload(req)
 	if err != nil {
 		data["error"] = err.Error()
```
itswisdomagain, line 122: The `err` returned from `retrieveSendPagePayload` doesn't have accompanying context. Need to add that here before sending to client. `fmt.Sprintf("Error reading values from form submitted form: %s", err.Error())`

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+	payload.SendDestinations = sendDestinations
+
+	account, err := strconv.ParseUint(selectedAccount, 10, 32)
+	if err != nil {
+		return nil, err
```
itswisdomagain: Do not return the direct error from parse failures; it's usually not presentable. Instead return `fmt.Errorf(Invalid source account selected)`

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+
+	payload.RequiredConfirmations = walletcore.DefaultRequiredConfirmations
+	if spendUnconfirmed != "" {
+		payload.RequiredConfirmations = 0
+	}
+
+	totalSelectedInputAmountDcr := req.FormValue("totalSelectedInputAmountDcr")
+	totalInputAmountDcr, err := strconv.ParseFloat(totalSelectedInputAmountDcr, 64)
+	if err != nil {
+		return
```
itswisdomagain: Do not return the direct error from parse failures; it's usually not presentable. Instead 
```
err = `fmt.Errorf(Cannot read total send amount value)`
return
```

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
+	if err != nil {
+		return
```
itswisdomagain:
```
err = `fmt.Errorf(Invalid total send amount value)`
return
```

web/routes/helpers.go
```diff
@@ -0,0 +1,78 @@
...
+	}
+
+	payload.TotalInputAmount = int64(totalInputAmount)
+
+	changeOutputAddresses := req.Form["change-output-address"]
+	changeOutputAmounts := req.Form["change-output-amount"]
+
+	changeDestinations, err := walletcore.BuildTxDestinations(changeOutputAddresses, changeOutputAmounts)
+	if err != nil {
+		return
```
itswisdomagain: Confirm if `walletcore.BuildTxDestinations` returns proper error messages, otherwise, add some extra information here before returning.

web/static/app/src/controllers/send_controller.js
```diff
@@ -197,6 +274,10 @@ export default class extends Controller {
   toggleCustomChangeOutputsVisibility () {
     this.clearMessages()
     this.useCustomChangeOutput = !this.useCustomChangeOutput
+    this.resetChangeOutput()
+  }
+
+  resetChangeOutput () {
```
itswisdomagain, line 396: I think this should also reset the `numberOfChangeOutputs` value, so a user does not think that a previously entered value is still valid.

itswisdomagain (changes requested):

web/static/app/src/controllers/send_controller.js
ademuanthony: yes it does

web/routes/handlers.go
```diff
@@ -3,14 +3,14 @@ package routes
...
-	"strconv"
-
 	"github.com/decred/dcrd/dcrutil"
 	"github.com/go-chi/chi"
 	"github.com/raedahgroup/dcrlibwallet"
+	"github.com/raedahgroup/dcrlibwallet/txhelper"
 	"github.com/raedahgroup/godcr/app/walletcore"
 	"github.com/skip2/go-qrcode"
+	"net/http"
+	"strconv"
```
itswisdomagain: Re-arrange imports

web/routes/handlers.go
```diff
@@ -79,78 +79,82 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
-	useCustom := req.FormValue("use-custom")
-
-	destinationAddresses := req.Form["destination-address"]
-	destinationAmounts := req.Form["destination-amount"]
-
-	sendDestinations, err := walletcore.BuildTxDestinations(destinationAddresses, destinationAmounts)
+	payload, err := retrieveSendPagePayload(req)
 	if err != nil {
-		data["error"] = err.Error()
+		data["error"] = fmt.Sprintf("cannot get max amount: %s", err.Error())
```
itswisdomagain: This can start with a capital letter, only `fmt.Errorf` has that restriction. Do same for other similar occurences.

web/routes/handlers.go
```diff
@@ -79,78 +79,82 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
-	sourceAccount := uint32(account)
 
-	var requiredConfirmations int32 = walletcore.DefaultRequiredConfirmations
-	if spendUnconfirmed != "" {
-		requiredConfirmations = 0
+	selectedAddress := req.FormValue("selected-address")
+	for i := 0; i < len(payload.sendDestinations); i++ {
+		// the amount field of the selected address is set to 0 from the frontend
+		if payload.sendDestinations[i].Address == selectedAddress && payload.sendDestinations[i].Amount == 0 {
+			payload.sendDestinations = append(payload.sendDestinations[:i], payload.sendDestinations[i+1:]...)
```
itswisdomagain: Instead of doing this, modify the `retrieveSendPagePayload` to ignore 0 amount destinations, as even a normal send operation will fail with a 0 amount destination.

web/routes/handlers.go
```diff
@@ -79,78 +79,82 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
-		changeOutputAddreses := req.Form["change-output-address"]
-		changeOutputAmounts := req.Form["change-output-amount"]
-
-		changeDestinations, err := walletcore.BuildTxDestinations(changeOutputAddreses, changeOutputAmounts)
+	var totalSendAmount int64
+	for _, destination := range payload.sendDestinations {
+		amountInAtom, err := txhelper.AmountToAtom(destination.Amount)
 		if err != nil {
-			data["error"] = err.Error()
+			data["error"] = fmt.Errorf("cannot get max amount, error in converting input amount to atom: %s", err.Error())
```
itswisdomagain: - No need to return the actual `err.Error`, it won't be human readable
- Also, it's not input amount you're converting but a destination amount...

web/routes/handlers.go
```diff
@@ -79,78 +79,82 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
 			return
 		}
+		totalSendAmount += amountInAtom
+	}
 
-		if len(changeDestinations) < 1 {
-			// add at-least one change output
-			totalSelectedInputAmountDcr := req.FormValue("totalSelectedInputAmountDcr")
+	if payload.totalInputAmount < totalSendAmount {
```
itswisdomagain: Change to `totalSendAmount >= payload.totalInputAmount`, makes the next line easier to understand: the total send amount is greater than the total input amount, therefore, send amount is at maximum.

web/routes/handlers.go
```diff
@@ -241,58 +245,24 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
...
 		data["error"] = err.Error()
 		return
 	}
 
-	totalInputAmount, err := dcrutil.NewAmount(totalInputAmountDcr)
-	if err != nil {
-		data["error"] = err.Error()
-		return
-	}
+	destinations := append(payload.sendDestinations, payload.changeDestinations...)
```
itswisdomagain: If we're trying to get change outputs, why are we adding previously set change destinations to the actual send destinations? This will return a wrong change address. I think you should only use the actual send destinations so you can get the actual change amount which you can then randomize.

PS: I noticed that the frontend clears the previously generated change destinations before calling this endpoint but that's even the more reason why we shouldn't do this.

web/routes/handlers.go
```diff
@@ -241,58 +245,24 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
...
 
-	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, int(nChangeOutputs), int64(totalInputAmount), sourceAccount, len(utxos), destinations)
+	changeOutputDestinations, err := walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, int(nChangeOutputs), payload.totalInputAmount, payload.sourceAccount, len(payload.utxos), destinations)
```
itswisdomagain: - This is a long line, please split
- Before calling this function, also check `if totalSendAmount >= payload.totalInputAmount`, because the error returned is not very helpful. It says `error in getting change amount: total input amount not enough to cover transaction fee`, message should be `error in getting change amount: total input amount cannot cover total send amount and transaction fee`
- You may need to modify the `retrieveSendPagePayload` to also return `totalSendAmount`

web/routes/helpers.go
```diff
@@ -0,0 +1,79 @@
+package routes
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
```
itswisdomagain: Wrong import, goland can be dumb sometimes.

itswisdomagain (changes requested):

itswisdomagain:
2 issues just surfaced while trying to use the Send Max button with a destination address
- if I select custom inputs, I get a success message with hash undefined, the tx didn't really go through
- if I don't select custom inputs, I get insufficient balance error

I think there's need to re-look at how we're getting the max send amount. This feature already exists in dcrandroid, we should get some insight from there.

ademuanthony:
> 2 issues just surfaced while trying to use the Send Max button with a destination address
> 
> * if I select custom inputs, I get a success message with hash undefined, the tx didn't really go through
> * if I don't select custom inputs, I get insufficient balance error
> 
> I think there's need to re-look at how we're getting the max send amount. This feature already exists in dcrandroid, we should get some insight from there.

dcrandroid only set spendable as max amount without fee considerable and there does seem to be multiple address support

itswisdomagain:
These issues are still there. Does it work on your end?

> 2 issues just surfaced while trying to use the Send Max button with a destination address
> 
> * if I select custom inputs, I get a success message with hash undefined, the tx didn't really go through
> * if I don't select custom inputs, I get insufficient balance error
> 
> I think there's need to re-look at how we're getting the max send amount. This feature already exists in dcrandroid, we should get some insight from there.

web/routes/handlers.go
```diff
@@ -101,13 +101,9 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request)
...
-	for i := 0; i < len(payload.sendDestinations); i++ {
-		// the amount field of the selected address is set to 0 from the frontend
-		if payload.sendDestinations[i].Address == selectedAddress && payload.sendDestinations[i].Amount == 0 {
-			payload.sendDestinations = append(payload.sendDestinations[:i], payload.sendDestinations[i+1:]...)
-			break
-		}
-	}
+
+	// add selected address to destination so the fee for the address will be consider in the estimated change
+	payload.sendDestinations = append(payload.sendDestinations, txhelper.TransactionDestination{Address:selectedAddress})
```
itswisdomagain: This doesn't work, neither should it. I'm working on a fix.

itswisdomagain:
Made changes to
- the max amount estimation code
- normal send and utxo send with dcrlibwallet (to account for max change amount recipient, this isn't done for dcrwallet rpc yet)
- error messages display

Haven't been able to confirm if sending max amount works now, but it probably should. Here's the new look:
Screen Shot 2019-04-07 at 11 34 47 PM

ademuanthony:
New update

send-remove-destination-button

ademuanthony:
Added transaction summary to the passphrase modal and worked on the other requested changes
send-with-confirm

ademuanthony:
send-page-14-4-empty

send-page-14-4-with-pass

web/routes/handlers.go
```diff
@@ -255,58 +238,28 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
...
-	changeOutputAmounts := req.Form["change-output-amount"]
-
-	existingChangeDestinations, err := walletcore.BuildTxDestinations(changeOutputAddreses, changeOutputAmounts)
-	if err != nil {
-		data["error"] = err.Error()
-		return
-	}
-
-	destinations = append(destinations, existingChangeDestinations...)
+	req.ParseForm()
```
itswisdomagain: This is already called in `retrieveSendPagePayload`, no longer necessary here

web/routes/templates.go
```diff
@@ -45,7 +46,17 @@ func templateFuncMap() template.FuncMap {
...
 		},
 		"accountString": func(account *walletcore.Account) string {
-			return account.String()
+			unconfirmedBalance := account.Balance.Total - account.Balance.Spendable
+			return fmt.Sprintf("%s %s (unconfirmed %s)", account.Name,
+				walletcore.NormalizeBalance(account.Balance.Total.ToCoin()), walletcore.NormalizeBalance(unconfirmedBalance.ToCoin()))
+		},
+		"haveUnconfirmedBalance": func(accounts []*walletcore.Account) bool {
+			for _, account := range accounts {
+				if account.Balance.Total != account.Balance.Spendable {
```
itswisdomagain: `Total` might be different from `Spendable` but doesn't necessarily mean there is unconfirmed balance that can be spent. Some funds might be locked by tickets. Should check `if account.Balance.Unconfirmed > 0`

web/views/send.html
```diff
@@ -1,197 +1,267 @@
...
+                                        <select data-target="send.sourceAccount" class="form-control"
+                                                id="source-account" name="source-account">
+                                        {{ range $account := .accounts }}
+                                            <option value="{{ $account.Number }}">{{ accountString $account }}</option>
+                                        {{ end }}
+                                        </select>
+                                    {{ end }}
+                                    </div>
+
+                                    <div class="form-group form-check mb-0 {{ if haveUnconfirmedBalance .accounts }} d-none {{ end }}">
```
itswisdomagain: This says the checkbox should be hidden if there is unconfirmed balance. The reverse should be the case. Should modify the function to `if noUnconfirmedBalance`.

Also, this puts a lot of spaces between `d-none` in the class attribute. Some spaces can be removed.

Use class `mt-1 mb-0` for margin on this element so that when it is visible, there'd be a space between it and the previous element. Only when it is visible.

web/views/send.html
```diff
@@ -1,197 +1,267 @@
...
+                        <div class="card-body no-btm-pad">
+                            <h5 class="card-title">Sending Decred</h5>
+                        </div>
+                    </div>
+                    <!-- from account and spend unconfirmed checkbox -->
+                    <div class="card">
+                        <div class="card-body">
+                            <div class="row">
+                                <div class="col-sm-12 col-md-6">
+                                    <div class="form-group mb-1">
```
itswisdomagain: Should be `mb-0`, margin should only be present if the next element (spend unconfirmed is visible). See previous comment.

web/views/send.html
```diff
@@ -1,197 +1,267 @@
...
+                                    <thead>
+                                    <tr>
+                                        <th></th>
+                                        <th>Address</th>
+                                        <th>Amount</th>
+                                        <th>Time</th>
+                                        <th>Confirmations</th>
+                                    </tr>
+                                    </thead>
+                                    <tbody data-target="send.customInputsTable"></tbody>
```
itswisdomagain, line 75: When adding utxos to this table, the code sets a utxo selected event listener that calculates the custom inputs percentage. Let's modify that a bit. The utxo checkbox click event listener in line 276 of `send_controller.js` should not directly trigger `calculateCustomInputsPercentage`. Instead it should trigger the following function:

```
  utxoSelectedOrDeselected () {
    this.calculateCustomInputsPercentage()
    this.updateMaxAmountFieldIfSet()
  }
```

The goal is to re-calculate the max amount if there is a change in utxo selection. We cannot trigger this max amount recalculation inside the `calculateCustomInputsPercentage` function because that function is also called after the max amount is calculated and that would lead to circular function calls.

web/routes/handlers.go
```diff
@@ -81,78 +82,60 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
-			}
-
-			changeDestinations, err = walletcore.GetChangeDestinationsWithRandomAmounts(routes.walletMiddleware, 1, int64(totalInputAmount),
-				sourceAccount, len(utxos), sendDestinations)
-			if err != nil {
-				data["error"] = err.Error()
-				return
-			}
-		}
+	if payload.totalSendAmount >= payload.totalInputAmount {
```
itswisdomagain: This if check is no longer necessary. The new dcrlibwallet `txhelper.EstimateMaxSendAmount` function returns a pretty okay error message that can be displayed to the user such as

`Error in estimating max send amount: total send amount (20 DCR) is higher than or equal to the total input amount (20 DCR)`
and
`Error in estimating max send amount: total send amount (20 DCR) is higher than or equal to the total input amount (9 DCR)`

web/views/send.html
```diff
@@ -1,197 +1,267 @@
...
-                                            <input data-action="click->send#toggleSpendUnconfirmed" data-target="send.spendUnconfirmed" type="checkbox" class="form-check-input" name="spend-unconfirmed" id="spend-unconfirmed"
-                                            {{ if .spendUnconfirmedFunds }} checked {{ end }}>
-                                            <label class="form-check-label" for="spend-unconfirmed">Spend unconfirmed</label>
-                                        </div>
+<body data-controller="send">
+<div class="body">
+{{ template "header" .connectionInfo }}
+    <div class="content">
+        <div class="container">
+            <div data-target="send.errorMessage" class="alert alert-danger d-none"></div>
```
itswisdomagain: I think both error and success message boxes should go down to where we have the next button. If I try to send a tx by clicking send and there's an error, I have to scroll back up to see the error message. Same thing if the send is successful. I can't immediately tell if there was an error or success until I scroll up

web/static/app/src/controllers/send_controller.js
```diff
@@ -47,35 +183,55 @@ export default class extends Controller {
...
     }
+
+    const targetElement = event.currentTarget
+    const index = parseInt(targetElement.getAttribute('data-index'))
+
+    this.destinationsTarget.removeChild(this.destinationsTarget.querySelector(`div.destination[data-index="${index}"]`))
+    this.destinationCount--
   }
 
   toggleSpendUnconfirmed () {
```
itswisdomagain: - re-order the functions on this file so that `toggleSpendUnconfirmed`, `toggleUseCustom` and other input selection related functions are displayed at the top
- when use custom is selected, check if there is a send max destination set. If so, reset and hide the change outputs customization section.
- similarly, when a destination's send max checkbox is checked and use custom is checked, reset and hide the change outputs customization section.

itswisdomagain (changes requested):
Something else I noticed is that when you generate a high (5+) number of change outputs, the total is either less than or more than 100% and you get this kind of error when you want to send: `Total change amount percentage must be equal to 100. Current total is 101`. This doesn't happen for manual change outputs even if the number of outputs is high.

yaowork (changes requested):
I'm not able to send funds. Seems my passphrase isn't included in the form data sent to the server as seen from chrome inspector. 
Screenshot from 2019-04-17 10-14-17

yaowork (changes requested):
The remove address/amount pair button should be hidden for the first input pair if there's only one pair on the form, and should only be visible if there's more than one input pairs.

Screenshot from 2019-04-17 10-30-59

web/routes/templates.go
```diff
@@ -47,7 +48,17 @@ func templateFuncMap() template.FuncMap {
 			return
 		},
 		"accountString": func(account *walletcore.Account) string {
-			return account.String()
+			unconfirmedBalance := account.Balance.Total - account.Balance.Spendable
```
itswisdomagain: - `unconfirmedBalance` is not `account.Balance.Total - account.Balance.Spendable` but rather `account.Balance.Unconfirmed`
- shouldn't you only show unconfirmed balance if it is greater that zero?

itswisdomagain (changes requested):

PR #233, opened 2019-03-21
rearrange balance page on the web

ademuanthony:
Resolves [Issue 230](https://github.com/raedahgroup/godcr/issues/230)
reduce the font size and conditionally show spendable balance on second line
balance-page


ademuanthony:
Following @raedah's suggestion, the page is now modified to have the following look
balance-page-mod

app/walletcore/helper.go
```diff
@@ -49,6 +49,30 @@ func SimpleBalance(balance *Balance, detailed bool) string {
 	}
 }
 
+func SpendableBalance(balance *Balance) string {
+	return balance.Spendable.String()
```
itswisdomagain: Don't see a need for this `SpendableBalance` function, can always simply invoke `balance.Spendable.String()` where necessary instead of having this wrapper function

app/walletcore/helper.go
```diff
@@ -49,6 +49,30 @@ func SimpleBalance(balance *Balance, detailed bool) string {
...
+}
+
+// NormalizeBalance adds 0s to the left and the right of balanceString to make it xxxx.xxxxxxxx DCR
+func NormalizeBalance(balanceString string) string {
```
itswisdomagain: - Make this function take the balance as float (`balance.ToCoin()`)
- `return fmt.Sprintf("%013.8f", balance)`

xxxx.xxxxxxxx = 13 characters, 4 before the decimal point, 8 characters after plus the decimal point character = 13
`%013.8f` says use '0' for leading and trailing padding, there should be a minimum of 13 characters and 8 of those should appear after the decimal point.

web/views/balance.html
```diff
@@ -19,8 +19,16 @@ <h5 class="card-title">
 
                         <!-- if just one account and not detailed, show simple balance -->
                         {{ if and (not .detailed) (eq (len .accounts) 1) }}
-                        {{ $account := index .accounts 0 }}
-                        <h3 class="mb-0">{{ simpleBalance $account.Balance false }}</h3>
+                        <div class="row">
+                            <div class="col-md-3 col-sm-6">
+                            {{ $account := index .accounts 0 }}
+                                <p style="font-size: 16px;" class="mb-0"><b>Total</b>: &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
```
itswisdomagain: Maybe use a table without borders to ensure the balance texts on the right are aligned; or use html tab escape character; or use `<span>` with `display: inline-block` and a set fixed width for the "Total" and "Spendable" texts, so that the actual balance texts in both cases begin at the same point to the right of the spans

itswisdomagain (changes requested):

ademuanthony:
I have worked on the requested changes

blance-page-use-table

web/views/balance.html
```diff
@@ -19,8 +19,29 @@ <h5 class="card-title">
...
+                                        <td style="text-align: right;">{{ simpleBalance $account.Balance true }}</td>
+                                    </tr>
+                                {{ if ne $account.Balance.Total $account.Balance.Spendable }}
+                                    <tr>
+                                        <td><b>Spendable:</b></td>
+                                        <td style="text-align: right;">{{ spendableBalance $account.Balance }}</td>
+                                    </tr>
+                                {{ end }}
+                                </table>
+                            {{/*{{ $account := index .accounts 0 }}
```
itswisdomagain: Please take off this commented code

web/views/balance.html
```diff
@@ -19,8 +19,29 @@ <h5 class="card-title">
 
                         <!-- if just one account and not detailed, show simple balance -->
                         {{ if and (not .detailed) (eq (len .accounts) 1) }}
-                        {{ $account := index .accounts 0 }}
-                        <h3 class="mb-0">{{ simpleBalance $account.Balance false }}</h3>
+                        <div class="row">
+                            <div class="col-md-3 col-sm-6">
+                            {{ $account := index .accounts 0 }}
```
itswisdomagain: Move this to after line 21, it's easier to make sense of this declaration when it is placed just after the if statement:
If it's not detailed and it's one account, then declare $account

web/views/balance.html
```diff
@@ -19,8 +19,29 @@ <h5 class="card-title">
...
+                                <table class="balance-table">
+                                    <tr>
+                                        <td><b>Total:</b></td>
+                                        <td style="text-align: right;">{{ simpleBalance $account.Balance true }}</td>
+                                    </tr>
+                                {{ if ne $account.Balance.Total $account.Balance.Spendable }}
```
itswisdomagain: Indent this line

web/views/balance.html
```diff
@@ -19,8 +19,29 @@ <h5 class="card-title">
...
+                                    <tr>
+                                        <td><b>Spendable:</b></td>
+                                        <td style="text-align: right;">{{ spendableBalance $account.Balance }}</td>
+                                    </tr>
+                                {{ end }}
```
itswisdomagain: Indent this line

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #245, opened 2019-03-26
Auto-launch browser for http mode

matthewjamesr:
Gents,

This PR implements auto-launching of user's default browser if http server starts successfully.


web/server.go
```diff
@@ -141,6 +142,8 @@ func startServer(ctx context.Context, address string, router chi.Router) error {
 		return ctx.Err()
 	case <-t.C:
 		fmt.Printf("Web server running on %s\n", address)
+		fmt.Printf("Launching browser....\n")
```
ademuanthony: `fmt.Println("Launching browser....")`

web/server.go
```diff
@@ -141,6 +142,8 @@ func startServer(ctx context.Context, address string, router chi.Router) error {
...
+		browser.OpenURL("http://" + address)
 		return nil
```
ademuanthony, line 112: `fmt.Println("Ready")`
`return nil`

ademuanthony (dismissed):
few changes needed

web/server.go
```diff
@@ -11,6 +11,7 @@ import (
 	"time"
 
 	"github.com/go-chi/chi"
+	"github.com/pkg/browser"
```
itswisdomagain: We aim to use external libraries only if the solution provided cannot be quite easily accomplished with standard go libraries. For this use case, we don't need an external library. See this [SO answer](https://stackoverflow.com/a/39324149/6086027) and this [GH gist](https://gist.github.com/hyg/9c4afcd91fe24316cbf0).

itswisdomagain (changes requested):

itswisdomagain (approved):

ademuanthony (approved):

Issue #238, opened 2019-03-23
Wallet Selection Value Error


matthewjamesr:
Godcr detects multiple wallets, but will not allow selection of one.

```
C:\Users\me\go\src\github.com\raedahgroup\godcr>godcr detect
Select default wallet
 [1]: mainnet wallet from decredition
 [2]: testnet3 wallet from decredition
=> 2
invalid selection, select a number between 1 and 2
```

itswisdomagain:
This is fixed in #248

PR #248, opened 2019-03-28
fix: cli - read input from terminal

### LINKED ISSUE #220: problem with creating wallet on window cmd
error-creating-wallet-windows

itswisdomagain:
Fixes #220 and #238. Issue is only known to affect Windows OS. User input gotten from the terminal/command prompt contains line break characters that ought to be trimmed off.


ademuanthony (approved):
tested okay

PR #251, opened 2019-04-01
[WIP] Unify terminal colour to align with other godcr interfaces

Sirmorrison:


itswisdomagain:
The bulk of the work done is here is finalized in #259

PR #252, opened 2019-04-01
Overview page - web

### LINKED ISSUE #241: Overview page
The balance page should be replaced with an overview page that shows total balance, recent activities and links to frequently used features

ademuanthony:
Resolves #241 
Added overview page to the web with balance, recent transactions and common navigation buttons

overview-page-web


web/routes/handlers.go
```diff
@@ -51,11 +51,18 @@ func (routes *Routes) balancePage(res http.ResponseWriter, req *http.Request) {
 	req.ParseForm()
 	showDetails := req.FormValue("detailed") != ""
 
+	txns, _, err := routes.walletMiddleware.TransactionHistory(routes.ctx, -1, 5)
+	if err != nil {
+		routes.renderError(fmt.Sprintf("Error fetching recent activities: %s", err.Error()), res)
```
itswisdomagain: - `recent **activity**`
- also, no need to render an error if the account balance was retrieved successfully, instead show the balance, then show the error message just below the "Recent Activity" header on the page

web/views/overview.html
```diff
@@ -66,6 +66,55 @@ <h5 class="card-title">
                         </tbody>
                     </table>
                 {{ end }}
+
+                    <h5 class="mt-5">Recent Activities</h5>
```
itswisdomagain: Use `h4` too; and `Recent **Activity**`

web/views/overview.html
```diff
@@ -66,6 +66,55 @@ <h5 class="card-title">
...
+                                <td>{{ .Direction }}</td>
+                                <td>{{ .Type }}</td>
+                                <td><a href="/transaction-details/{{ .Hash }}" >{{ .Hash }}</a></td>
+                            </tr>
+                            {{ end }}
+                            </tbody>
+                        </table>
+                    {{ end }}
+
+                    <div class="row">
```
itswisdomagain: @raedah says these buttons are not necessary. We do already have the nav menu just above...

itswisdomagain (changes requested):

web/views/overview.html
```diff
@@ -66,6 +66,38 @@ <h5 class="card-title">
                         </tbody>
                     </table>
                 {{ end }}
+
+                    <h4 class="mt-5">Recent Activities</h4>
+                    <div class="alert-danger"><p>{{.loadTransactionErr}}</p></div>
```
itswisdomagain, line 72: Wrap this div with `{{ if . loadTransactionErr }}`, then the rest should be in else. No need to show error and say no transaction found.

itswisdomagain (changes requested):
Update the screenshot too

ademuanthony:
Updated to only show transactions if there is no error while loading tnxs

overview-page-error

itswisdomagain (approved):

PR #236, opened 2019-03-22
Nuklear layout

yaowork:
This is a general UI cleanup for the nuklear interface. It 
- Updates the menu items, closes #260 
- Changes the color and size of buttons 
- Uses static sizes instead of fluid sizes for widgets 
- Cleans up the page headers 
- Maintains UI consistency across pages

Screen Shot 2019-04-12 at 12 57 28 AM 1
Screen Shot 2019-04-12 at 12 48 23 AM
Screen Shot 2019-04-12 at 12 45 33 AM
Screen Shot 2019-04-12 at 12 45 24 AM
Screen Shot 2019-04-12 at 12 45 56 AM
Screen Shot 2019-04-12 at 12 46 10 AM
Screen Shot 2019-04-12 at 12 53 57 AM
Screen Shot 2019-04-12 at 12 54 09 AM

PR #258, opened 2019-04-03
Account selection widget

yaowork:
This creates a reusable account selection widget. The widget displays as just text when the wallet has just one account, and displays a combo if the wallet has multiple accounts.

It also auto generates a receive address on the receive page if the wallet has just one account
Screenshot from 2019-04-03 09-52-43
Screenshot from 2019-04-03 09-52-59

### DIFF
--- a/nuklear/handlers/receive.go
+++ b/nuklear/handlers/receive.go
@@ -35,7 +35,7 @@ func (handler *ReceiveHandler) BeforeRender() {
 func (handler *ReceiveHandler) Render(window *nucular.Window, wallet walletcore.Wallet) {
 	if !handler.isRendering {
 		handler.isRendering = true
-		accounts, err := walletMiddleware.AccountsOverview(walletcore.DefaultRequiredConfirmations)
+		accounts, err := wallet.AccountsOverview(walletcore.DefaultRequiredConfirmations)
 		if err != nil {
 			handler.err = err
 		} else {
@@ -61,7 +61,7 @@ func (handler *ReceiveHandler) Render(window *nucular.Window, wallet walletcore.
 				}
 
 				if handler.numAccounts == 1 && handler.generatedAddress == "" {
-					handler.generatedAddress, handler.err = walletMiddleware.ReceiveAddress(handler.accountSelector.GetSelectedAccountNumber())
+					handler.generatedAddress, handler.err = wallet.ReceiveAddress(handler.accountSelector.GetSelectedAccountNumber())
 					handler.RenderAddress(contentWindow)
 				} else {
 					handler.accountSelector.Render(contentWindow.Window)
@@ -69,7 +69,7 @@ func (handler *ReceiveHandler) Render(window *nucular.Window, wallet walletcore.
 					// draw submit button
 					if contentWindow.Button(label.T(buttonLabel), false) {
 						// get address
-						handler.generatedAddress, handler.err = walletMiddleware.ReceiveAddress(handler.accountSelector.GetSelectedAccountNumber())
+						handler.generatedAddress, handler.err = wallet.ReceiveAddress(handler.accountSelector.GetSelectedAccountNumber())
 						if handler.err != nil {
 							contentWindow.SetErrorMessage(handler.err.Error())
 						} else {
--- a/nuklear/handlers/receive.go
+++ b/nuklear/handlers/receive.go
@@ -7,6 +7,7 @@ import (
 	"github.com/aarzilli/nucular"
 	"github.com/aarzilli/nucular/label"
 	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/nuklear/handlers/widgets"
 	"github.com/raedahgroup/godcr/nuklear/helpers"
 	qrcode "github.com/skip2/go-qrcode"
 )
@@ -14,11 +15,11 @@ import (
 type ReceiveHandler struct {
 	err         error
 	isRendering bool
-	accounts    []*walletcore.Account
 
-	// form selector index
-	selectedAccountIndex  int
-	selectedAccountNumber uint32
+	numAccounts     int
+	accountSelector *widgets.AccountSelection
+
+	addressInput nucular.TextEditor
 
 	// generatedAddress
 	generatedAddress string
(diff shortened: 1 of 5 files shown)


nuklear/handlers/widgets/account-selection.go
```diff
@@ -0,0 +1,62 @@
...
+
+	return &AccountSelection{
+		accounts:             accounts,
+		comboItems:           comboItems,
+		selectedAccountIndex: 0,
+	}
+}
+
+func (a *AccountSelection) Render(window *nucular.Window) {
+	if len(a.accounts) > 1 {
```
itswisdomagain, line 29: Need the "Account" or "Accounts" label before rendering the dropdown or account name/balance widget. Probably should be done separately in interfaces requiring this widget since the label text may change. Or we pass the label text to use to this widget in the `NewAccountSelectionWidget` function and draw the label here. The second option is preferable, so we don't always write code to draw "Accounts" label on each page that needs this widget.

nuklear/handlers/widgets/account-selection.go
```diff
@@ -0,0 +1,62 @@
...
+	return nil
+}
+
+func (a *AccountSelection) GetSelectedAccountNumber() uint32 {
+	selectedAccount := a.GetSelectedAccount()
+	if selectedAccount != nil {
+		return selectedAccount.Number
+	}
+
+	return 0
```
itswisdomagain, line 55: If selected account is nil, then the account number cannot be 0. Should return -1 instead so the caller knows to notify the user that a valid selection has not been made.

nuklear/handlers/receive.go
```diff
@@ -7,37 +7,41 @@ import (
...
 
-	// form selector index
-	selectedAccountIndex  int
-	selectedAccountNumber uint32
+	numAccounts     int
+	accountSelector *widgets.AccountSelection
+
+	addressInput nucular.TextEditor
 
 	// generatedAddress
```
itswisdomagain, line 24: the comment is basically the same as the variable name, tells me it's unnecessary.

nuklear/handlers/receive.go
```diff
@@ -7,37 +7,41 @@ import (
...
 
 func (handler *ReceiveHandler) BeforeRender() {
 	handler.err = nil
-	handler.accounts = nil
 	handler.isRendering = false
 
-	// form selector index
-	handler.selectedAccountIndex = 0
-	handler.selectedAccountNumber = uint32(0)
+	handler.addressInput.Flags = nucular.EditClipboard | nucular.EditNoCursor
```
itswisdomagain, line 32: I can edit the generated address, not cool! It disregards any edit I make immediately though, causing some kind of flicker. Add a copy button instead.

itswisdomagain (changes requested):
For the receive address page, compare what's being done for the http interface in #231, pay particular attention to the review comments regarding the hint text size and copy button using a text and not an icon.

Other things to note:
- Need to have a "Generate New Address" button to ignore the current address and generate a new address using `wallet.GenerateNewAddress`. There should be no "Regenerate" button, it currently doesn't generate a new address (just returns the same address). Even if it did, the button text should be "Generate New Address", not "Regenerate Address"
- Can the "Generate" button be removed entirely, and address generation done when page is loaded and when the account selection changes? This is done in the PR mentioned above.
- Even if there are more than 1 accounts, when the page is loaded, address should be generated and displayed for the default account. So user doesn't need to click "Generate" before seeing an address to use. If the user changes the account from the dropdown, another address can then be generated
- When there's only 1 account, the account widget should look like a disabled edit text (with border and a dim background), it looks too plain currently. Compare similar widget in the PR above.

nuklear/handlers/receive.go
```diff
@@ -7,37 +7,41 @@ import (
 	"github.com/aarzilli/nucular"
 	"github.com/aarzilli/nucular/label"
 	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/nuklear/handlers/widgets"
```
itswisdomagain, line 10: Does `widgets` have to be in `handlers`?

nuklear/handlers/receive.go
```diff
@@ -7,37 +7,41 @@ import (
...
 	isRendering bool
-	accounts    []*walletcore.Account
 
-	// form selector index
-	selectedAccountIndex  int
-	selectedAccountNumber uint32
+	numAccounts     int
+	accountSelector *widgets.AccountSelection
+
+	addressInput nucular.TextEditor
```
itswisdomagain, line 22: - `generatedAddressInput` - don't leave a reader of your code wondering which address
- of course, this shouldn't be an input but I hope the point is clear, the widget should be called `generatedAddressWidget`...

nuklear/handlers/receive.go
```diff
@@ -89,14 +90,19 @@ func (handler *ReceiveHandler) Render(window *nucular.Window, wallet walletcore.
...
+	handler.addressInput.Buffer = []rune(handler.generatedAddress)
+	handler.addressInput.Edit(window.Window)
+	helpers.ResetInputStyle(window)
 
 	// generate qrcode
 	png, err := qrcode.New(handler.generatedAddress, qrcode.Medium)
 	if err != nil {
 		window.Row(300).Dynamic(1)
 		window.LabelWrap(err.Error())
+
```
itswisdomagain, line 105: This space isn't necessary

nuklear/handlers/send.go
```diff
@@ -66,7 +62,13 @@ func (handler *SendHandler) BeforeRender() {
 func (handler *SendHandler) Render(window *nucular.Window, wallet walletcore.Wallet) {
 	if !handler.isRendering {
 		handler.isRendering = true
-		handler.fetchAccounts(wallet)
+		accounts, err := wallet.AccountsOverview(walletcore.DefaultRequiredConfirmations)
```
itswisdomagain, line 65: if the `spend unconfirmed` checkbox is checked, use 0 as required confirmations to fetch balance.

nuklear/handlers/send.go
```diff
@@ -94,7 +96,7 @@ func (handler *SendHandler) Render(window *nucular.Window, wallet walletcore.Wal
 			contentWindow.Label("Source Account", "LC")
 
 			contentWindow.Row(25).Dynamic(1)
-			handler.selectedAccountIndex = contentWindow.ComboSimple(handler.accountOverviews, handler.selectedAccountIndex, 25)
+			handler.accountSelector.Render(contentWindow.Window)
 
 			contentWindow.Row(15).Dynamic(2)
 			if contentWindow.CheckboxText("Spend Unconfirmed", &handler.spendUnconfirmed) {
```
itswisdomagain, line 102: Also update the balance to use 0 as required confirmations. The spendable balance might be 0 initially, but once you say spend unconfirmed, that can rise to a higher value.

nuklear/handlers/widgets/account-selection.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type AccountSelection struct {
+	accounts   []*walletcore.Account
+	comboItems []string
+
+	selectedAccountIndex int
+}
+
+func NewAccountSelectionWidget(accounts []*walletcore.Account) *AccountSelection {
```
itswisdomagain, line 15: Can we make this smarter? Allow this widget to fetch the accounts itself? I've noticed that whoever uses this widget calls `wallet.AccountsOverview` before using this widget, that's repeated code that can be included here once and for all instead.

Going with this, this function should take `wallet, spendUnconfirmed` instead of `accounts`. Then depending on whether `spendUnconfirmed` is true or not, this function fetches the balance with the default required confirmations or 0 as required confirmations.

itswisdomagain (commented):
The staking page also needs to use this new widget.

itswisdomagain:
This is attended to in #236

PR #266, opened 2019-04-09
Receive page design alignment

ademuanthony:
This is a redesign of the' receive page' that normalizes the font and elements alignment

receive-page-amend

### DIFF
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -170,6 +170,7 @@ func (routes *Routes) receivePage(res http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	accounts = append(accounts, accounts...)
 	data := map[string]interface{}{
 		"accounts": accounts,
 	}
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -170,12 +170,11 @@ func (routes *Routes) receivePage(res http.ResponseWriter, req *http.Request) {
 		return
 	}
 
-	accounts = append(accounts, accounts...)
 	data := map[string]interface{}{
 		"accounts": accounts,
 	}
 
-	data = routes.fillAddressData(data, accounts[0].Number)
+	data = routes.fillAddressData(data, accounts[0].Number, true)
 	if _, has := data["imageData"]; has {
 		data["imageStr"] = fmt.Sprintf("<img data-target=\"receive.image\" src=\"%s\"/>", data["imageData"])
 	}
@@ -194,11 +193,19 @@ func (routes *Routes) generateReceiveAddress(res http.ResponseWriter, req *http.
 		data["message"] = err.Error()
 		return
 	}
-	data = routes.fillAddressData(data, uint32(accountNumber))
+
+	newAddress := req.URL.Query().Get("new")
+	data = routes.fillAddressData(data, uint32(accountNumber), newAddress == "1")
 }
 
-func (routes *Routes) fillAddressData(data map[string]interface{}, accountNumber uint32) map[string]interface{} {
-	address, err := routes.walletMiddleware.GenerateNewAddress(accountNumber)
+func (routes *Routes) fillAddressData(data map[string]interface{}, accountNumber uint32, newAddress bool) map[string]interface{} {
+	var address string
+	var err error
+	if newAddress {
+		address, err = routes.walletMiddleware.GenerateNewAddress(accountNumber)
+	} else {
+		address, err = routes.walletMiddleware.ReceiveAddress(accountNumber)
+	}
 	if err != nil {
 		data["success"] = false
 		data["message"] = err.Error()
--- a/web/routes/handlers.go
+++ b/web/routes/handlers.go
@@ -174,11 +174,8 @@ func (routes *Routes) receivePage(res http.ResponseWriter, req *http.Request) {
 		"accounts": accounts,
 	}
 
-	data = routes.fillAddressData(data, accounts[0].Number, true)
-	if _, has := data["imageData"]; has {
-		data["imageStr"] = fmt.Sprintf("<img data-target=\"receive.image\" src=\"%s\"/>", data["imageData"])
-	}
-
+	// don't generate new address by default, return previous unused address if it exists
+	data = routes.generateAddress(data, accounts[0].Number, false)
 	routes.renderPage("receive.html", data, res)
 }
 
@@ -190,42 +187,39 @@ func (routes *Routes) generateReceiveAddress(res http.ResponseWriter, req *http.
 	accountNumber, err := strconv.ParseUint(accountNumberStr, 10, 32)
 	if err != nil {
 		data["success"] = false
-		data["message"] = err.Error()
+		data["errorMessage"] = fmt.Sprintf("Invalid account selected: %s", accountNumberStr)
 		return
 	}
 
-	newAddress := req.URL.Query().Get("new")
-	data = routes.fillAddressData(data, uint32(accountNumber), newAddress == "1")
+	generateNewAddress := req.URL.Query().Get("new") == "yes"
+	data = routes.generateAddress(data, uint32(accountNumber), generateNewAddress)
 }
 
-func (routes *Routes) fillAddressData(data map[string]interface{}, accountNumber uint32, newAddress bool) map[string]interface{} {
+func (routes *Routes) generateAddress(data map[string]interface{}, accountNumber uint32, generateNewAddress bool) map[string]interface{} {
 	var address string
 	var err error
-	if newAddress {
+	if generateNewAddress {
 		address, err = routes.walletMiddleware.GenerateNewAddress(accountNumber)
 	} else {
 		address, err = routes.walletMiddleware.ReceiveAddress(accountNumber)
 	}
 	if err != nil {
 		data["success"] = false
-		data["message"] = err.Error()
+		data["errorMessage"] = err.Error()
 		return data
 	}
 
 	png, err := qrcode.Encode(address, qrcode.Medium, 256)
 	if err != nil {
 		data["success"] = false
-		data["message"] = err.Error()
+		data["errorMessage"] = err.Error()
 		return data
 	}
 
-	// encode to base64
-	encodedStr := base64.StdEncoding.EncodeToString(png)
-	imgStr := fmt.Sprintf("data:image/png;base64,%s", encodedStr)
-
 	data["success"] = true
-	data["address"] = address
-	data["imageData"] = imgStr
+	data["generatedAddress"] = address
+	data["qrCodeBase64Image"] = base64.StdEncoding.EncodeToString(png)
+
 	return data
 }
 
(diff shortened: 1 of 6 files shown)


ademuanthony:
The 'receive page' has been updated to align the QR code and the address horizontally and to hide the 'generate new address' button in a dropdown

receive-page-amed p2 - plain

receive-page-amed p2 -dropped

ademuanthony:
receive-page-same-font

web/views/receive.html
```diff
@@ -13,54 +13,59 @@
...
                                    {{ range $account := .accounts }}
-                                       <option value="{{ $account.Number }}">{{ accountString $account }}</option>
+                                       <option value="{{ $account.Number }}">{{ $account.Name }}</option>
                                    {{ end }}
                                    </select>
                                {{ else }}
                                {{ range $account := .accounts }}
                                        <input data-target="receive.account" value="{{ $account.Number }}" type="hidden" name="account">
-                                   <input class="form-control" readonly value="{{ accountString $account }}">
+                                   {{/*<input class="form-control singleAccountTextView" disabled value="{{ $account.Name }}">*/}}
```
itswisdomagain: This comment can go

web/views/receive.html
```diff
@@ -13,54 +13,59 @@
...
-                                <span data-target="receive.address" id="address">{{ .address }}</span>
-                                <button data-action="click->receive#copyAddressToClipboard" class="btn btn-xs btnCopy" title="Copy to clipboard"><i class="fas fa-copy"></i> (<span data-target="receive.copyButtonText" >copy</span>)</button>
-                            </div>
-                            <div id="result-img">{{ rawHtml .imageStr }}</div>
-                        </div>
                        <div class="row">
                            <div class="col-md-6">
-                               <button data-target="receive.generateNewAddressButton" data-action="click->receive#generateNewAddress"
-                                       class="btn btn-primary">Generate New Address</button>
+                               <div class="result-card">
```
itswisdomagain: Can use the utility class `text-center` instead. The font-size need not be changed really...

web/static/app/src/css/style.scss
```diff
@@ -293,3 +293,20 @@ label.list-group-item div {
...
+
+.singleAccountTextView {
+  width: 80% !important;
+  margin-left: 11px !important;
+  border: none;
+  background: inherit !important;
+  padding: 0;
+}
+
+.result-card {
```
itswisdomagain: Think we can do without this. The class name isn't very helpful either, doesn't give an immediate idea of where it could be used...

web/static/app/src/css/style.scss
```diff
@@ -258,7 +258,7 @@ label.list-group-item div {
 
 .btnCopy {
   font-size: 14px;
-  padding: 3px;
+  padding: 10px;
```
itswisdomagain: This space is meant to separate this element from the elements before and after it, should be `0 10px` for horizontal padding only.

web/static/app/src/css/style.scss
```diff
@@ -293,3 +293,20 @@ label.list-group-item div {
...
+  padding: 0;
+}
+
+.result-card {
+  font-size: 15px;
+  text-align: center;
+}
+
+.address-qr-img img {
+  margin-left: -60px;
```
itswisdomagain: - I'm guessing this is to take care of the padding embedded in the qr image, a comment here would help.
- `margin-left: -4%;` seem to produce a better result across varying screen widths, the current margin looks like this on mobile screen width:
Screenshot 2019-04-12 at 9 08 47 AM

web/views/receive.html
```diff
@@ -13,54 +13,59 @@
...
+                                   <div data-action="click->receive#copyAddressToClipboard" style="margin-top: -20px;"
+                                        class="address-qr-img">{{ rawHtml .imageStr }}</div>
+                                   <div id="result-text" style="margin-top: -25px;">
```
itswisdomagain: Is this id used anywhere? styles or js code?

web/static/app/src/css/style.scss
```diff
@@ -293,3 +293,20 @@ label.list-group-item div {
...
+.singleAccountTextView {
+  width: 80% !important;
+  margin-left: 11px !important;
+  border: none;
+  background: inherit !important;
+  padding: 0;
+}
+
+.result-card {
+  font-size: 15px;
```
itswisdomagain: I see a lot of elements are individually set to use font size as 15px, perhaps this can be done in body instead? Or some other parent element...
the account text and the lead text are other instances I've noticed

web/static/app/src/css/style.scss
```diff
@@ -293,3 +293,20 @@ label.list-group-item div {
 label.list-group-item h5 {
   display: inline !important;
 }
+
+.singleAccountTextView {
```
itswisdomagain: no longer necessary

web/views/receive.html
```diff
@@ -13,54 +13,59 @@
...
-                                <button data-action="click->receive#copyAddressToClipboard" class="btn btn-xs btnCopy" title="Copy to clipboard"><i class="fas fa-copy"></i> (<span data-target="receive.copyButtonText" >copy</span>)</button>
-                            </div>
-                            <div id="result-img">{{ rawHtml .imageStr }}</div>
-                        </div>
                        <div class="row">
                            <div class="col-md-6">
-                               <button data-target="receive.generateNewAddressButton" data-action="click->receive#generateNewAddress"
-                                       class="btn btn-primary">Generate New Address</button>
+                               <div class="result-card">
+                                   <div data-action="click->receive#copyAddressToClipboard" style="margin-top: -20px;"
```
itswisdomagain: The margin here and in the next 2 lines, are they really necessary? I took them out and the look wasn't adversely affected imo.
If they are necessary though, comments should be added before each of those lines explaining that the negative margins are applied to reduce/eliminate some extra white space around the qr code image.

itswisdomagain (changes requested):

web/views/receive.html
```diff
@@ -28,19 +28,20 @@ <h5 class="card-title">Receiving Decred</h5>
...
                                {{ end }}
                                {{ end  }}
                                </div>
                            </div>
                        </div>
                        <div class="row">
                            <div class="col-md-6">
-                               <div class="result-card">
+                               <div class="text-center">
+                                   {{/* the nagative margin this is to take care of the padding embedded in the qr image*/}}
```
itswisdomagain: html comments should be like this
`<!-- comment here -->`

itswisdomagain (changes requested):

itswisdomagain:
Just noticed a display issue when there are multiple accounts in wallet. The account drop down appears cut off. Something `z-index` can prolly fix.

Screen Shot 2019-04-13 at 9 21 31 PM

Also, @raedah once said these dropdowns are too wide and can/should be reduced to simply fit the content adequately.

itswisdomagain:
Another issue with multiple accounts, this time when you select a different account to generate new address. Ideally, as soon as a different account is selected, the currently displayed address and qr code should be cleared so a user does not mistake it for the newly generated address.

In the following screenshot, the "whew" account doesn't really exist (has a wrong account number) and there was an error trying to generate the address. The error could not be displayed.

Screen Shot 2019-04-13 at 9 24 28 PM

ademuanthony:
updated

receive-page-14-4-error-page

receive-page-14-4-overlap-fixed

web/routes/handlers.go
```diff
@@ -170,6 +170,7 @@ func (routes *Routes) receivePage(res http.ResponseWriter, req *http.Request) {
 		return
 	}
 
+	accounts = append(accounts, accounts...)
```
itswisdomagain: I believe this is test code, please delete.

web/static/app/src/controllers/receive_controller.js
```diff
@@ -1,43 +1,34 @@
 import { Controller } from 'stimulus'
 import axios from 'axios'
-import { clearMessages, copyToClipboard, setErrorMessage } from '../utils'
+import { clearMessages, copyToClipboard, hide, setErrorMessage, show, showSuccessNotification } from '../utils'
 
 export default class extends Controller {
   static get targets () {
     return [
-      'account', 'address', 'image', 'copyButtonText', 'generateNewAddressButton'
+      'errorMessage', 'successMessage',
```
itswisdomagain: Do we need `successMessage` on this page? We can show success messages using toastr, but error messages need to stay visible on the screen, so we should retain the `errorMessage` element

web/static/app/src/controllers/receive_controller.js
```diff
@@ -1,43 +1,34 @@
...
+      'account', 'addressContainer', 'address', 'image', 'copyButtonText', 'generateNewAddressButton'
```
itswisdomagain: Can split this into 3 lines to semantically differentiate the sections
- account selector widget
- generated address container with qr code, address text and copy button
- generate new address button

The functions in this file should be ordered similarly:
- `getRecieveAddress` for when the account selection changes (see related comment below)
- `copyAddressToClipboard `
- `generateNewAddress`

web/static/app/src/controllers/receive_controller.js
```diff
@@ -1,43 +1,34 @@
...
-    let _this = this
-    setTimeout(function () {
-      _this.copyButtonTextTarget.textContent = 'copy'
-    }, 1500)
+    showSuccessNotification('Copied to clipboard')
   }
 
   generateNewAddress () {
+    hide(this.addressContainerTarget)
     this.generateNewAddressButtonTarget.textContent = 'Generating...'
```
itswisdomagain: The button's text was all small letters, should probably stick to that.

web/static/app/src/controllers/receive_controller.js
```diff
@@ -46,9 +37,8 @@ export default class extends Controller {
         setErrorMessage(_this, 'Unable to generate address. Something went wrong.')
       })
       .then(function () {
-        if (onComplete) {
-          onComplete()
-        }
+        _this.generateNewAddressButtonTarget.textContent = 'Generate New Address'
```
itswisdomagain: This isn't the exact original button text, it was all small letters.

web/static/app/src/controllers/receive_controller.js
```diff
@@ -1,43 +1,34 @@
...
-      _this.generateNewAddressButtonTarget.textContent = 'Generate New Address'
-      _this.generateNewAddressButtonTarget.removeAttribute('disabled')
-    })
-  }
-
-  generate (onComplete) {
-    let _this = this
+    const _this = this
     clearMessages(this)
     axios.get('/generate-address/' + this.accountTarget.value)
```
itswisdomagain: When the account selection changes, we don't need to generate a new address, we should return previously generated address that is unused. The `wallet.ReceiveAddress` method ensures that no new address is generated unless the previously generated address has been used. That method should be used for this scenario.

When user clicks generate new address however, the `wallet.GenerateNewAddress` method should be used to ignore previously unused addresses and return a new one.

We can make this endpoint take a `new` query parameter that can be used the route handler to determine which of the above methods to use.

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #276, opened 2019-04-12
fix dcrlibwallet version

itswisdomagain:
A recent rebase on https://github.com/raedahgroup/dcrlibwallet/tree/dcrlibwallet-wip affected the commit hash used as dcrlibwallet version in godcr go.mod.

This PR fixes the resulting godcr build issue by using the latest version of dcrlibwallet and applying new API changes in dcrlibwallet.

The new changes in dcrlibwallet that godcr now benefits from include
- transaction indexing after every sync operation
- ability to send max amount to a particular tx destination address

More work will need to be done in godcr and dcrlibwallet for these changes to become very noticeable however. This PR primarily ensures that godcr points to a correct dcrlibwallet revision for successful builds.


ademuanthony (approved):
tested ok

PR #321, opened 2019-04-18
Quit app if wallet database is in use by other process

### LINKED ISSUE #282: wallet already open
"It's taking longer than expected to open your wallet. The wallet may already be opened by another app."

This condition should be caught earlier. It shouldnt require waiting. We should be able to tell that the wallet is already open without delay and exit immediately rather then causing the user to force quite. This should also happen before the 'Launching desktop app with nuklear' msg.

itswisdomagain:
Fixes #282.

Screen Shot 2019-04-18 at 12 21 28 PM


Sirmorrison (approved):
tested works fine for dcrlibwallet. good for merge.

PR #325, opened 2019-04-19
HTTP history page filter, count and account name

### LINKED ISSUE #302: http, history page and filter and count
See android app.

Add filter by transaction type. See logic android uses for which filters to display. Also notice the count of transactions being displayed.

Amount and Fee headers should be better centered over the column.

Use the font size change on the decimal places for the amounts.

Remove the # column.

Truncate the hash to 10 chars. example f387255410...

Put Amount column before the Direction column

ademuanthony:
Resolves #302 

The transaction hash on the history page is truncated and the table columns are re-arranged.
This PR also added transaction filtering, pagination info and account name to the history page.
Added coloured icons for each transaction direction as found on the Android app.
Show the number of transaction for each filter on the dropdown

image

image

image

image


Sirmorrison:
The `fee` table header can be align to the center. its too far right. Also the `amount` can be aligned a little to the left.

generally are the headers center, left or right align? hard to tell.

image

Sirmorrison:
run `go fmt ./...`

web/routes/templates.go
```diff
@@ -2,9 +2,11 @@ package routes
 
 import (
 	"fmt"
+	"github.com/raedahgroup/dcrlibwallet/txhelper"
```
Sirmorrison: re-arrange imports

Sirmorrison (changes requested):

ademuanthony:
> The `fee` table header can be align to the center. its too far right. Also the `amount` can be aligned a little to the left.
> 
> generally are the headers center, left or right align? hard to tell.
> 
> image

The headers are aligned with the content

web/routes/handlers.go
```diff
@@ -264,7 +265,20 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
...
+		txCount, txCountErr := routes.walletMiddleware.TransactionCount(txindex.Filter().WithTxTypes(txType))
+		if txCountErr != nil {
+			routes.renderError(fmt.Sprintf("Cannot load history page. "+
+				"Error getting total transaction count: %s", txCountErr.Error()), res)
+			return
+		}
+		transactionCountByType[txType] = txCount
+	}
+
+	alTxCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
```
Sirmorrison: this is supposed to be `allTxCount` if am not mistaken

web/routes/handlers.go
```diff
@@ -287,15 +301,24 @@ func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
...
-		"txs":          txns,
-		"currentPage":  int(pageToLoad),
-		"previousPage": int(pageToLoad - 1),
-		"totalPages":   int(math.Ceil(float64(txCount) / float64(txPerPage))),
+		"transactionCountByType": transactionCountByType,
+		"txs":                    txns,
+		"currentPage":            int(pageToLoad),
+		"previousPage":           int(pageToLoad - 1),
+		"totalPages":             int(math.Ceil(float64(alTxCount) / float64(txPerPage))),
+		"transactionTotalCount":  alTxCount,
```
Sirmorrison: same as above `allTxCount`

web/routes/handlers.go
```diff
@@ -287,15 +301,24 @@ func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
...
+		"accountsCount":          len(accounts),
 	}
 
 	totalTxLoaded := int(offset) + len(txns)
-	if totalTxLoaded < txCount {
+	if totalTxLoaded < alTxCount {
```
Sirmorrison: same as above

web/routes/handlers.go
```diff
@@ -306,7 +329,12 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
 
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	filter := txindex.Filter()
```
Sirmorrison, line 329: camel case `txIndex`

Sirmorrison (changes requested):

web/routes/handlers.go
ademuanthony: this is an imported package name

Sirmorrison: ok

web/views/history.html
```diff
@@ -22,28 +38,37 @@
...
+                            <td>{{ accountName $txn }}</td>
                             <td>{{ $txn.LongTime }}</td>
                             <td>{{ $txn.Type }}</td>
-                            <td>{{ $txn.Direction }}</td>
-                            <td style="text-align: right">{{ amountDcr $txn.Amount }}</td>
-                            <td style="text-align: right">{{ amountDcr $txn.Fee }}</td>
+                            <td><img style="width: 15px" src="/static/images/{{directionImage $txn}}"> {{ $txn.Direction }}</td>
+                            <td style="text-align: right">
+                                {{ $amountParts := splitAmountIntoParts $txn.Amount }}
+                                {{ index $amountParts 0 }}{{ index $amountParts 1 }}<span>{{ index $amountParts 2 }}</span>
```
Sirmorrison: im not sure why you are splitting the tx.Amount but the output im seeing from testing is same as the one you changed. If there isn't any sure reason why please revert to the `{{ amountDcr $txn.Amount }}` same goes for the fees also.

Sirmorrison (changes requested):

Sirmorrison:
image

image

i noticed your `amount` and `fees` columns outputs needs to be properly formatted for your filters.  
use the `amountDcr` function where necessary.

Sirmorrison (approved):

app/walletcore/helper.go
```diff
@@ -20,6 +21,16 @@ const (
 	ReceivingDecredHint = "Each time you request payment, a new address is generated to protect your privacy."
 )
 
+func TransactionTypes() map[wallet.TransactionType]string {
```
itswisdomagain: This should probably use `txhelper.FormatTxType` from dcrlibwallet. See https://github.com/raedahgroup/dcrlibwallet/blob/dcrlibwallet-wip/txhelper/types.go#L113

web/routes/handlers.go
```diff
@@ -264,7 +265,20 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 }
 
 func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	txTypes := walletcore.TransactionTypes()
+	transactionCountByType := make(map[string]int, len(txTypes))
+
+	for _, txType := range txTypes {
```
itswisdomagain: This should probably use `txhelper.FormatTxType` from dcrlibwallet. See https://github.com/raedahgroup/dcrlibwallet/blob/dcrlibwallet-wip/txhelper/types.go#L113

web/routes/handlers.go
```diff
@@ -287,15 +301,24 @@ func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
...
-		"currentPage":  int(pageToLoad),
-		"previousPage": int(pageToLoad - 1),
-		"totalPages":   int(math.Ceil(float64(txCount) / float64(txPerPage))),
+		"transactionCountByType": transactionCountByType,
+		"txs":                    txns,
+		"currentPage":            int(pageToLoad),
+		"previousPage":           int(pageToLoad - 1),
+		"totalPages":             int(math.Ceil(float64(allTxCount) / float64(txPerPage))),
+		"transactionTotalCount":  allTxCount,
+		"accountsCount":          len(accounts),
```
itswisdomagain: Not used, can remove.

web/routes/handlers.go
```diff
@@ -306,7 +329,12 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
 
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	filter := txindex.Filter()
+	if transactionType := req.FormValue("trans-type"); transactionType != "" {
```
itswisdomagain: `tx-type` instead of `trans-type`

web/routes/templates.go
```diff
@@ -94,5 +97,75 @@ func templateFuncMap() template.FuncMap {
 		"timestamp": func() int64 {
 			return time.Now().Unix()
 		},
+		"extractDateTime": func(timestamp int64) string {
```
itswisdomagain: Can simplify to `"extractDateTime": utils.FormatUTCTime`. See https://github.com/raedahgroup/dcrlibwallet/blob/dcrlibwallet-wip/utils/utils.go#L63.

web/routes/templates.go
```diff
@@ -94,5 +97,75 @@ func templateFuncMap() template.FuncMap {
...
+			utcTime := time.Unix(timestamp, 0).UTC()
+			return utcTime.Format("2006-01-02 15:04:05")
+		},
+		"truncate": func(input string, maxLength int) string {
```
itswisdomagain: `text string, maxNumberOfCharacters int`

web/routes/templates.go
```diff
@@ -94,5 +97,75 @@ func templateFuncMap() template.FuncMap {
...
+			if len(input) <= maxLength {
```
itswisdomagain: `if len(input) < maxLength` is fine, quicker to make sense of too.

web/routes/templates.go
```diff
@@ -94,5 +97,75 @@ func templateFuncMap() template.FuncMap {
...
+			splitBalance := strings.Split(balanceStr, ".")
+			balanceParts[0] = splitBalance[0]
+			balanceParts[1] = "." + splitBalance[1][0:2]
+
+			if len(splitBalance[1]) > 2 {
+				balanceParts[2] = splitBalance[1][2:]
+			}
+			return
+		},
+		"accountName": func(txn *walletcore.Transaction) string {
```
itswisdomagain, line 96: Let's move this into a `(tx *walletcore.Transaction) WalletAccountForTx() string` function. Of course, this would have to be in the `walletcore` package, inside the `types.go` file there, below the `Transaction` type declaration.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,35 @@ export default class extends Controller {
     }
   }
 
+  changeTransactionType () {
+    this.historyTableTarget.innerHTML = ''
+    this.transactionCountTarget.setAttribute('data-transactionCount', 0)
```
itswisdomagain: `data-tx-count`

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,35 @@ export default class extends Controller {
...
+    hide(this.transactionCountContainerTarget)
+    this.nextPage = 1
+    this.fetchMoreTxs()
+  }
+
   fetchMoreTxs () {
     show(this.loadingIndicatorTarget)
 
+    let transType = this.transactionTypeTarget.value
```
itswisdomagain: `txType`

web/static/app/src/controllers/history_controller.js
```diff
@@ -90,7 +116,8 @@ export default class extends Controller {
         } else {
           _this.setErrorMessage(result.message)
         }
-      }).catch(function () {
+      }).catch(function (e) {
+        console.log(e)
```
itswisdomagain, line 115: For debugging purposes I believe. This can be removed.

web/static/app/src/controllers/history_controller.js
```diff
@@ -110,18 +137,58 @@ export default class extends Controller {
       return `${amount / 100000000} DCR`
     }
 
+    const accountName = tx => {
+      let accountNames = []
+      if (tx.direction === 1) {
+        tx.outputs.forEach(output => {
+          if (output.account_numberr === -1 || accountNames.indexOf(output.account_name) !== -1) {
```
itswisdomagain: `account_numberr`? See https://github.com/raedahgroup/dcrlibwallet/blob/dcrlibwallet-wip/txhelper/types.go#L109.

In order to use the `tx.account_number` json key specified there, you'll need to update the dcrlibwallet mod version. Reason is, dcrlibwallet previously used a wrong json tag (`previous_account`).

- Run `go get github.com/raedahgroup/dcrlibwallet@dcrlibwallet-wip`
- Then `go mod tidy`

Please test to ensure that txs loaded this way display properly. You can temporarily reduce the number of txs loaded per page in order to trigger an async call to fetch more txs. Remember to restore the default minimum tx per page count.

web/views/history.html
```diff
@@ -7,10 +7,26 @@
         {{ template "header" .connectionInfo }}
         <div class="content">
             <div class="container">
+                <div class="row">
+                    <div class="col-md-2 col-sm-12 mb-2">
+                        <select data-target="history.transactionType" data-action="change->history#changeTransactionType" class="form-control">
+                            <option value="">All Transaction({{ .transactionTotalCount }})</option>
```
itswisdomagain: - `All Transactions`
- Also add a space before the bracket.
- This doesn't show fully on your screenshot, may need to increase the width of this element.

web/views/history.html
```diff
@@ -7,10 +7,26 @@
...
+                            {{ range $txType, $count := .transactionCountByType}}
+                            <option value="{{ $txType }}">{{ $txType }}({{ $count }})</option>
```
itswisdomagain: Space between tx type and (count)

web/views/history.html
```diff
@@ -22,28 +38,31 @@
                     </thead>
                 </table>
 
-                <table class="table">
+                <table class="table history-table" style="table-layout: fixed">
```
itswisdomagain: This style isn't necessary as the class already defines it.

web/views/history.html
```diff
@@ -22,28 +38,31 @@
...
-                            <th>Fee</th>
-                            <th>Status</th>
-                            <th>Hash</th>
-                        </tr>
+                    <tr>
+                        <th style="width: 120px;">Account</th>
+                        <th style="width: 142px;">Date (UTC)</th>
+                        <th style="width: 68px;">Type</th>
+                        <th style="width: 105px;">Direction</th>
+                        <th style="width: 115px;text-align: center;">Amount</th>
```
itswisdomagain: space between width and text-align properties.

web/views/history.html
```diff
@@ -22,28 +38,31 @@
...
+                        <th style="width: 100px;text-align: center;">Fee</th>
+                        <th style="width: 85px;">Status</th>
+                        <th>Hash</th>
+                    </tr>
                     </thead>
                     <tbody data-target="history.historyTable">
+                    {{ $accountsCount := .accountsCount}}
```
itswisdomagain: Not necessary, remove.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,35 @@ export default class extends Controller {
...
+    axios.get(`/next-history-page?page=${this.nextPage}&trans-type=${transType}`)
       .then(function (response) {
         let result = response.data
         if (result.success) {
+          if (!result.txs || result.txs.length === 0) {
+            _this.setErrorMessage('No Transaction Found')
+            return
+          }
+
+          let transCount = parseInt(_this.transactionCountTarget.getAttribute('data-transactionCount')) + result.txs.length
```
itswisdomagain: Isn't this same as counting the number of rows in the table? Generally a better idea to not store data that can be inferred elsehow.

itswisdomagain (changes requested):

web/routes/handlers.go
ademuanthony: Here we are trying to loop through all the types, but the suggested function only gives a string representation of a specified transaction type and not a list of all the types

raedah:
If 0 items are found for a filter type, the line should be hidden and not be a choice.

app/walletcore/helper.go
itswisdomagain: I believe we can do away with this function. Make it a slice instead:
```
var TransactionTypes = []wallet.TransactionType {
	wallet.TransactionTypeRegular,
	wallet.TransactionTypeCoinbase,
	wallet.TransactionTypeTicketPurchase,
	wallet.TransactionTypeVote,
	wallet.TransactionTypeRevocation,
}
```

I'll add more thoughts on the 2 sites calling this function.

app/walletcore/types.go
```diff
@@ -78,3 +80,31 @@ type Transaction struct {
...
+		for _, name := range accountNames {
+			if name == accountName {
+				return true
+			}
+		}
+		return false
+	}
+	if tx.Direction == txhelper.TransactionDirectionReceived {
+		for _, output := range tx.Outputs {
+			if output.AccountNumber == -1 || isInArray(output.AccountName) {
```
itswisdomagain: `isInArray` as a function name begs the question, which array? The function could be renamed to `addWalletAccount` and the function modified to check if the account name already exists, then appending if it doesn't.

This line and similar lines would become (also notice the comment, it's not immediately obvious what `-1` signifies):
```
if output.AccountNumber != -1 {
     // output belongs to this wallet and not an external wallet address
     addWalletAccount(output.AccountName)
}
```

The `isInArray` modified to `addWalletAccount` would look like this:
```
addWalletAccount := func(accountName string) {
	for _, name := range accountNames {
		if name == accountName {
			return
		}
	}
	accountNames = append(accountNames, accountName)
}
```

app/walletcore/types.go
```diff
@@ -78,3 +80,31 @@ type Transaction struct {
...
+func (tx *Transaction) WalletAccountForTx() string {
+	var accountNames []string
+	isInArray := func(accountName string) bool {
+		for _, name := range accountNames {
+			if name == accountName {
+				return true
+			}
+		}
+		return false
+	}
```
itswisdomagain, line 96: Please add a single line space after this function.

app/walletcore/types.go
```diff
@@ -78,3 +80,31 @@ type Transaction struct {
...
+	} else {
+		for _, input := range tx.Inputs {
+			if input.AccountNumber == -1 || isInArray(input.AccountName) {
+				continue
+			}
+			accountNames = append(accountNames, input.AccountName)
+		}
+	}
+	return strings.Join(accountNames, ", ")
+}
```
itswisdomagain: go fmt

web/routes/handlers.go
```diff
@@ -264,7 +265,23 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 }
 
 func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	txTypes := walletcore.TransactionTypes()
+	transactionCountByType := make(map[string]int, len(txTypes))
+
+	for _, txType := range txTypes {
```
itswisdomagain: I see here that what you need is the string representation of each txType, not the `wallet.TransactionType` itself. If you loop over `walletcore.TransactionTypes()`, you can get the string representation by calling `txhelper.FormatTransactionType(txType)`

web/routes/templates.go
```diff
@@ -3,9 +3,13 @@ package routes
 import (
 	"fmt"
 	"html/template"
+	"strings"
 	"time"
 
 	"github.com/decred/dcrd/dcrutil"
+	"github.com/decred/dcrwallet/wallet"
+	"github.com/raedahgroup/dcrlibwallet/txhelper"
+	libutils "github.com/raedahgroup/dcrlibwallet/utils"
```
itswisdomagain: Whenever an external package clashes with an internal package, I oft prefer to alias the internal package, prefixing with `godcr`. An example of that is `godcrApp` used in `fyne/fyne.go`. I think it's easier to make sense of. Could do same here.

web/routes/templates.go
```diff
@@ -83,5 +87,52 @@ func templateFuncMap() template.FuncMap {
...
+				for _, input := range txn.Inputs {
+					if input.AccountNumber == -1 || isInArray(input.AccountName) {
+						continue
+					}
+					accountNames = append(accountNames, input.AccountName)
+				}
+			}
+			return strings.Join(accountNames, ", ")
+		},
+		"directionImage": func(txn *walletcore.Transaction) (imageFile string) {
```
itswisdomagain: I don't see that naming the return variable is necessary.

web/routes/templates.go
```diff
@@ -83,5 +87,52 @@ func templateFuncMap() template.FuncMap {
...
 		},
+		"extractDateTime": libutils.FormatUTCTime,
+		"truncate": func(text string, maxNumberOfCharacters int) string {
+			if len(text) < maxNumberOfCharacters {
+				return text
+			}
+			return fmt.Sprintf("%s...", text[0:maxNumberOfCharacters])
+		},
+		"accountName" : func(txn *walletcore.Transaction) string {
+			var accountNames []string
```
itswisdomagain: This can `return tx. WalletAccountForTx()`

web/routes/handlers.go
```diff
@@ -326,8 +350,8 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 
 	var txPerPage int32 = walletcore.TransactionHistoryCountPerPage
 	offset := (int32(pageToLoad) - 1) * txPerPage
-	txns, err := routes.walletMiddleware.TransactionHistory(offset, txPerPage, nil)
 
+	txns, err := routes.walletMiddleware.TransactionHistory(offset, txPerPage, filter)
```
itswisdomagain, line 355: Please correct the condition of the if statements in lines 295 and 245 of this file to be `pageToLoad <= 0` instead of `pageToLoad < 0`

web/static/app/src/controllers/history_controller.js
```diff
@@ -10,7 +10,9 @@ export default class extends Controller {
       'errorMessage',
       'previousPageButton', 'nextPageButton',
       'pageReport',
-      'loadingIndicator'
+      'loadingIndicator',
+      'transactionType',
+      'transactionCountContainer', 'transactionCount', 'transactionTotalCount'
```
itswisdomagain: The targets listed here should be ordered the same way they appear on the page. I believe `loadingIndicator` is the last on the page and `transactionType` is the first.

web/static/app/src/controllers/history_controller.js
```diff
@@ -22,6 +24,9 @@ export default class extends Controller {
   alignTableHeaderWithStickyHeader () {
     this.stickyTableHeaderTarget.style.width = `${this.historyTableTarget.clientWidth}px`
 
+    if (this.historyTableTarget.innerHTML === '') {
```
itswisdomagain, line 24: I think this if statement can move to the top of this function, returning before any resizing is done.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,32 @@ export default class extends Controller {
     }
   }
 
+  changeTransactionType () {
```
itswisdomagain: This function isn't triggered **to** `changeTransactionType `, it is triggered **on** `transactionTypeFilterChanged`. Function should be renamed.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,32 @@ export default class extends Controller {
...
+    this.historyTableTarget.innerHTML = ''
+    this.transactionCountTarget.setAttribute('data-tx-count', 0)
```
itswisdomagain: I don't see this attribute used anywhere; probably don't need to set it.

web/static/app/src/controllers/history_controller.js
```diff
@@ -10,7 +10,9 @@ export default class extends Controller {
       'errorMessage',
       'previousPageButton', 'nextPageButton',
       'pageReport',
-      'loadingIndicator'
+      'loadingIndicator',
+      'transactionType',
```
itswisdomagain: This could/should be `selectedTransactionTypeFilter` or `selectedTxTypeFilter` for short.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,32 @@ export default class extends Controller {
...
 
+    const txType = this.transactionTypeTarget.value
+
     const _this = this
-    axios.get(`/next-history-page?page=${this.nextPage}`)
+    axios.get(`/next-history-page?page=${this.nextPage}&tx-type=${txType}`)
       .then(function (response) {
         let result = response.data
         if (result.success) {
+          if (!result.txs || result.txs.length === 0) {
```
itswisdomagain: `result.success` is true but there's either no txs in result or txs count is 0? This shouldn't happen. Why are we writing code for such possibility?

If `result.success` is true, then txs should be returned. Unless there's no tx for the type specified (in which case it shouldn't be on the filter) or the page used in the query does not have data to return (in which case the handler should return an error message, not `result.success = true`; the error would be that there's `no transaction for page  ${x}`

web/static/app/src/controllers/history_controller.js
```diff
@@ -110,20 +133,62 @@ export default class extends Controller {
       return `${amount / 100000000} DCR`
     }
 
+    const accountName = tx => {
```
itswisdomagain: `tx` in bracket `(tx)` might be more readable

web/static/app/src/controllers/history_controller.js
```diff
@@ -110,20 +133,62 @@ export default class extends Controller {
...
+      let accountNames = []
+      if (tx.direction === 1) {
+        tx.outputs.forEach(output => {
+          if (output.account_numberr === -1 || accountNames.indexOf(output.account_name) !== -1) {
```
itswisdomagain: - `account_number`
- You can (should) test that this `fetchMoreTxs` function works as it should by temporarily reducing the number of txs to load per page, so that this function is triggered initial load. Remember to reset the number of txs loaded per page.
- The if statement can be made 1 line shorter by using:
```
if (output.account_number !== -1 && accountNames.indexOf(output.account_name) < 0)
```
- You can also use a `Set` instead of an array so you don't need to check if the account name has been previously added, you just add to the set knowing that it will not re-add the value if it has already been added.

web/static/app/src/controllers/history_controller.js
```diff
@@ -110,20 +133,62 @@ export default class extends Controller {
...
+
+      fields[4].innerHTML = amountDcr(tx.amount)
+      fields[5].innerHTML = amountDcr(tx.fee)
+
+      fields[6].innerText = tx.status
+      fields[7].innerHTML = `<a href="/transaction-details/${tx.hash}">${truncate(tx.hash, 10)}}</a>`
 
       _this.historyTableTarget.appendChild(txRow)
+      _this.transactionCountTarget.textContent = _this.historyTableTarget.childElementCount
+
```
itswisdomagain: Can do without this space.

web/static/app/src/controllers/history_controller.js
```diff
@@ -1,6 +1,6 @@
 import { Controller } from 'stimulus'
 import axios from 'axios'
-import { hide, show } from '../utils'
+import { hide, show, splitAmountIntoParts, truncate } from '../utils'
```
itswisdomagain: `splitAmountIntoParts` is not used.

web/routes/handlers.go
```diff
@@ -306,7 +325,12 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
 
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	filter := txindex.Filter()
+	if transactionType := req.FormValue("tx-type"); transactionType != "" {
+		filter = filter.WithTxTypes(transactionType)
+	}
+
+	txCount, txCountErr := routes.walletMiddleware.TransactionCount(filter)
```
itswisdomagain: Should be `allTxCount` since it's not the count of txs being returned but the count of all txs for this filter (and if no filter is specified, then it's the count of all txs in wallet)

web/routes/handlers.go
```diff
@@ -337,6 +361,7 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 		data["currentPage"] = int(pageToLoad)
 		data["previousPage"] = int(pageToLoad - 1)
 		data["totalPages"] = int(math.Ceil(float64(txCount) / float64(txPerPage)))
+		data["txCount"] = txCount
```
itswisdomagain: Key should be `transactionTotalCount` since the value assigned is not the count of txs being returned but the total count of txs for the given filter.

web/routes/handlers.go
```diff
@@ -306,7 +325,12 @@ func (routes *Routes) getNextHistoryPage(res http.ResponseWriter, req *http.Requ
 	data := map[string]interface{}{}
 	defer renderJSON(data, res)
 
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	filter := txindex.Filter()
+	if transactionType := req.FormValue("tx-type"); transactionType != "" {
```
itswisdomagain: The complete list of filters on android history page are `All`, `Sent`, `Received`, `Yourself`, `Staking` and `Coinbase`. `Staking` is a combination of `Ticket`, `Vote` and `Revocation`.

 See https://github.com/decred/dcrandroid/blob/master/app/src/main/java/com/dcrandroid/fragments/HistoryFragment.kt#L223-L260.

web/routes/handlers.go
```diff
@@ -264,7 +265,23 @@ func (routes *Routes) getRandomChangeOutputs(res http.ResponseWriter, req *http.
 }
 
 func (routes *Routes) historyPage(res http.ResponseWriter, req *http.Request) {
-	txCount, txCountErr := routes.walletMiddleware.TransactionCount(nil)
+	txTypes := walletcore.TransactionTypes()
+	transactionCountByType := make(map[string]int, len(txTypes))
```
itswisdomagain: The complete list of filters on android history page are `All`, `Sent`, `Received`, `Yourself`, `Staking` and `Coinbase`. `Staking` is a combination of `Ticket`, `Vote` and `Revocation`.

See https://github.com/decred/dcrandroid/blob/master/app/src/main/java/com/dcrandroid/fragments/HistoryFragment.kt#L223-L260.

We'll also need to rename `Transferred` to `Yourself`. This has been done in the latest commit to `dcrlibwallet-wip`, you would need to update the mod version on godcr for this to reflect. Remember to run `go mod tidy` after updating the version with `go get github.com/raedahgroup/dcrlibwallet@dcrlibwallet-wip`.

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +78,32 @@ export default class extends Controller {
...
       .then(function (response) {
         let result = response.data
         if (result.success) {
+          if (!result.txs || result.txs.length === 0) {
+            _this.setErrorMessage('No Transaction Found')
+            return
+          }
+
+          _this.transactionTotalCountTarget.textContent = result.txCount
+          show(_this.transactionCountContainerTarget)
```
itswisdomagain, line 103: `transactionCountTarget` should be updated as well

web/routes/handlers.go
itswisdomagain: I'm concerned about setting the length of this `transactionCountByType` map to `len(txTypes)` as not all tx types may be accounted for. Best to set the capacity instead: `make(map[string]int, 0, len(txTypes))`; or just `make(map[string]int, 0)`

web/views/history.html
```diff
@@ -22,28 +39,30 @@
...
                         {{ range $index, $txn := .txs }}
                         <tr>
+                            <td>{{ $txn.WalletAccountForTx }}</td>
                             <td>{{ $txn.LongTime }}</td>
                             <td>{{ $txn.Type }}</td>
-                            <td>{{ $txn.Direction }}</td>
-                            <td style="text-align: right">{{ amountDcr $txn.Amount }}</td>
-                            <td style="text-align: right">{{ amountDcr $txn.Fee }}</td>
+                            <td><img style="width: 15px" src="/static/images/{{directionImage $txn}}"> {{ $txn.Direction }}</td>
+                            <td style="text-align: right">{{ amountDcr $txn.Amount}}</td>
```
itswisdomagain: - `{{ amountDcr $txn.Amount }}`
- return the trailing space at the next line too

itswisdomagain (changes requested):

itswisdomagain:
**These anomalies are observed on page load:**
Screenshot 2019-05-03 at 4 00 03 AM
- Direction icons are same
- Date, amount and fee fields getting truncated (why are we using fixed column widths again?)

**When you change the tx filter, these additional anomalies are observed:**
Screenshot 2019-05-03 at 4 00 19 AM
- tx count report at the top right
- External account listed
- Hash getting `}` at the end instead of `...`

web/static/app/src/controllers/history_controller.js
ademuanthony: `transactionCountTarget ` is updated in `displayTxs`

app/walletcore/helper.go
```diff
@@ -18,8 +20,50 @@ const (
...
+var TransactionFilters = []string {
+	TransactionFilterSent,
+	TransactionFilterReceived,
+	TransactionFilterYourself,
+	TransactionFilterStaking,
+	TransactionFilterCoinbase,
+}
+
+func BuildTransactionFilter(filters ...string) *txindex.ReadFilter {
+	var (
```
itswisdomagain, line 45: This can be inlined

web/static/app/src/controllers/history_controller.js
```diff
@@ -73,14 +77,31 @@ export default class extends Controller {
...
     show(this.loadingIndicatorTarget)
 
+    const filter = this.selectedFilterTarget.value
+
     const _this = this
-    axios.get(`/next-history-page?page=${this.nextPage}`)
+    axios.get(`/next-history-page?page=${this.nextPage}&filter=${filter}`)
       .then(function (response) {
+        // since results are appended to the table, discard this response
+        // if the user have changed the filter before the result is gottend
```
itswisdomagain: `if the user **has** changed the filter before the result is **gotten**`

itswisdomagain (changes requested):
I'm unable to build, getting the following error:
Screenshot 2019-05-06 at 8 11 57 PM

If possible, make the following adjustments:

app/walletcore/helper.go
ademuanthony: The variable needs to be initialized immediately

Sirmorrison:
> 
> 
> If 0 items are found for a filter type, the line should be hidden and not be a choice.

image

Im still getting the filters with 0 items please check

also rebase your PR

app/walletcore/helper.go
```diff
@@ -23,8 +25,53 @@ const (
 	LegacyTestnetHDPath = "m / 44' / 11' / "
 	MainnetHDPath       = "m / 44' / 42' / "
 	LegacyMainnetHDPath = "m / 44' / 20' / "
+	TransactionFilterSent     = "Sent"
```
itswisdomagain, line 29: Add an empty line space before this, to show separation of concerns.

itswisdomagain (approved):

Sirmorrison (changes requested):
image

* first i noticed that on smaller screens the hash column on the history page is hidden and the table is not scroll-able. what this implies is that a user will not be to view the details of the transaction.

* Also on full screen i noticed the hashes are truncated thereby leaving the too much space on the history table.
we could either reduce the width of the hash column on the history table, or display the full length of the hashes and detect screen width which will truncate the hash on smaller screens rather than hide it. 
image

ademuanthony:
image

Sirmorrison:
image

i think this error is messing with the output.

image

also the issue with the hash column
image

web/routes/templates.go
```diff
@@ -83,5 +86,28 @@ func templateFuncMap() template.FuncMap {
...
+		"truncate": func(text string, maxNumberOfCharacters int) string {
+			if len(text) < maxNumberOfCharacters {
+				return text
+			}
+			return fmt.Sprintf("%s...", text[:maxNumberOfCharacters])
+		},
+		"accountName": func(txn *walletcore.Transaction) string {
+			return txn.WalletAccountForTx()
+		},
+		"directionImage": func(txn *walletcore.Transaction) string {
```
Sirmorrison: i think this will make more sense as `txDirectionImage`

Sirmorrison (changes requested):

Sirmorrison (approved):

Issue #311, opened 2019-04-18
terminal, history view bug


raedah:
Is not showing a full screen worth of transactions on first view. Once scrolled down to the next page, then the screen properly fills with transactions.

Columns are changing width each page. This is not ideal. Fixed size is preferred.

Spacing between columns should be increased. Someone said the toolkit does not support this. If so, we should file an upstream issue.

Show the total count of transactions at the top of the page.

Sirmorrison:
>Show the total count of transactions at the top of the page.

this task will be fixed with #280 reason is @itswisdomagain is working on tx indexing which will make tx counting possible.

raedah:
closed by https://github.com/raedahgroup/godcr/pull/327

PR #327, opened 2019-04-20
Terminal navigation, history table column padding and fixed width fix.

### LINKED ISSUE #310: terminal, navigations
First thing I notice when I load the terminal view is I can not use the arrow keys to navigate up and down the menu. This should be fixed. Cause seems to be that overview page is showing, but its not obvious because the highlighting makes it look like the control is still on the nav bar. The nav bar items should have a different type of highlight when actually selected as in use, versus when simply being selected by the cursor. Left and Right arrow keys should be able to open and close menus. Up and down arrows should work within opened menus rather then having to use the tab keys.

Esc key works but there is a large delay in its responsiveness. This should be fixed.

nav bar is showing keys but the keys only work when the nav bar is in focus. I suggest hiding the keys when they are not usable.

Border is shown around the nav bar which implies it is in focus. Set the border to be around whatever part of the screen is the currently focused pane.

ESC from the nav bar should prompt the user if they want to quit the app.

Accounts, Security, Settings pages say nothing. Should say coming soon at least.

Sirmorrison:
fixes #310 and #334 

new nav option now have `coming soon` message
image

exit by pressing esc from navigation menu or exit button
image

shortcuts hide when a nav option has focus
image

*when focus is returned to the nav menu shortcuts are displayed
image

transaction history table column padding and fixed width
image

*when a menu item is selected, the menu option text-colours are changed from white to gray and the shourtcuts keys are hidden also the border on the nav menu changes for 2lines border to a single line border.

image

*When focus is shifted to the menu option, the border is move to the right and when returned to the the nav manu, the border is move to the right.
image

image

### DIFF
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -23,7 +23,7 @@ const (
 	DefaultRequiredConfirmations = 2
 
 	// minimum number of transactions to return per call to Wallet.TransactionHistory()
-	TransactionHistoryCountPerPage = 20
+	TransactionHistoryCountPerPage = 25
 
 	ConfirmedStatus = "Confirmed"
 
--- a/terminal/pages/accounts.go
+++ b/terminal/pages/accounts.go
@@ -1,0 +1,28 @@
+package pages
+
+import (
+	"github.com/gdamore/tcell"
+	"github.com/raedahgroup/godcr/terminal/primitives"
+	"github.com/rivo/tview"
+		"github.com/raedahgroup/godcr/terminal/helpers"
+)
+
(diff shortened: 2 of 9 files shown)


raedah:
design ack

ademuanthony (changes requested):
Update the description to show that this PR is a continuation of #334 which resolves #311.
#311 also requires that you should `Show the total count of transactions at the top of the page`. This will require that #330 by @itswisdomagain is completed. We should either add this to the description and create a new issue for it or mark this PR as WIP

itswisdomagain:
> #311 also requires that you should `Show the total count of transactions at the top of the page`.

This requirement is implemented in #330.

ademuanthony (approved):

PR #328, opened 2019-04-20
Display transaction history in pages on nuklear

### LINKED ISSUE #314: nuklear, crash on large history list
nuklear history view ends up going into high cpu use due to it trying to load a very large history. App has to be force quit. There needs to be some type of safety limiting mechanism here. The entire history should not be loaded. We likely only want to load items as the user scrolls if nuklear can do this.

yaowork:
This fetches and displays transaction history in pages on the history page of the nuklear interface. In navigating using "previous" and "next" buttons, this PR makes sure that already fetched transaction pages are just re-rendered instead of  fetched again from the walletrpc server. 

It should be noted that the nucular/nuklear project does not provide us with a scroll listener, so this PR makes use of "next" and "previous" buttons to navigate through pages instead of the infinite scrolling suggested by #314   

Fixes #314 

Screenshot from 2019-04-22 18-01-23
Screenshot from 2019-04-22 18-01-44

### DIFF
--- a/nuklear/pagehandlers/sync.go
+++ b/nuklear/pagehandlers/sync.go
@@ -91,6 +91,9 @@ func (s *SyncHandler) Render(window *nucular.Window, changePage func(*nucular.Wi
 		return
 	}
 
+	changePage(window, "overview")
+	return
+
 	widgets.NoScrollGroupWindow("sync-page", window, func(pageWindow *widgets.Window) {
 		pageWindow.Master().Style().GroupWindow.Padding = image.Point{10, 10}
 		pageWindow.AddHorizontalSpace(20)
--- a/nuklear/pagehandlers/sync.go
+++ b/nuklear/pagehandlers/sync.go
@@ -91,9 +91,6 @@ func (s *SyncHandler) Render(window *nucular.Window, changePage func(*nucular.Wi
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


raedah:
Currently a random number of items are being returned on each page. This should be fixed so the number of items returned are consistent. This is also an obvious behavior that should have been noted in the PR description when it was submitted.

ademuanthony (changes requested):
Is infinite scrolling not possible as requested in the #314 being resolved? Please update the description to reflect this.
Also https://github.com/raedahgroup/godcr/pull/328#issuecomment-485585342

yaowork:
The PR has been updated to display a consistent number of transactions per page.

ademuanthony (approved):

nuklear/pagehandlers/sync.go
```diff
@@ -91,6 +91,9 @@ func (s *SyncHandler) Render(window *nucular.Window, changePage func(*nucular.Wi
 		return
 	}
 
+	changePage(window, "overview")
+	return
```
itswisdomagain: Forgot to remove this.

itswisdomagain (changes requested):

itswisdomagain (approved):

Issue #303, opened 2019-04-18
http, transaction details layout


raedah:
Change the layout of the data so that it is a single pane view rather then double pane view. Model this after how android transaction details are displayed. 

Datetime should be showing the full format. 2019-04-18 05:25:53 (UTC)

http view header is failing to show peer and block status currently and needs to be fixed

Issue #280, opened 2019-04-12
cli, history


raedah:
show total transaction count

Showing transactions 1-32 of 1493,

be able to view (p)revious page instead of just (m)ore. Should probably be (n)ext.
(f)irst and (l)ast are good too.

Do not exit after showing transaction details. Stay at the history menu.

Issue #274, opened 2019-04-10
cli: improvements to history and show transaction commands


itswisdomagain:
### history command
Running `godcr history` produces a bit of an eye sore on terminal/command prompt windows with small width:
![Screen Shot 2019-04-10 at 8 33 54 PM](https://user-images.githubusercontent.com/18400051/55907997-2b57a480-5bd0-11e9-8339-2b9e00e1b2f9.png)

It looks better when the width is wider:
![Screen Shot 2019-04-10 at 8 34 11 PM](https://user-images.githubusercontent.com/18400051/55908014-33174900-5bd0-11e9-8a97-62b4ce59a229.png)

In line with #256, the following adjustments are necessary:
- remove fee and hash from the displayed table
- show only time if less than 24 hours ago and only date if greater than 24 hours ago

### show transaction command
Running `godcr showtransaction txid` does not currently show the inputs and outputs, unless `-d` is added to the command:
![Screen Shot 2019-04-10 at 10 15 58 PM](https://user-images.githubusercontent.com/18400051/55914117-3dd8da80-5bde-11e9-8334-3ae12ff72200.png)

- retire/remove the `-d` flag for the `showtransaction` command
- always show complete tx details (inputs and outputs included) when `godcr showtransaction txid` is run
- format the inputs/outputs **amounts** the same way it is done on terminal, see below:
**cli ❌**
![Screen Shot 2019-04-10 at 10 15 58 PM](https://user-images.githubusercontent.com/18400051/55914223-82647600-5bde-11e9-8b32-1d775ccb3e5f.png)
**terminal 👍**
![Screen Shot 2019-04-10 at 10 19 07 PM](https://user-images.githubusercontent.com/18400051/55914302-b3dd4180-5bde-11e9-8e50-25d92665e2d1.png)

PR #330, opened 2019-04-21
save transactions to, fetch transactions from indexed db

### LINKED ISSUE #268: tx details page may be displaying incomplete information
An analysis of the tx details page in dcrandroid shows that the data used to display inputs/outputs on the app comes from 2 dcrlibwallet functions: `libwallet.GetTransaction` and `libwallet.DecodeTransaction`. See how both functions are used on dcrandroid [here](https://github.com/decred/dcrandroid/blob/9efb3e5a48752f2a749716a7d7410bd81ed111e2/app/src/main/java/com/dcrandroid/activities/TransactionDetailsActivity.java#L207), [here](https://github.com/decred/dcrandroid/blob/9efb3e5a48752f2a749716a7d7410bd81ed111e2/app/src/main/java/com/dcrandroid/activities/TransactionDetailsActivity.java#L257) and [here](https://github.com/decred/dcrandroid/blob/9efb3e5a48752f2a749716a7d7410bd81ed111e2/app/src/main/java/com/dcrandroid/activities/TransactionDetailsActivity.java#L266).

Godcr code currently uses only `libwallet.DecodeTransaction` to fetch the inputs/outputs data to be displayed when showing tx details.

Perhaps both dcrlibwallet functions can be merged to ensure consistent display on the godcr and dcrandroid apps.

### LINKED ISSUE #295: reduce wallet history load time
The `GetTransactions` function used by `dcrwalletrpc` and `dcrlibwallet` fetches transactions one block at a time. The `GetTransactions` function take starting and ending block heights/hashes and returns all transactions in the provided block range.

Prior to #244, godcr does not specify a block range and thus all transactions in all blocks (including unmined transactions, "block height -1") are queried and all transactions in the wallet returned.

#244 was an effort to reduce wallet history load time by fetching transactions in batches instead of all at once. Beginning with that PR, calls to the `GetTransactions` function used by `dcrwalletrpc` and `dcrlibwallet` are accompanied with a block range. The first batch start at block height -1 (unmined transactions) to the best block height (latest block). Subsequent batches continue from the previous block height fetched to the next block height in descending order. This is done continuously until a specified minimum number of transactions are retrieved at which point, further batches of transactions are not requested.

For a wallet without any transaction, however, all block ranges will have to be processed progressively before the history function returns an empty list. This results in longer load times for wallets with no transactions.

This process can be improved by fetching transactions for all blocks after every sync operation and saving transactions found to an indexed database. This is already done for dcrlibwallet but needs some modification to work reliably with godcr.

The tx index database is saved together with the wallet database in the same directory e.g `~/.godcr/testnet3`. Following modifications are needed:

- Support for transaction indexing when connecting with dcrwallet over rpc. This is especially tough since it is difficult (currently impossible?) to accurately tell the location of the wallet database file that the active dcrwallet process is using.

- A user could manually delete a `wallet.db` file leaving behind the accompanying `tx.db` (tx index database) file. If a user creates a new wallet or copies over a different `wallet.db` file to the directory already having a `tx.db` file, the wallet's history that will be displayed by godcr will include transactions not belonging to the wallet. It's important to find a way to uniquely tie wallet databases to tx index databases.

### LINKED ISSUE #312: terminal, tx detail block number
Currently block number is always showing 0 which is not accurate.

itswisdomagain:
Fixes #295.

In order to speed up history load time and efficiently count total number of transactions in wallet, this PR applies the dcrlibwallet tx indexing feature to godcr, regardless of whether wallet connection is established via dcrlibwallet or dcrwalletrpc. 

In addition to improving history load times due to indexing, the following improvements come with this PR:

### Transaction Details
- The information processed/indexed and returned per transaction is updated in this PR. Tx details now contain a complete list of inputs and outputs (previously only inputs/outputs relating to wallet were processed). This resolves #268.
- Tx details now also return the account number and account name for each input in a transaction. Returning account name for inputs and address for outputs is necessary in resolving
  - #302 (wrt displaying account column in http history page) and
  - #250 (wrt showing account names for inputs in tx details page).
- Currently, account names are not returned for outputs but a look at the tx details page implementation (in terminal for example) exposes how to get the account name for each output using the address _(should probably fix this, have the core code return the account name rather than leaving interfaces to convert address to account name)_.

### Transaction Count and Filtering
- Necessary in resolving #302 (wrt displaying count of all transactions, getting count of transactions by type and/or direction; and filtering by transaction type and direction).
- Part of #311 (regarding showing total tx count at the top of the terminal history page) is fixed in this PR.
Screenshot 2019-04-23 at 11 19 51 AM

- This PR also resolves #312 relating to terminal tx detail page not showing block height. 
Screenshot 2019-04-23 at 11 11 36 AM

- Part of #280 (wrt showing total transaction count in cli history) is also fixed in this PR.
Screenshot 2019-04-23 at 11 32 06 AM

- Part of #274 and #303 (re showing the full format for datetime in tx detail page in cli and http history).
Screenshot 2019-04-23 at 11 31 51 AM
Screenshot 2019-04-23 at 11 33 01 AM

### Using the Filter
See the updated comments on `wallet.TransactionCount` and `wallet.TransactionHistory` for guide in
- how to get tx count for each filter
- knowing what filters to display (where tx count for that filter is > 0)
- retrieving transactions for a particular filter
- combining filters

_Need to test properly to confirm that this PR does not cause breaking changes in history and tx detail pages across all interfaces when connecting via dcrlibwallet and dcrwalletrpc._


raedah:
Overall looks good. Tx details views need to show timezone on the date time. Resolve conflicts.

ademuanthony (approved):
okay to merge

PR #333, opened 2019-04-21
HTTP send page improvement

### LINKED ISSUE #308: http, send page
Address input field is not wide enough to show the full address. 
Input should be verified as a valid address and have an error shown if it is not.
Amount input should not allow input over the valid amount size of this many digits 12345678.12345678 . Refer to the android app for rules on valid amount inputs.
The X to get ride of a row should not have a box around it or it looks too similar to the input boxes.  "Amount must be a non-zero positive number" shows for the amount, but it is located under the address box. Error should be under the proper input box. Also see proper error text from android. 'Add another address' should not be in a box that looks similar to the input fields. Reset button should instead be named Clear Fields. Rename Next button to Send. Send button should be grey until all needed fields are properly filled out. Send Max checkbox is failing, not sure why. The user should be able to collapse the list of custom inputs after they have selected the inputs they want to use.

Display additional fields that android shows,  Fee, Estimated Size, Balance After, Exchange Rate.

Support the Send to Account option as does android.

ademuanthony:
Resolves #308 
The following changes where made
- The address field was increased
- Instant address validation was implemented
- Errors in destination fields are shown immediately after the affected fields
image
- Added validation to the amount field to stop the user from specifying more than his balance for sending
image
- The send button will stay disabled until the send page is valid
image
image
- Implemented a function for getting the exchange rate from bitrex and show the rate to the user on the confirmation modal.
- Added two new functions to the `utils` package for estimating transaction fee and size and show this information on the transaction confirmation modal.
image
image
- Added 'send to account' feature with send max, add account, remove account and all the validations in 'send to address'
image
image
image
- Added field for entering dollar value as destination amount. The DCR field is automatically updated as the USD field is being changed and vice versa. Send max feature is available on the USD field.
image
- Input validation is available on the USD field
image
- Show USD value on the summary modal
image
-Modelled the buttons after the android app
image

image


nuklear/pagehandlers/transaction_details.go
```diff
@@ -3,11 +3,11 @@ package pagehandlers
 import (
 	"strconv"
 
+	"fmt"
```
Sirmorrison: re-arrange import

web/routes/handlers.go
```diff
@@ -11,7 +11,10 @@ import (
 	"github.com/raedahgroup/dcrlibwallet"
 	"github.com/raedahgroup/dcrlibwallet/txhelper"
 	"github.com/raedahgroup/godcr/app/config"
+	"github.com/raedahgroup/godcr/app/conversion/bitrex"
+	"github.com/raedahgroup/godcr/app/utils"
 	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/web/weblog"
 	"github.com/skip2/go-qrcode"
 	"math"
```
Sirmorrison: re-arrange import

Sirmorrison (changes requested):

Sirmorrison (approved):

app/conversion/bitrex/bitrex.go
```diff
@@ -0,0 +1,89 @@
...
+		return 0, err
+	}
+	return summary.Rate * btcAmount, nil
+}
+
+func fetchMarketSummary(marketName string) (*marketSummary, error) {
+	url := fmt.Sprintf("https://api.bittrex.com/api/v1.1/public/getmarketsummary?market=%s", marketName)
+
+	httpClient := http.Client{
+		Timeout: time.Second * 2,
```
itswisdomagain: 2 seconds timeout seems small, no?

app/utils/utils.go
```diff
@@ -32,3 +37,64 @@ func SplitAmountIntoParts(amount float64) []string {
...
+
+func EstimateSerializeSize(numberOfInputs int, destinations []txhelper.TransactionDestination) (int, error) {
+	outputs, err := makeTxOutputs(destinations)
+	if err != nil {
+		return 0, err
+	}
+
+	var changeAddresses []string
+	for _, destination := range destinations {
+		changeAddresses = append(changeAddresses, destination.Address)
```
itswisdomagain, line 121: `changeAddresses` and `outputs` are actually not the same.

I see that the new functions required are `EstimateFee` and `EstimateSerializeSize`.

These operations are already quite well handled in `dcrlibwallet`, so what I did was to export a public function that can be used from the `txhelper` package (`txhelper.EstimateTxFeeAndSize`). You'll need to update the dcrlibwallet version in go.mod and run `go mod tidy` to use these new functions.

Usage should be something like `txhelper.EstimateTxFeeAndSize(len(payload.utxos), payload.sendDestinations, payload.changeDestinations)`

web/routes/handlers.go
```diff
@@ -113,11 +125,69 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request)
...
+func (routes *Routes) getFeeAndSize(res http.ResponseWriter, req *http.Request) {
+	data := map[string]interface{}{}
+	defer renderJSON(data, res)
+
+	payload, err := retrieveSendPagePayload(req, routes.walletMiddleware.GenerateNewAddress)
+	if err != nil {
+		data["error"] = fmt.Sprintf("Cannot get summary: %s", err.Error())
+		return
+	}
+	// If no input is selected, use all inputs in the account to determine
```
itswisdomagain: This comment needs to be updated. The reason we need all inputs is to properly determine the number of inputs to take into account when estimating fee and tx serialize size.

web/routes/handlers.go
```diff
@@ -113,11 +125,69 @@ func (routes *Routes) maxSendAmount(res http.ResponseWriter, req *http.Request)
...
+
+func (routes *Routes) getFeeAndSize(res http.ResponseWriter, req *http.Request) {
+	data := map[string]interface{}{}
+	defer renderJSON(data, res)
+
+	payload, err := retrieveSendPagePayload(req, routes.walletMiddleware.GenerateNewAddress)
+	if err != nil {
+		data["error"] = fmt.Sprintf("Cannot get summary: %s", err.Error())
+		return
+	}
```
itswisdomagain, line 164: Adding an empty line after each if block in this function would be helpful.

app/walletcore/helper.go
```diff
@@ -81,12 +81,34 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
...
+func BuildTxDestinations(destinationAddresses, destinationAccounts,
+	destinationAmounts, sendMaxAmountValues []string, addressFunc func(accountNumber uint32) (string, error)) (
 	destinations []txhelper.TransactionDestination, totalAmount dcrutil.Amount, err error) {
 
-	for i := range destinationAddresses {
+	destinationAccountAddresses := make([]string, len(destinationAccounts))
+	for i, accountStr := range destinationAccounts {
+		account, err := strconv.ParseInt(accountStr, 10, 32)
+		if err != nil {
+			return nil, 0, err
```
itswisdomagain: `fmt.Errorf("Invalid account number")`

app/walletcore/helper.go
```diff
@@ -81,12 +81,34 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
...
 }
 
-func BuildTxDestinations(destinationAddresses, destinationAmounts, sendMaxAmountValues []string) (
+func BuildTxDestinations(destinationAddresses, destinationAccounts,
+	destinationAmounts, sendMaxAmountValues []string, addressFunc func(accountNumber uint32) (string, error)) (
 	destinations []txhelper.TransactionDestination, totalAmount dcrutil.Amount, err error) {
 
-	for i := range destinationAddresses {
+	destinationAccountAddresses := make([]string, len(destinationAccounts))
+	for i, accountStr := range destinationAccounts {
```
itswisdomagain: `accountNumber` or `accountNumberStr`

app/walletcore/helper.go
```diff
@@ -81,12 +81,34 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
 	return
 }
 
-func BuildTxDestinations(destinationAddresses, destinationAmounts, sendMaxAmountValues []string) (
+func BuildTxDestinations(destinationAddresses, destinationAccounts,
+	destinationAmounts, sendMaxAmountValues []string, addressFunc func(accountNumber uint32) (string, error)) (
```
itswisdomagain: `addressFunc` can be `generateAddress` or `generateAddressFn`

app/walletcore/helper.go
```diff
@@ -81,12 +81,34 @@ func GetChangeDestinationsWithRandomAmounts(wallet Wallet, nChangeOutputs int, a
...
 	destinations []txhelper.TransactionDestination, totalAmount dcrutil.Amount, err error) {
 
-	for i := range destinationAddresses {
+	destinationAccountAddresses := make([]string, len(destinationAccounts))
```
itswisdomagain, line 141: Append to `destinationAddresses` instead of creating a new `destinationAccountAddresses` slice.

web/routes/helpers.go
```diff
@@ -66,6 +66,7 @@ func retrieveSendPagePayload(req *http.Request) (payload *sendPagePayload, err e
 
 	// parse send destinations form data
 	destinationAddresses := req.Form["destination-address"]
+	destinationAccounts := req.Form["destination-account"]
```
itswisdomagain: `destinationAccountNumbers`, `destination-account-number`

itswisdomagain (changes requested):
Let's work on these changes first, will test and review more afterwards.

Sirmorrison (changes requested):
image

image

*on the send page? `send to account` column doesn't show the input fields and when you return to `send to address` the input fields are also hidden until the page is refresh.

web/views/send.html
```diff
@@ -82,50 +104,141 @@ <h5 class="card-title">Sending Decred</h5>
...
-                                                    class="btn btn-outline-secondary removeDestinationBtn"
-                                                    style="border: 1px solid #ced4da !important;" type="button">X</button>
                                         </div>
+                                    </template>
 
-                                        <div class="col-sm-12 invalid-feedback d-none"></div>
+                                    <div class="form-group mb-0">
+                                        <button type="button" data-action="click->send#newDestinationAccount"
+                                                class="btn btn-outline-primary shadow-sm">
+                                            Add another output
```
Sirmorrison: ` Add another output` should be ` Add another account`

Sirmorrison (changes requested):

web/views/send.html
ademuanthony: it was changed to output so that the user will not confuse it with creating a new account

app/utils/utils.go
Sirmorrison: > 
> 
> `changeAddresses` and `outputs` are actually not the same.
> 
> I see that the new functions required are `EstimateFee` and `EstimateSerializeSize`.
> 
> These operations are already quite well handled in `dcrlibwallet`, so what I did was to export a public function that can be used from the `txhelper` package (`txhelper.EstimateTxFeeAndSize`). You'll need to update the dcrlibwallet version in go.mod and run `go mod tidy` to use these new functions.
> 
> Usage should be something like `txhelper.EstimateTxFeeAndSize(len(payload.utxos), payload.sendDestinations, payload.changeDestinations)`

have this request been implemented?

web/routes/handlers.go
Sirmorrison: > 
> 
> This comment needs to be updated. The reason we need all inputs is to properly determine the number of inputs to take into account when estimating fee and tx serialize size.

this also?

Sirmorrison:
image
These issues should be looked at.

image
I think `add another output` should be `add another account`

Sirmorrison:
image
I noticed that when i select `custom inputs` it brings out this `change output` at the bottom of the screen.
Now the issue is.. when i select an output say `1` then change my mind not to use `custom input` but `automatic input` instead, it clears everything but i still get this error at the bottom saying `Total change amount percentage must be equal to 100. Current total is 0` which prevents me from sending the fund.

Sirmorrison:
I also noticed i can send funds to the address of same account... is that a normal something or an issue?

ademuanthony:
> I also noticed i can send funds to the address of same account... is that a normal something or an issue?

yes

Sirmorrison (approved):

app/utils/utils.go
ademuanthony: `changeAddresses` and `output` are used differently here though

app/rebase.txt
```diff
@@ -0,0 +1 @@
+delete me again
```
itswisdomagain: Please delete

app/utils/utils.go
```diff
@@ -6,10 +6,12 @@ import (
 	"strconv"
 	"strings"
 
-	"github.com/decred/dcrd/dcrutil"
-
 	"github.com/raedahgroup/godcr/app/config"
 	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/decred/dcrd/dcrutil"
```
itswisdomagain, line 9: Organize your imports alphabetically.

itswisdomagain (changes requested):

web/routes/handlers.go
```diff
@@ -81,14 +84,24 @@ func (routes *Routes) sendPage(res http.ResponseWriter, req *http.Request) {
...
 		"spendUnconfirmedFunds": routes.settings.SpendUnconfirmed,
 	}
+
+	if routes.settings.CurrencyConverter == "bitrex" {
+		exchangeRate, err := bitrex.DcrToUsd(1)
+		if err != nil {
+			weblog.LogError(fmt.Errorf("error fetching exchange rate: %s", err.Error()))
+		}
+
+		data["exchangeRate"] = fmt.Sprintf("%.8f", exchangeRate)
```
itswisdomagain: This should be in an else block.

itswisdomagain (changes requested):

itswisdomagain (approved):

PR #334, opened 2019-04-22
Terminal history bug fix

### LINKED ISSUE #311: terminal, history view bug

Sirmorrison:
fixes #311 

overview recent activity table
image

history transaction table
image

### DIFF
--- a/app/walletcore/helper.go
+++ b/app/walletcore/helper.go
@@ -14,7 +14,7 @@ const (
 	DefaultRequiredConfirmations = 2
 
 	// minimum number of transactions to return per call to Wallet.TransactionHistory()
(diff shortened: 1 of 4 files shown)


Sirmorrison:
>Show the total count of transactions at the top of the page.

the tx count would be done in a later PR after #280 is merged.

PR #349, opened 2019-04-25
display nuklear transaction details on one column

### LINKED ISSUE #335: nuklear, tx details view single column
Currently the tx details page on the nuklear interface is using multiple columns in the layout to display the information and it requires more page width then the other views. The data should be layed out simply in a single column view. Most importantly is not the design, but just displaying the data properly and easily in a simple way. If a model is needed to follow, use the way that android displays tx details.

yaowork:
This changes the layout of the nuklear transaction details page to display transaction details using a single column instead of two columns. 

This fixes #335 

Screenshot from 2019-04-25 15-34-42

### DIFF
--- a/nuklear/pagehandlers/transaction_details.go
+++ b/nuklear/pagehandlers/transaction_details.go
@@ -35,12 +35,6 @@ func (handler *HistoryHandler) renderTransactionDetailsPage(window *nucular.Wind
 	}
 
 	widgets.PageContentWindowDefaultPadding("Transaction Details", window, func(contentWindow *widgets.Window) {
-		contentWindow.AddButton("Back", func() {
-			// clear tx details data so that history page is re-displayed
-			handler.clearTxDetails()
-			contentWindow.Master().Changed()
-		})
-
 		if handler.fetchTxDetailsError != nil {
 			contentWindow.DisplayErrorMessage("Error fetching transaction details", handler.fetchTxDetailsError)
 		} else if handler.selectedTxDetails != nil {
@@ -52,12 +46,29 @@ func (handler *HistoryHandler) renderTransactionDetailsPage(window *nucular.Wind
 }
 
 func (handler *HistoryHandler) displayTransactionDetails(contentWindow *widgets.Window) {
-
 	// Create row to hold tx details in 2 columns
 	// Each column will display data about the tx in a group window.
 	// Row height is calculated based on the max group items total height
 	contentWindow.Window.Row(handler.calculateTxDetailsPageHeight()).Static(670)
 	widgets.NoScrollGroupWindow("tx-details-group-1", contentWindow.Window, func(window *widgets.Window) {
+		breadcrumb := []*widgets.Breadcrumb{
+			{
+				Text: "History",
+				Action: func(text string, window *widgets.Window) {
+					handler.clearTxDetails()
+					window.Master().Changed()
+				},
+			},
+			{
+				Text:   "Transaction Details",
+				Action: nil,
+			},
+		}
+		window.AddBreadcrumb(breadcrumb)
+		if handler.selectedTxHash == "" {
+			return
+		}
+
 		txDetailsTable1 := widgets.NewTable()
 		txDetailsTable1.AddRow(
 			widgets.NewLabelTableCell("Confirmations", "LC"),
--- a/nuklear/pagehandlers/transaction_details.go
+++ b/nuklear/pagehandlers/transaction_details.go
@@ -10,6 +10,10 @@ import (
 	"github.com/raedahgroup/godcr/nuklear/widgets"
 )
 
+const (
+	dividerHeight = 10
+)
+
 func (handler *HistoryHandler) clearTxDetails() {
 	handler.selectedTxHash = ""
 	handler.selectedTxDetails = nil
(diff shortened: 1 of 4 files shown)


raedah:
Details look okay. Placement of the Back button is no good. Can probably just use the same 'bread crumb' style that http uses.

yaowork:
> Details look okay. Placement of the Back button is no good. Can probably just use the same 'bread crumb' style that http uses.

Use of breadcrumb as been implemented

raedah:
tested locally, worked fine.

nuklear/pagehandlers/transaction_details.go
```diff
@@ -3,11 +3,12 @@ package pagehandlers
 import (
 	"strconv"
```
Sirmorrison: this space is not needed

Sirmorrison (changes requested):

Sirmorrison:
image

the space between the `type` and the `amount` is too much (blue line on screen shot).
and would look nicer if the table items at the bottom half of the tx's aligned with those on the top half (red lines on screenshot).

yaowork:
updated screenshot

Screenshot from 2019-05-01 23-42-21

Sirmorrison (approved):

itswisdomagain (changes requested):
Inputs need an account name column as well, compare the screenshot from the mobile app:



Other things to note:
- **Date** not Time
- Show **Status**

yaowork:
> Inputs need an account name column as well, compare the screenshot from the mobile app:
> 
> 
> 
> Other things to note:
> 
> * **Date** not Time
> * Show **Status**

done

nuklear/pagehandlers/transaction_details.go
```diff
@@ -53,76 +47,100 @@ func (handler *HistoryHandler) displayTransactionDetails(contentWindow *widgets.
...
+					window.Master().Changed()
+				},
+			},
+			{
+				Text:   "Transaction Details",
+				Action: nil,
+			},
+		}
+		window.AddBreadcrumb(breadcrumb)
+		if handler.selectedTxHash == "" {
```
itswisdomagain: `if handler.selectedTxHash == ""`, why are we adding a breadcrumb to the page in the previous line? Don't think this group window should even be created at all if that's the case.

nuklear/pagehandlers/transaction_details.go
```diff
@@ -53,76 +47,100 @@ func (handler *HistoryHandler) displayTransactionDetails(contentWindow *widgets.
 	// Create row to hold tx details in 2 columns
 	// Each column will display data about the tx in a group window.
 	// Row height is calculated based on the max group items total height
-	contentWindow.Window.Row(handler.calculateTxDetailsPageHeight()).Static(670, 700)
-	widgets.NoScrollGroupWindow("tx-details-col-1", contentWindow.Window, func(window *widgets.Window) {
-		txDetailsTable1 := widgets.NewTable()
-		txDetailsTable1.AddRow(
+	contentWindow.Window.Row(handler.calculateTxDetailsPageHeight()).Static(730)
```
itswisdomagain: Why is a row as high as the tx details page created before the breadcrumb? Should it be after the breadcrumb? Or is the breadcrumb placed inside this pretty tall row?

nuklear/pagehandlers/transaction_details.go
```diff
@@ -146,16 +164,10 @@ func (handler *HistoryHandler) displayTransactionDetails(contentWindow *widgets.
...
-	inputsLen := len(handler.selectedTxDetails.Inputs)
-
-	var secondSectionLines int
-	if outputsLen > inputsLen {
-		secondSectionLines = outputsLen
-	} else {
-		secondSectionLines = inputsLen
-	}
+	firstSectionHeight := 460
+	lineHeight := 27
```
itswisdomagain: - What is `firstSection` whose height is 460?
- How do we know the height of the first section is 460? Should probably be declared as a constant with a comment explaining how the value was determined. I'd prefer if the height is calculated though, after creating the table, before rendering the table on a window, by calculating `len(table.Rows) * table.RowHeight`. Might need to declare a `table.RowHeight` constant too.

- Going with the previous point, we might no longer need `lineHeight`

nuklear/pagehandlers/transaction_details.go
```diff
@@ -146,16 +164,10 @@ func (handler *HistoryHandler) displayTransactionDetails(contentWindow *widgets.
...
+	outputsHeight := len(handler.selectedTxDetails.Outputs) * lineHeight
```
itswisdomagain: I believe there's a row for the "Inputs" and "Outputs" headers; and some spacing between tables.

nuklear/widgets/breadcrumb.go
```diff
@@ -0,0 +1,28 @@
...
+	Action func(string, *Window)
+}
+
+func (window *Window) AddBreadcrumb(breadcrumb []*Breadcrumb) {
+	tableCells := []TableCell{}
+
+	for index := range breadcrumb {
+		var cell TableCell
+		isLastItem := index == len(breadcrumb)-1
+		if isLastItem {
```
itswisdomagain, line 14: - `breadcrumb []*Breadcrumb` is a slice, should be `breadcrumbs`
- Instead of checking if `isLastItem`, should probably check if `breadcrumb.action == nil`.

itswisdomagain (dismissed):

nuklear/pagehandlers/transaction_details.go
yaowork: This has been fixed

Sirmorrison:
image


the issue with the tx alignment doesnt seem resolved from what im see here please take a look and ensure u pushed ur modified codes.

Sirmorrison (changes requested):
please take a look at the screenshot and notify when done.

Also add the `status` field to the transaction details page is status `confirmed` or `unconfirmed`. Status is part of the tx returned just call it.

image

yaowork:
> please take a look at the screenshot and notify when done.
> 
> Also add the `status` field to the transaction details page is status `confirmed` or `unconfirmed`. Status is part of the tx returned just call it.
> 
> image

Done

yaowork:
Updated screenshot 
Screenshot from 2019-05-09 11-34-08

nuklear/pagehandlers/transaction_details.go
```diff
@@ -50,112 +49,138 @@ func (handler *HistoryHandler) renderTransactionDetailsPage(window *nucular.Wind
...
+		widgets.NewLabelTableCell("Fee", "LC"),
+		widgets.NewLabelTableCell(dcrutil.Amount(handler.selectedTxDetails.Fee).String(), "LC"),
+	)
+	txDetailsTable.AddRow(
+		widgets.NewLabelTableCell("Fee Rate", "LC"),
+		widgets.NewLabelTableCell(dcrutil.Amount(handler.selectedTxDetails.FeeRate).String(), "LC"),
+	)
+	txDetailsTable.AddRow(
+		widgets.NewLabelTableCell("Status", "LC"),
+		widgets.NewColoredLabelTableCell(status, "LC", statusColor),
```
Sirmorrison: we have a `handler.selectedTxDetails.Status` in the return values from the call to `renderTransactionDetailsPage()` please make use of it.

also run `go fmt ./...`

Sirmorrison (changes requested):

Sirmorrison (approved):

Issue #324, opened 2019-04-18
exit button on fyne causes crash


itswisdomagain:
Clicking the Exit button on the fyne interface produces the following crash log:
```
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x7e0 pc=0x7fff9065ed89]

runtime stack:
runtime.throw(0x4d6c976, 0x2a)
        /usr/local/go/src/runtime/panic.go:617 +0x72
runtime.sigpanic()
        /usr/local/go/src/runtime/signal_unix.go:374 +0x4a9

goroutine 288 [syscall]:
runtime.cgocall(0x4a36130, 0xc0008c8b68, 0xc0016720fc)
        /usr/local/go/src/runtime/cgocall.go:128 +0x5b fp=0xc0008c8b38 sp=0xc0008c8b00 pc=0x4005f3b
github.com/go-gl/gl/v3.2-core/gl._Cfunc_glowScissor(0x7fff9065ed73, 0xde, 0x23b00000244)
        _cgo_gotypes.go:18129 +0x45 fp=0xc0008c8b68 sp=0xc0008c8b38 pc=0x4822ae5
github.com/go-gl/gl/v3.2-core/gl.Scissor(...)
        /Users/itswisdomagain/go/pkg/mod/github.com/go-gl/gl@v0.0.0-20190320180904-bf2b1f2f34d7/v3.2-core/gl/package.go:11924
fyne.io/fyne/driver/gl.(*glCanvas).walkObjects(0xc0000f4280, 0x5000380, 0xc0001670e0, 0xde, 0x25, 0x4eacf70)
        /Users/itswisdomagain/go/pkg/mod/fyne.io/fyne@v0.0.0-20190411071008-b3687258b083/driver/gl/draw.go:31 +0x382 fp=0xc0008c8c20 sp=0xc0008c8b68 pc=0x48cc6d2
fyne.io/fyne/driver/gl.(*glCanvas).walkObjects(0xc0000f4280, 0x4fffd20, 0xc000167130, 0xde, 0x4, 0x4eacf70)
        /Users/itswisdomagain/go/pkg/mod/fyne.io/fyne@v0.0.0-20190411071008-b3687258b083/driver/gl/draw.go:21 +0x5ab fp=0xc0008c8cd8 sp=0xc0008c8c20 pc=0x48cc8fb
[...]

Issue #358, opened 2019-04-30
Potential data race issue in fyne mode


ademuanthony:
The app panic with the following log while running in fyne mode
![image](https://user-images.githubusercontent.com/11990092/56962709-20e55680-6b4f-11e9-8e86-876bfaf13e0f.png)
![image](https://user-images.githubusercontent.com/11990092/56962725-280c6480-6b4f-11e9-922d-3d0d0259fef8.png)
[Error log](https://github.com/raedahgroup/godcr/files/3121273/fyne-26-4-data-race-err.txt)

metaclips:
https://github.com/raedahgroup/godcr/pull/391 fixed this issue.

PR #360, opened 2019-05-02
Refactor fyne

yaowork:
This refactors code on the fyne interface. It improves performance, and simplifies code for adding other pages to the interface. 

It adds the following pages to the main UI. 

- Sync Page 
- Transaction Details Page

It also fixes the crash issue on fyne stated in #324. 

It should be noted that, the fyne library does not currently support scrolling.

Screenshot from 2019-05-02 17-10-18
Screenshot from 2019-05-02 17-10-32
Screenshot from 2019-05-02 17-11-12


fyne/pages/transaction_details.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/decred/dcrd/dcrutil"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/raedahgroup/godcr/fyne/widgets"
+)
+
+type TransactionDetails struct {
+	err    error
+	ctx    context.Context
+	wallet walletcore.Wallet
+
```
ademuanthony: this empty line can be deleted

ademuanthony (changes requested):
The overview page is randomly rendered automatically from the other pages. Since this PR is refactoring the interface, I think this issue should also be addressed here.

yaowork:
> The overview page is randomly rendered automatically from the other pages. Since this PR is refactoring the interface, I think this issue should also be addressed here.

Other than rendering the Overview page from the sync page (which is intended), i don't experience this

itswisdomagain:
#### Window content does not auto-resize to fill available window space when window is resized.
Look closely enough, you'll see the initial window size having a slightly different background.
Screenshot 2019-05-17 at 8 11 23 AM

#### Re @ademuanthony's comment:
> The overview page is randomly rendered automatically from the other pages. Since this PR is refactoring the interface, I think this issue should also be addressed here.

You can reproduce the issue by starting the app, waiting for sync to complete and the overview page to be loaded, then switch to any other page, turn off your internet for a while (will cause the wallet to lose connection to peers), turn your internet back on (will re-connect to peers, which will re-trigger the sync callbacks and cause the syncCompleted method to be re-called). It's when this method is recalled, that the window "resets" to the overview page. I believe this to also be the cause of #358.

#### Overview page -> Recent activity table -> Hash not properly aligned vertically.
Screenshot 2019-05-17 at 8 18 42 AM

Issue #277, opened 2019-04-12
fyne gui layout improvements


itswisdomagain:
Currently, the fyne gui does not naturally auto-rearrange views when the window is resized. A go routine was added to the godcr fyne gui code to watch for changes to window size and trigger the re-arrangement of views.

This https://github.com/raedahgroup/godcr/pull/243#issuecomment-480627282 suggests that the current approach is faulty and unnecessary. There is need to research and identify the appropriate way to add views to a fyne window such that they are automatically re arranged when the window is resized.

PR #365, opened 2019-05-09
Fyne refactored code base

metaclips:
image

 added theme preference in settings

image

added overview page

image

added history page with pagination support

fixed all issues below
#358 , #324 #277


app/walletcore/types.go
```diff
@@ -78,3 +80,31 @@ type Transaction struct {
...
+		for _, name := range accountNames {
+			if name == accountName {
+				return true
+			}
+		}
+		return false
+	}
+	if txn.Direction == txhelper.TransactionDirectionReceived {
+		for _, output := range txn.Outputs {
+			if output.AccountNumber == -1 || isInArray(output.AccountName) {
```
ademuanthony: please add a comment for the `-1`

app/walletcore/types.go
```diff
@@ -78,3 +80,31 @@ type Transaction struct {
...
+				continue
+			}
+			accountNames = append(accountNames, output.AccountName)
+		}
+	} else {
+		for _, input := range txn.Inputs {
+			if input.AccountNumber == -1 || isInArray(input.AccountName) {
```
ademuanthony: please add a comment for the `-1`

fyne/pages/history.go
```diff
@@ -0,0 +1,57 @@
+package pages
+
+import (
+	"fyne.io/fyne"
+	"fyne.io/fyne/widget"
+	"github.com/raedahgroup/godcr/fyne/widgets"
```
ademuanthony, line 8: An empty line above this line will be fine

itswisdomagain:
#### Initial windows size is quite small
Also, the nav menu black text should be white font color in this theme.
Screenshot 2019-05-17 at 8 48 54 AM

#### Peer count on sync window and nav menu do not update simultaneously.
Screenshot 2019-05-17 at 8 50 14 AM

#### Peer count on nav menu is temporarily invisible(?) when you switch to a page
Screenshot 2019-05-17 at 8 56 26 AM

fyne/fyne.go
```diff
@@ -3,138 +3,15 @@ package fyne
...
-		}
-	}
-
-	// refresh menu section so the changes made in this function reflects
-	app.mainWindow.Canvas().Refresh(app.menuSectionOnLeft)
-}
+func LaunchFyne(ctx context.Context, walletMiddleware godcrApp.WalletMiddleware) {
+	fyneApp := app.New()
+	window := fyneApp.NewWindow(godcrApp.DisplayName)
+	window.SetContent(pages.Menu(pages.ShowSyncWindow(walletMiddleware, window, fyneApp), window, fyneApp))
```
itswisdomagain, line 14: Can we
- create the nav menu views once (in this `LaunchFyne` function)
- set the window content once (in this `LaunchFyne` function)

Changing between pages should not need to re-create the nav menu or change the entire window content; just the relevant area of the window (the section for displaying page content) should be updated.

itswisdomagain: Also, this line 14 is too congested, could have been simplified thus:
```
syncWindow := pages.ShowSyncWindow(walletMiddleware, window, fyneApp)
pageContent := pages.Menu(syncWindow, window, fyneApp)
window. SetContent(pageContent)
```

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
 	godcrApp "github.com/raedahgroup/godcr/app"
 	"github.com/raedahgroup/godcr/fyne/widgets"
 )
 
-const (
-	defaultWindowWidth  = 800
-	defaultWindowHeight = 600
-)
+var SyncDone bool
+var Wallet godcrApp.WalletMiddleware
```
itswisdomagain, line 19: This feels a lot like a bad idea. Should not make the wallet publicly accessible from anywhere. Instead, the wallet should be passed to whatever function requires it to maintain the principle of least responsibility. Compare how this is done in other interfaces.

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
 
-func (app *fyneApp) showSyncWindow() {
+func ShowSyncWindow(wallet godcrApp.WalletMiddleware, window fyne.Window, App fyne.App) fyne.CanvasObject {
+	Wallet = wallet
+	go func() {
+		//wait for sync to be done before calling overview
```
itswisdomagain, line 24: Why are we using a goroutine and a sleeping forloop to determine when sync is done and when to load overview page? This is a potentially long running background operation that **needlessly** consumes extra computing power.

There is a sync progress callback that tells us when sync is done, why can't we load overview then and there, rather than here.

To be specific, you make `SyncDone = true` in line 78. Why not display the overview page there as well, thereby eliminating the need for this separate goroutine that does nothing else other than display overview page when `SyncDone == true`.

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
+		for SyncDone == false {
+			sec, _ := time.ParseDuration("1s")
+			time.Sleep(sec)
+		}
+		window.SetContent(Menu(OverviewPage(window, App), window, App))
+	}()
+	go func() {
+		for{
```
itswisdomagain, line 32: This is another terrible, terrible idea. An **infinite forloop** in a separate goroutine to update the number of peers and best block height? This is a long running background operation that **needlessly** consumes extra computing power.

Similar work is done for the http interface (websocket is used to send the updates to the browser). You'd notice there that updates are only sent to the UI when peer count changes and when there's a new block.

The function that sends updates via websocket is here: https://github.com/raedahgroup/godcr/blob/master/web/routes/websocket.go#L63. Do a thorough analysis of when and where that function is called, you'll find that is isn't called randomly but only when there's actually new information.

For example, that function is called here: https://github.com/raedahgroup/godcr/blob/master/web/routes/walletloader.go#L75, when there is update from the sync operation. It's worth noting that this sync operation update is received when there is a change to the number of connected peers.

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
+			info, _ := wallet.WalletConnectionInfo()
+			if info.PeersConnected <= 1 {
+				PeerConn.Text = strconv.Itoa(int(info.PeersConnected)) + " Peer Connected"
```
itswisdomagain, line 35: I see 2 issues with this statement:
- `PeerConn` is a property of the nav menu but being accessed from a `sync.go` file. Would be ideal to group `PeerConn`, `BlkHeight` and other related items into a `NavMenu` struct.
- `PeerConn` is obviously a package-wide variable. Can we reduce the scope? We certainly don't want any and every file in this package to be able to access this variable otherwise someone could add a code tomorrow in some other file that sets `PeerConn == nil` and we might not easily be able to detect that this is a wrong step. It'd be easier to detect foul usage if `PeerConn` was placed in a `NavMenu` struct, so we'd have `navMenu.PeerConn = nil`. `PeerConn == nil` may not raise attention but `navMenu.PeerConn = nil` likely will - we'd wonder why someone is modifying the nav menu from a different file. However, we should not even make it possible for anyone to update the nav menu from anywhere. Limit the scope, only provide necessary access to functions that need it. See https://en.wikipedia.org/wiki/Principle_of_least_privilege

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
-	syncWindowContent := widget.NewVBox(
-		widgets.NewVSpacer(10),
-		widget.NewLabelWithStyle("Synchronizing", fyne.TextAlignCenter, fyne.TextStyle{Italic: true}),
-		progressBar,
-		reportLabel,
-		fyne.NewContainerWithLayout(layout.NewFixedGridLayout(showDetailsButton.MinSize()), showDetailsButton),
-		errorLabel,
-	)
+	reportLabel := widget.NewLabel("")
+	widget.Refresh(reportLabel)
```
itswisdomagain, line 51: From `fyne.io/fyne/widget`:
> // Refresh instructs the containing canvas to refresh the specified widget.

At this point, `reportLabel` is just freshly created and not added to any container yet, why are we refreshing it?

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
+	reportLabel.Hide()
+	reportLabel.Alignment = fyne.TextAlignCenter
+	var infoButton *widget.Button
 
-	app.mainWindow.SetTitle(godcrApp.DisplayName)
-	app.resizeAndCenterMainWindow(syncWindowContent)
-	app.mainWindow.Show()
+	infoButton = widget.NewButton("Tap to view informations", func() {
```
itswisdomagain, line 56: `Tap to view information` not `Tap to view informations`

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
+		if infoButton.Text == "Tap to view informations" {
+
+			widget.Refresh(reportLabel)
+			reportLabel.Show()
+			infoButton.SetText("Tap to hide informations")
```
itswisdomagain, line 61: This does not follow the pattern in the rest of godcr or in the decred mobile apps. Please re-check.

fyne/pages/sync.go
```diff
@@ -1,59 +1,81 @@
...
+
+		} else {
 
-	app.walletMiddleware.SyncBlockChain(false, func(report *defaultsynclistener.ProgressReport) {
+			widget.Refresh(reportLabel)
+			reportLabel.Hide()
```
itswisdomagain, line 66: Shouldn't you hide before refreshing? Make me wonder if the call to `widget.Refresh(reportLabel)` is necessary, we can probably do without it.

fyne/pages/sync.go
```diff
@@ -64,9 +86,8 @@ func (app *fyneApp) showSyncWindow() {
 			stringReport.WriteString(fmt.Sprintf("%d%% completed, %s remaining.\n", progressReport.TotalSyncProgress, progressReport.TotalTimeRemaining))
 		}
 
-		if !showDetails {
-			reportLabel.SetText(strings.TrimSpace(stringReport.String()))
-		}
+		reportLabel.SetText(strings.TrimSpace(stringReport.String()))
```
itswisdomagain, line 89: Setting text in this label always even when it is hidden is not a memory-efficient move. We're talking bytes, yes, but the writing memory-efficient code should be a practice.

fyne/pages/sync.go
```diff
@@ -92,28 +112,23 @@ func (app *fyneApp) showSyncWindow() {
 		}
 
 		// show peer count last
-		netType := app.walletMiddleware.NetType()
+		netType := wallet.NetType()
```
itswisdomagain, line 115: There's a difference between `wallet` and `walletMiddleware`. If the type of the variable is `app.WalletMiddleware`, variable name should be `walletMiddleware`. Just semantics, but it helps code readers make accurate inference just by looking at the variable name.

fyne/pages/sync.go
```diff
@@ -92,28 +112,23 @@ func (app *fyneApp) showSyncWindow() {
...
-	// create a fixedgrid wrapper around window content so that window.CenterOnScreen will work
-	windowSize := fyne.NewSize(defaultWindowWidth, defaultWindowHeight)
-	app.mainWindow.SetContent(fyne.NewContainerWithLayout(layout.NewFixedGridLayout(windowSize), windowContent))
-	app.mainWindow.CenterOnScreen()
+	return widget.NewVBox(
+		widgets.NewVSpacer(10),
+		widget.NewLabelWithStyle("Synchronizing....", fyne.TextAlignLeading, fyne.TextStyle{Italic: true, Bold: true}),
+		widgets.NewVSpacer(10),
+		progressBar,
+		fyne.NewContainerWithLayout(layout.NewFixedGridLayout(infoButton.MinSize()), infoButton),
```
itswisdomagain, line 132: What happens if you use `infoButton` directly instead of wrapping in a `NewFixedGridLayout`?

fyne/widgets/tables.go
```diff
@@ -20,14 +20,39 @@ func NewTable() *Table {
 	return &Table{}
```
itswisdomagain, line 20: Rename this file back to `table.go`. Objects should be named in singular terms, not plural. `Class person` not `Class persons`. And of course, if the class name is `person`, it should go into a `person.go` file, not `persons.go`.

fyne/widgets/tables.go
```diff
@@ -20,14 +20,39 @@ func NewTable() *Table {
...
+	}
+
+	return widget.NewLabelWithStyle(textCell.text, textCell.alignment, textCell.style)
+}
+
 func (table *Table) AddRow(rowObjects ...TableCell) {
 	table.Rows = append(table.Rows, &TableRow{rowObjects})
 }
 
-func (table *Table) AddRowSimple(texts ...string) {
```
itswisdomagain, line 27: We should keep this method, if there's ever a need to add row items that are plain strings, no formatting.

go.mod
```diff
@@ -1,15 +1,18 @@
 module github.com/raedahgroup/godcr
 
 require (
-	fyne.io/fyne v0.0.0-20190411071008-b3687258b083
+	fyne.io/fyne v1.0.1
```
itswisdomagain, line 4: Need to run `go mod tidy` to remove previous version from `go.sum`

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+	BlkHeight *canvas.Text
+)
+
+func PageNotImplemented(win fyne.Window, App fyne.App) fyne.CanvasObject {
+	label := widget.NewLabelWithStyle("This page has not been implemented yet", fyne.TextAlignLeading, fyne.TextStyle{Italic: true})
+	return Menu(label, win, App)
+}
+
+//Menu outputs CanvasObjects "widgets" in vertical formats
+func Menu(object fyne.CanvasObject, win fyne.Window, App fyne.App) fyne.CanvasObject {
```
itswisdomagain, line 24: - This function name can be improved, it does not ONLY create the menu, it creates the entire app window: the menu and the main content.
- The `object` should be properly renamed to better describe what is to be passed in.
- `win` should be `window`, `App` should be `app` or `fyneApp`.

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"fyne.io/fyne"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/widget"
+	"github.com/raedahgroup/godcr/fyne/widgets"
+)
+
+var (
+	PeerConn  *canvas.Text
```
itswisdomagain, line 14: - Are these needed outside this package? If not, should be `peerConn` and `blkHeight`
- These variable names can be improved, seeing `PeerConn` somewhere (without seeing the type) could lead someone to think that the variables stores peer connection, `BlkHeight` sounds like a number, representing the height of a particular block. `numberOfConnectedPeersLabel` and `bestBlockHeightLabel` are more descriptive.

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+	BlkHeight *canvas.Text
+)
+
+func PageNotImplemented(win fyne.Window, App fyne.App) fyne.CanvasObject {
```
itswisdomagain, line 18: Why is `PageNotImplemented` in `menu.go`?

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+	label := widget.NewLabelWithStyle("This page has not been implemented yet", fyne.TextAlignLeading, fyne.TextStyle{Italic: true})
+	return Menu(label, win, App)
+}
+
+//Menu outputs CanvasObjects "widgets" in vertical formats
```
itswisdomagain, line 23: - There should be a single space after the second slash, before the start of a comment proper: `// Menu outputs...`
- If a comment starts with capital letter, it should end with punctuation.

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+	return fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, box, nil), box, objContainer)
+}
+
+func Button(App fyne.App, window fyne.Window) fyne.CanvasObject {
+	return widget.NewGroup("Menu",
+
+		widget.NewButton("Overview", func() {
+			if SyncDone == true {
+				window.SetContent(Menu(widget.NewLabelWithStyle("fetching data...", fyne.TextAlignTrailing, fyne.TextStyle{Italic: true, Bold: true}), window, App))
+				window.SetContent(Menu(OverviewPage(window, App), window, App))
```
itswisdomagain, line 49: Why calling `window.SetContent` twice in quick succession with different contents? Does the first content ever display for longer than a split second?

fyne/pages/menu.go
```diff
@@ -0,0 +1,102 @@
...
+		widget.NewButton("Settings", func() {
+			if SyncDone == true {
+				window.SetContent(SettingsPage(window, App))
+			}
+		}),
+
+		widget.NewButton("Exit", func() {
+			window.Close()
+			App.Quit()
+			//return
```
itswisdomagain, line 99: Remove unwanted code.

fyne/pages/overview.go
```diff
@@ -1,114 +1,94 @@
...
+func OverviewPage(windows fyne.Window, App fyne.App) fyne.CanvasObject {
+	label := widget.NewLabelWithStyle("Overview", fyne.TextAlignLeading, fyne.TextStyle{Italic: true, Bold: true})
+	balanceLabel := widget.NewLabel("Current Total Balance")
+	activityLabel := widget.NewLabel("Recent Activity")
+
+	balance := widget.NewLabel(FetchBalance())
+	balanceLabel.TextStyle = fyne.TextStyle{Bold: true}
+	activityLabel.TextStyle = fyne.TextStyle{Bold: true}
+
+	noOfTxns, _ := Wallet.TransactionCount(nil)
```
itswisdomagain, line 21: No need to count tx in wallet BEFORE fetching the top 5 txs if there are txs to fetch. Proceed to fetch the top 5, if you do so and the number of txs returned from the fetch is 0, then you know there's no tx in wallet.

itswisdomagain (changes requested):
These are required changes for now.

Also, 
- Window should have a set minimum size, the initial window size on page load is too small.
- Windows can no longer be resized beyond a the minimum size required to display a page's content. This means that the minimum window size for the history page for example is quite big. User should be able to resize even further.

PR #375, opened 2019-05-19
Spend unconfirmed balance

ReevesAk:


ademuanthony (changes requested):
Please provide a description for the changes you made in this PR and also add a screenshot if applicable.

The `handler.spendUnconfirmed` which is set to `false` in `send.go` line 58 needs to use the value of `config.Settings.SpendUnconfirmed`

PR #376, opened 2019-05-27
Accountname for transaction inputs

ReevesAk:
The tx details page (and godcr showtransaction command for cli) show account name for each tx output. 

The fix is to show account name for each tx input.


ReevesAk:
Account name display for each input transactions as it is with output transaction.

itswisdomagain:
No relevant (any) change on this PR.

PR #379, opened 2019-06-05
Ensure wallet type

ReevesAk:
The condition set to ascertain if the folder name(wallet directory) starts with any of the supported nettypes(mainnet, etc) did not include the condition and expected behavior if the nettype found in the directory folder was the same as the folder name.

The pull request was to fix that; the condition and expected behavior.

### DIFF
--- a/cli/walletloader/createwallet.go
+++ b/cli/walletloader/createwallet.go
@@ -8,7 +8,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/decred/dcrd/dcrutil"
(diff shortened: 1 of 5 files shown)


Sirmorrison:
please add some description to this PR.

ReevesAk:
The condition set to ascertain if the folder name(wallet directory) starts with any of the supported nettypes(mainnet, etc) did not include the condition and expected behavior if the  nettype found in the directory folder was the same as the folder name. 

The pull request was to fix include that; condition and expected behavior

itswisdomagain:
Not the solution to the issue referenced.

PR #383, opened 2019-06-18
Terminal ui transaction filter 

### LINKED ISSUE #378: history page transaction filter
Filter transaction history for terminal, nuklear and fyne ui's same as the android app and (or) http.

Sirmorrison:
Fixes #378 
Added transaction filter to terminal history page.

image

image

image

### DIFF
--- a/cli/commands/history.go
+++ b/cli/commands/history.go
@@ -71,7 +71,7 @@ func (history HistoryCommand) Run(ctx context.Context, wallet walletcore.Wallet)
 			pageTxRows[i] = append(pageTxRows[i], formatFee(tx.Fee))
 			pageTxRows[i] = append(pageTxRows[i], tx.Type)
 		}
-		fmt.Println(len(history.displayedTxHashes))
+
 		previousPageTxCount = len(transactions)
 		termio.PrintTabularResult(termio.StdoutWriter, columns, pageTxRows)
 
--- a/cli/commands/history.go
+++ b/cli/commands/history.go
@@ -71,7 +71,7 @@ func (history HistoryCommand) Run(ctx context.Context, wallet walletcore.Wallet)
 			pageTxRows[i] = append(pageTxRows[i], formatFee(tx.Fee))
 			pageTxRows[i] = append(pageTxRows[i], tx.Type)
 		}
-
+		fmt.Println(len(history.displayedTxHashes))
 		previousPageTxCount = len(transactions)
 		termio.PrintTabularResult(termio.StdoutWriter, columns, pageTxRows)
 
--- a/terminal/pages/history.go
+++ b/terminal/pages/history.go
@@ -22,7 +22,7 @@ var totalTxCount int
 var selectedFilter, lastSelectedFilter string
 type messageKind uint32
 const (
-	ErrorMessage messageKind = iota
+	MessageKindError messageKind = iota
 	InfoMessage messageKind = iota
 )
 
@@ -46,7 +46,7 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
 		tviewApp.QueueUpdateDraw(func() {
 			historyPage.RemoveItem(messageTextView)
 			if message != "" {
-				if kind == ErrorMessage {
+				if kind == MessageKindError {
 					messageTextView.SetTextColor(helpers.DecredOrangeColor)
 				} else {
 					messageTextView.SetTextColor(tcell.ColorWhite)
@@ -62,34 +62,36 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
 	titleTextView := primitives.NewLeftAlignedTextView("History")
 	historyPage.AddItem(titleTextView, 2, 0, false)
 
-	// tx filter
-	var transactionCountByFilter []string
+	// get total tx count
+	txCount, txCountErr := wallet.TransactionCount(nil)
+	if txCountErr != nil {
+		displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), MessageKindError)
+		tviewApp.SetFocus(historyPage)
+		return historyPage
+	}
+	totalTxCount = txCount
 
-	// var txCount int
-	// var txCountErr error
+	// tx filter
+	var transactionFilters []string
+	var selectedTxFilter *txindex.ReadFilter
 	for _, filter := range walletcore.TransactionFilters {
 		if filter == "All" {
-			txCount, txCountErr := wallet.TransactionCount(nil)
-			if txCountErr != nil {
-				displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), ErrorMessage)
-				tviewApp.SetFocus(historyPage)
-				return historyPage
-			}
-
-			totalTxCount = txCount
+			selectedTxFilter = nil
+		} else {
+			selectedTxFilter = walletcore.BuildTransactionFilter(filter)
 		}
 
-		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		filteredTxCount, txCountErr := wallet.TransactionCount(selectedTxFilter)
 		if txCountErr != nil {
-			displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), ErrorMessage)
+			displayMessage(fmt.Sprintf("Cannot load history page. Error getting transaction count for filter %s: %s", selectedTxFilter, txCountErr.Error()), MessageKindError)
 			tviewApp.SetFocus(historyPage)
 			return historyPage
 		}
-		if txCount == 0 {
+		if filteredTxCount == 0 {
 			continue
 		}
 
-		transactionCountByFilter = append(transactionCountByFilter, fmt.Sprintf("%s (%d)", filter, txCount))
+		transactionFilters = append(transactionFilters, fmt.Sprintf("%s (%d)", filter, filteredTxCount))
 	}
 
 	if totalTxCount == 0 {
@@ -104,25 +106,19 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
 	historyPage.AddItem(txFilterDropdown, 2, 0, false)
 
 	historyTable := tview.NewTable().
-		SetBorders(false).
-		SetFixed(1, 0).
-		SetSelectable(true, false)
+	SetBorders(false).
+	SetFixed(1, 0).
+	SetSelectable(true, false)
 
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
 	// filter dropDown sleection options
-	txFilterDropdown.AddDropDown("", transactionCountByFilter, 0, func(filterAndCount string, index int) {
+	txFilterDropdown.AddDropDown("", transactionFilters, 0, func(filterAndCount string, index int) {
 		lastSelectedFilter = selectedFilter
 		selectedFilterAndCount := strings.Split(filterAndCount, " ")
 		selectedFilter = selectedFilterAndCount[0]
 
 		filter := walletcore.BuildTransactionFilter(selectedFilter)
-		_, txCountErr := wallet.TransactionCount(filter)
-		if txCountErr != nil {
-			displayMessage(txCountErr.Error(), ErrorMessage)
-			return
-		}
-
 		if selectedFilter != lastSelectedFilter {
 			if selectedFilter == "All"{
 				go fetchAndDisplayTransactions(0, wallet, nil, historyTable, tviewApp, displayMessage)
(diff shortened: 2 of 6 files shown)


terminal/pages/history.go
```diff
@@ -45,48 +48,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	transactionCountByFilter := make([]string, len(filters))
+
+	var txCount int
+	var txCountErr error
+	for index, filter := range filters {
+		if filter == "All" {
+			txCount, txCountErr = wallet.TransactionCount(nil)
+			totalTxCount = txCount
+		}
+		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
```
barisere: This will be executed even if the previous `if` block executes. Is that intended?

cli/commands/history.go
```diff
@@ -71,7 +71,7 @@ func (history HistoryCommand) Run(ctx context.Context, wallet walletcore.Wallet)
 			pageTxRows[i] = append(pageTxRows[i], formatFee(tx.Fee))
 			pageTxRows[i] = append(pageTxRows[i], tx.Type)
 		}
-
+		fmt.Println(len(history.displayedTxHashes))
```
itswisdomagain: I bet this isn't needed.

terminal/pages/history.go
```diff
@@ -45,48 +48,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+
+	// tx filter
+	filters := walletcore.TransactionFilters
+	transactionCountByFilter := make([]string, len(filters))
+
+	var txCount int
+	var txCountErr error
+	for index, filter := range filters {
+		if filter == "All" {
+			txCount, txCountErr = wallet.TransactionCount(nil)
```
itswisdomagain: Should check if `txCountErr` is not nil here and `break` or `return` from this loop instead of continuing to count txs for other filters, similar to what is done in lines 73 to 76

terminal/pages/history.go
itswisdomagain: Using `:=` here means you're creating and assigning to new variables and not the variables declared before the loop. Is this intentional?

terminal/pages/history.go
```diff
@@ -45,48 +48,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+			totalTxCount = txCount
+		}
+		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		if txCountErr != nil {
+			displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), true)
+			tviewApp.SetFocus(historyPage)
+			return historyPage
+		}
+		if txCount == 0 {
+			continue
```
itswisdomagain: This means `transactionCountByFilter[index]` will be empty.
Assuming there are 5 filters, line 63 above creates a `transactionCountByFilter` slice with 5 items. By default, each item's value is "".
Assume again that the filter at index 4 returns 0 count, transactionCountByFilter[4] will remain "". I'm pretty sure that's not what's intended.

You should use `append` instead.

itswisdomagain (changes requested):
Will continue review after these.

terminal/pages/history.go
```diff
@@ -36,7 +39,7 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
 	displayMessage := func(message string, error bool) {
 		// this function may be called from a goroutine, use tviewApp.QueueUpdateDraw
 		tviewApp.QueueUpdateDraw(func() {
-			body.RemoveItem(messageTextView)
+			historyPage.RemoveItem(messageTextView)
 			if message != "" {
 				if error {
```
barisere: `error` is also the name of a built-in interface. I suggest you rename the parameter to `isError`
Also, considering that this is a function, passing a boolean at its call site will be cryptic. An alternative is to use a type to describe the error condition instead. 
```go
type MessageKind string
const (
	ErrorMessage MessageKind = "error"
	InfoMessage MessageKind = "info"
)

displayMessage := func(message string, kind MessageKind) {
	// this function may be called from a goroutine, use tviewApp.QueueUpdateDraw
	tviewApp.QueueUpdateDraw(func() {
		historyPage.RemoveItem(messageTextView)
		if message != "" && kind == ErrorMessage {
			// other stuff
		}
	})
}
```

terminal/pages/history.go
```diff
@@ -45,48 +48,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 		SetFixed(1, 0).
 		SetSelectable(true, false)
 
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
+	// filter dropDown slection options
+	txDropdown.AddDropDown("", transactionCountByFilter, 0, func(filterAndCount string, index int) {
+		selectedFilterAndCount := strings.Split(filterAndCount, " ")
+		lastSelectedFilter = selectedFilter
+		selectedFilter = selectedFilterAndCount[0]
```
barisere: As a more idiomatic Go style, the previous three lines can be collapsed into the following.
```go
selectedFilter, lastSelectedFilter := strings.Split(filterAndCount, " ")[0], selectedFilter
```

terminal/pages/history.go
```diff
@@ -121,41 +185,60 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 
-	tviewApp.SetFocus(body)
+	tviewApp.SetFocus(historyPage)
 
-	return body
+	return historyPage
 }
 
-func fetchAndDisplayTransactions(txOffset int, wallet walletcore.Wallet, historyTable *tview.Table, tviewApp *tview.Application, displayMessage func(string, bool)) {
+func fetchAndDisplayTransactions(txOffset int, wallet walletcore.Wallet, filter *txindex.ReadFilter, historyTable *tview.Table, tviewApp *tview.Application, displayMessage func(string, bool)) {
```
barisere: The function signature is quite long. Break the parameter list onto several lines.
```go
func fetchAndDisplayTransactions(txOffset int,
	wallet walletcore.Wallet,
	filter *txindex.ReadFilter,
	historyTable *tview.Table,
	tviewApp *tview.Application,
	displayMessage func(string, bool)) {}
```

barisere (changes requested):
Several functions are defined within the `historyPage` function. Extract those that can stand alone (such as `displayMessage`) into function definitions at the top level. This should shorten the length of the `historyPage` function. The rationale for this is that the `historyPage` function should return the view primitive, and its helper functions can stand alone. The `displayMessage` function can become
```go
func displayMessage(tviewApp *tview.Application, message string, kind MessageKind) {
    // ...
}
```
See whether similar refactoring can be done for other functions, where necessary.

Also rebase your commits onto master, then see whether you can follow the changes made in #412. The entry point for building the CLI binary should be in the _cmd_ directory. The resulting binary should be named _godcr-tui_.

terminal/pages/history.go
Sirmorrison: doesn't work as expected.

terminal/pages/history.go
Sirmorrison: yes.

terminal/pages/history.go
barisere: How is it expected to work? I tried it and it works as I expect.

terminal/pages/history.go
barisere: Please take a look at this function signature, it's gotten even longer. It can be written on at least two lines.
```go
func fetchAndDisplayTransactions(txOffset int, wallet walletcore.Wallet, filter *txindex.ReadFilter,
	historyTable *tview.Table, tviewApp *tview.Application, displayMessage func(string, bool)) {
        // stuff here
}
```
@Sirmorrison

terminal/pages/history.go
barisere: @Sirmorrison You may have missed this. Please attend to it.

barisere (changes requested):
Previous reviews have not been attended to completely. Please go through the comments above and resolve any issue left.

Sirmorrison:
> ```go
> func displayMessage(tviewApp *tview.Application, message string, kind MessageKind) {
>     // ...
> }
> ```

The reason for the inline function used for the displayMessage().. is so as to have access to the `historyPage` primitive.

terminal/pages/history.go
Sirmorrison: Your approach yields the following results from the screenshots below. The idea is to clear the table before adding new data. But your approach doesn't seem be doing that.

image

image

image

What I want Is seen in the screenshots below. As it clears the table before populating it with selctedFilter data.

image

image

image

barisere (approved):
LGTM

terminal/pages/history.go
```diff
@@ -17,13 +19,19 @@ import (
 var displayedTxHashes []string
 var txPerPage int32 = walletcore.TransactionHistoryCountPerPage
 var totalTxCount int
+var selectedFilter, lastSelectedFilter string
+	type MessageKind string
```
itswisdomagain: This isn't properly formatted. Also, the following enum pattern is better as the values for the constants are not needed:

```
type messageKind string
const (
	ErrorMessage messageKind = iota
	InfoMessage messageKind
)
```

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	messageTextView := primitives.WordWrappedTextView("")
-	displayMessage := func(message string, error bool) {
+	displayMessage := func(message string, kind MessageKind) {
 		// this function may be called from a goroutine, use tviewApp.QueueUpdateDraw
 		tviewApp.QueueUpdateDraw(func() {
-			body.RemoveItem(messageTextView)
+			historyPage.RemoveItem(messageTextView)
 			if message != "" {
-				if error {
+				if message != "" && kind == ErrorMessage {
```
itswisdomagain: the preceding if statement already checks if `message != ""`

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	}
 
-	// get total tx count early on, so as to display on history header
-	txCount, txCountErr := wallet.TransactionCount(nil)
-	totalTxCount = txCount
-
 	// page title
-	historyPageTitle := fmt.Sprintf("History (%d transactions)", txCount)
+	historyPageTitle := "History"
 	titleTextView := primitives.NewLeftAlignedTextView(historyPageTitle)
```
itswisdomagain: Can inline the `"History"` text instead of using a `historyPageTitle` variable.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
-	body.AddItem(titleTextView, 2, 0, false)
+	historyPage.AddItem(titleTextView, 2, 0, false)
+
+	// tx filter
+	filters := walletcore.TransactionFilters
```
itswisdomagain: Use `walletcore.TransactionFilters` directly please.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	var transactionCountByFilter []string
```
itswisdomagain: A count should be a slice of numbers not string. Rename your variable, this slice does not actually hold counts, it holds texts.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+			return historyPage
+		}
+		if txCount == 0 {
+			continue
+		}
+
+		transactionCountByFilter = append(transactionCountByFilter, fmt.Sprintf("%s (%d)", filter, txCount))
+	}
 
 	if txCountErr != nil {
```
itswisdomagain: All cases of `txCountErr != nil` are already handled in the for loop above. This if statement is guaranteed to always return false, and is thus unneeded.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	titleTextView := primitives.NewLeftAlignedTextView(historyPageTitle)
-	body.AddItem(titleTextView, 2, 0, false)
+	historyPage.AddItem(titleTextView, 2, 0, false)
+
+	// tx filter
+	filters := walletcore.TransactionFilters
+	var transactionCountByFilter []string
+
+	var txCount int
+	var txCountErr error
```
itswisdomagain: These 2 variables do not need to be declared outside the for loop as they're really not needed outside the loop.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	}
 
+	if totalTxCount == 0 {
+		displayMessage("No transactions yet", InfoMessage)
+		hintTextView.SetText("TIP: ESC or BACKSPACE to return to navigation menu")
+		tviewApp.SetFocus(historyPage)
+		return historyPage
+	}
+
+	txDropdown := primitives.NewForm(false)
```
itswisdomagain: `txFilterDropdown`

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+
+		if selectedFilter == "All" {
+			if selectedFilter != lastSelectedFilter {
+				go fetchAndDisplayTransactions(0, wallet, nil, historyTable, tviewApp, displayMessage)
+			}
+			return
+		}
+
+		filter := txindex.Filter()
+		filter = walletcore.BuildTransactionFilter(selectedFilter)
```
itswisdomagain: You do recognize that those 2 lines above is same as `filter := walletcore.BuildTransactionFilter(selectedFilter)`.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
+	// filter dropDown sleection options
+	txDropdown.AddDropDown("", transactionCountByFilter, 0, func(filterAndCount string, index int) {
+		selectedFilterAndCount := strings.Split(filterAndCount, " ")
+		lastSelectedFilter = selectedFilter
+		selectedFilter = selectedFilterAndCount[0]
+
+		if selectedFilter == "All" {
+			if selectedFilter != lastSelectedFilter {
```
itswisdomagain: This check should happen whether selected filter is all or not. Regardless of what selected filter is, if it's the same as the last selected filter, nothing should happen.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 		SetBorders(false).
 		SetFixed(1, 0).
 		SetSelectable(true, false)
 
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
+	// filter dropDown sleection options
+	txDropdown.AddDropDown("", transactionCountByFilter, 0, func(filterAndCount string, index int) {
+		selectedFilterAndCount := strings.Split(filterAndCount, " ")
+		lastSelectedFilter = selectedFilter
```
itswisdomagain: Swap this line with the preceding line. The `selectedFilterAndCount` is created to be used in line 125. There shouldn't be any line of code separating the creation of that variable and it's usage.

terminal/pages/history.go
```diff
@@ -33,60 +41,127 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+			if selectedFilter != lastSelectedFilter {
+				go fetchAndDisplayTransactions(0, wallet, nil, historyTable, tviewApp, displayMessage)
+			}
+			return
+		}
+
+		filter := txindex.Filter()
+		filter = walletcore.BuildTransactionFilter(selectedFilter)
+
+		totalTxCount, txCountErr = wallet.TransactionCount(filter)
```
itswisdomagain: `totalTxCount`? What if filter is sent filter? Would this still return totalTxCount? Why do you even count the transactions for this filter? I don't see you using the count anywhere.

itswisdomagain (changes requested):
Will review again after these.

terminal/pages/history.go
Sirmorrison: image


the reason we count is for the dropdown.

terminal/pages/history.go
Sirmorrison: It does regardless of the filter. Only thing is the `fetchAndDisplayTransactions()` parameter takes a `nil` filter instead of an actual filter as seen in line 141.

terminal/pages/history.go
Sirmorrison: They are used outside the loop.

terminal/pages/history.go
Sirmorrison: image

I get this error.

terminal/pages/history.go
itswisdomagain: You first check if the selected filter is all BEFORE checking if the selected filter is the last selected filter. That should not be so.

terminal/pages/history.go
itswisdomagain: I didn't say they are not used outside. I said they are not REALLY NEEDED outside the loop.

terminal/pages/history.go
itswisdomagain: When you reply to a message without applying the suggestions, DO NOT resolve the conversation so I can easily see your reply. `type messageKind string` -> `type messageKind uint32`

terminal/pages/history.go
itswisdomagain: Look again. Do you really use that `totalTxCount` variable? Does that variable really hold the count for all transactions or for a set of transactions based on the filter selected?

itswisdomagain (changes requested):

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
-		tviewApp.SetFocus(body)
-		return body
+	titleTextView := primitives.NewLeftAlignedTextView("History")
+	historyPage.AddItem(titleTextView, 2, 0, false)
+
+	// tx filter
+	var transactionCountByFilter []string
+
+	// var txCount int
+	// var txCountErr error
```
itswisdomagain: Please remove.
Also no need for an empty line space between the declaration of `transactionCountByFilter` and this loop.
And again, I believe a previous review comment advised renaming the `transactionCountByFilter` variable. It does not hold COUNTS!!! If it did, the type would be `[]int` not `[]string`.

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	for _, filter := range walletcore.TransactionFilters {
+		if filter == "All" {
+			txCount, txCountErr := wallet.TransactionCount(nil)
+			if txCountErr != nil {
+				displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), ErrorMessage)
+				tviewApp.SetFocus(historyPage)
+				return historyPage
+			}
+
+			totalTxCount = txCount
```
itswisdomagain: There's no `continue` or `break` here. What this means is, if the filter is "All', the `wallet.TransactionCount` method will be called twice. Obviously not good.

Instead of calling it twice, call it once (below) and use do the following after calling the tx count function:
```
if filter == "All" {
      totalTxCount = txCount
}
```

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		}
+
+		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		if txCountErr != nil {
+			displayMessage(fmt.Sprintf("Cannot load history page. Error getting total transaction count: %s", txCountErr.Error()), ErrorMessage)
```
itswisdomagain: This is not an error getting TOTAL transaction count. Replace with
```suggestion
			displayMessage(fmt.Sprintf("Cannot load history page. Error getting transaction count for filter %s: %s", filter, txCountErr.Error()), ErrorMessage)
```

Also, reading this line of code makes it obvious that the variable `ErrorMessage` is poorly named. Rename to `MessageKindError`

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 	historyTable := tview.NewTable().
 		SetBorders(false).
 		SetFixed(1, 0).
 		SetSelectable(true, false)
 
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
+	// filter dropDown sleection options
+	txFilterDropdown.AddDropDown("", transactionCountByFilter, 0, func(filterAndCount string, index int) {
+		lastSelectedFilter = selectedFilter
```
itswisdomagain: You should first check if `lastSelectedFilter == selectedFilter` and return if so before making this assignment. Doing this check early on prevents making any further attempt to load txs for the newly selected filter.

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		selectedFilterAndCount := strings.Split(filterAndCount, " ")
+		selectedFilter = selectedFilterAndCount[0]
+
+		filter := walletcore.BuildTransactionFilter(selectedFilter)
+		_, txCountErr := wallet.TransactionCount(filter)
```
itswisdomagain: I see absolutely no reason to count the txs for this filter only to see if an error will occur. Remove please.

terminal/pages/history.go
```diff
@@ -33,60 +41,107 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		if txCountErr != nil {
+			displayMessage(txCountErr.Error(), ErrorMessage)
+			return
+		}
+
+		if selectedFilter != lastSelectedFilter {
+			if selectedFilter == "All"{
+				go fetchAndDisplayTransactions(0, wallet, nil, historyTable, tviewApp, displayMessage)
+			}else{
```
itswisdomagain: go fmt please
Always go fmt before committing.

itswisdomagain (changes requested):
Will review again after these.

terminal/pages/history.go
Sirmorrison: This is a cyclic process. there is not need checked if they are same. line 127 does the checks that is needed.

terminal/pages/history.go
```diff
@@ -17,13 +19,19 @@ import (
 var displayedTxHashes []string
 var txPerPage int32 = walletcore.TransactionHistoryCountPerPage
 var totalTxCount int
+var selectedFilter, lastSelectedFilter string
+type messageKind uint32
+const (
+	MessageKindError messageKind = iota
+	InfoMessage messageKind = iota
```
itswisdomagain: Did I really need to specifically mention this? `InfoMessage` -> `MessageKindInfo`

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 
 				messageTextView.SetText(message)
-				body.AddItem(messageTextView, 2, 0, false)
+				historyPage.AddItem(messageTextView, 2, 0, false)
 			}
 		})
 	}
 
-	// get total tx count early on, so as to display on history header
+	// page title
```
itswisdomagain: This comment is really unnecessary.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+			return historyPage
+		}
+		if filteredTxCount == 0 {
+			continue
+		}
+
+		transactionFilters = append(transactionFilters, fmt.Sprintf("%s (%d)", filter, filteredTxCount))
+	}
+
+	if totalTxCount == 0 {
```
itswisdomagain: This check can be moved to after line 72. There's no point doing anything further if totalTxCount is 0.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		tviewApp.SetFocus(historyPage)
+		return historyPage
+	}
 	totalTxCount = txCount
 
-	// page title
-	historyPageTitle := fmt.Sprintf("History (%d transactions)", txCount)
-	titleTextView := primitives.NewLeftAlignedTextView(historyPageTitle)
-	body.AddItem(titleTextView, 2, 0, false)
+	// tx filter
```
itswisdomagain: Unnecessary comment. Variable names already tell us what's going on here.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	var transactionFilters []string
+	var selectedTxFilter *txindex.ReadFilter
```
itswisdomagain: This isn't needed aside the loop, at least it SHOULDN'T be. Move it to inside the loop.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+			selectedTxFilter = nil
+		} else {
+			selectedTxFilter = walletcore.BuildTransactionFilter(filter)
+		}
 
-	if txCountErr != nil {
-		displayMessage(fmt.Sprintf("Cannot load history. Get total tx count error: %s", txCountErr.Error()), true)
-		tviewApp.SetFocus(body)
-		return body
+		filteredTxCount, txCountErr := wallet.TransactionCount(selectedTxFilter)
```
itswisdomagain: This isn't the selected tx filter. At this point, no selection has been made. Rename the variable please.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		if txCountErr != nil {
+			displayMessage(fmt.Sprintf("Cannot load history page. Error getting transaction count for filter %s: %s", selectedTxFilter, txCountErr.Error()), MessageKindError)
```
itswisdomagain: `selectedTxFilter` -> `filter`

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	txFilterDropdown.SetBorderPadding(0, 0, 0, 0)
+	historyPage.AddItem(txFilterDropdown, 2, 0, false)
+
 	historyTable := tview.NewTable().
-		SetBorders(false).
-		SetFixed(1, 0).
-		SetSelectable(true, false)
+	SetBorders(false).
+	SetFixed(1, 0).
+	SetSelectable(true, false)
```
itswisdomagain: The indenting in lines 109 to 110 are off.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 
 	transactionDetailsTable := tview.NewTable().SetBorders(false)
 
+	// filter dropDown sleection options
```
itswisdomagain: - typo
- also don't know what this comment describes

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+	txFilterDropdown.AddDropDown("", transactionFilters, 0, func(filterAndCount string, index int) {
+		lastSelectedFilter = selectedFilter
```
itswisdomagain: In my comment here https://github.com/raedahgroup/godcr/pull/383/files?file-filters%5B%5D=.go#r328647081, I said you should FIRST check and return. Check early. I did not say the check wasn't done below, I said the check should be done earlier. This isn't implemented and there's no good reason for not implementing it.

terminal/pages/history.go
```diff
@@ -33,60 +41,103 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
+		selectedFilterAndCount := strings.Split(filterAndCount, " ")
+		selectedFilter = selectedFilterAndCount[0]
+
+		filter := walletcore.BuildTransactionFilter(selectedFilter)
+		if selectedFilter != lastSelectedFilter {
+			if selectedFilter == "All"{
+				go fetchAndDisplayTransactions(0, wallet, nil, historyTable, tviewApp, displayMessage)
+			}else{
```
itswisdomagain: You still have not done go fmt

terminal/pages/history.go
```diff
@@ -121,150 +172,191 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
 
 		return event
 	})
 
-	tableHeaderCell := func(text string) *tview.TableCell {
-		return tview.NewTableCell(text).SetAlign(tview.AlignCenter).SetSelectable(false).SetMaxWidth(1).SetExpansion(1)
-	}
+	// handler for switching between dropDown and table
+	txFilterDropdown.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+		if event.Key() == tcell.KeyEscape || event.Key() == tcell.KeyBackspace || event.Key() == tcell.KeyBackspace2 {
```
itswisdomagain: Do we clear focus on backspace? iirc, we only do so on escape key press.

terminal/pages/history.go
```diff
@@ -121,150 +172,191 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
-	historyTable.SetCell(0, 1, tableHeaderCell(fmt.Sprintf("%10s", "Direction")))
-	historyTable.SetCell(0, 2, tableHeaderCell(fmt.Sprintf("%8s", "Amount")))
-	historyTable.SetCell(0, 3, tableHeaderCell(fmt.Sprintf("%5s", "Status")))
-	historyTable.SetCell(0, 4, tableHeaderCell(fmt.Sprintf("%-5s", "Type")))
+		return event
+	})
+
+	// handler for switching between dropDown and table
+	historyTable.SetInputCapture(func(event *tcell.EventKey) *tcell.EventKey {
+		if event.Key() == tcell.KeyEscape || event.Key() == tcell.KeyBackspace || event.Key() == tcell.KeyBackspace2 {
```
itswisdomagain: My previous comment applies here as well.

terminal/pages/history.go
```diff
@@ -95,23 +146,23 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
 		}
 	})
 
-	displayedTxHashes = []string{}
```
itswisdomagain, line 98: Why is this removed?

terminal/pages/history.go
```diff
@@ -121,150 +172,191 @@ func historyPage(wallet walletcore.Wallet, hintTextView *primitives.TextView, tv
...
-func fetchAndDisplayTransactions(txOffset int, wallet walletcore.Wallet, historyTable *tview.Table, tviewApp *tview.Application, displayMessage func(string, bool)) {
-	// show a loading text at the bottom of the table so user knows an op is in progress
-	displayMessage("Fetching data...", false)
+func fetchAndDisplayTransactions(txOffset int, 
+	wallet walletcore.Wallet, 
+	filter *txindex.ReadFilter, 
+	historyTable *tview.Table, 
+	tviewApp *tview.Application, 
+	displayMessage func(string, messageKind)) {
+		// show a loading text at the bottom of the table so user knows an op is in progress
```
itswisdomagain: Add an empty line space here because this function parameters list extends beyond 1 line. The extra line makes it easy to tell where the function signature stops and where the function body begins.

itswisdomagain (changes requested):
There are probably more changes to be requested, but I'm unwilling to review further as there's been too many changes yet to be made. I'm taking up this PR to complete properly.

itswisdomagain (approved):

PR #384, opened 2019-06-19
Nuklear ui tx filter

### LINKED ISSUE #378: history page transaction filter

Sirmorrison:
fixes #378
Added tx filter to nuklear ui, also added a message to be displayed when there is no transaction to be displayed instead of a blank screen.

image

image

### DIFF
--- a/nuklear/pagehandlers/history.go
+++ b/nuklear/pagehandlers/history.go
@@ -48,10 +48,10 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
 
 	handler.clearTxDetails()
 
-	// handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
-	// if handler.fetchHistoryError != nil {
-	// 	return true
-	// }
+	handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
+	if handler.fetchHistoryError == nil {
+		go handler.fetchTransactions(handler.filter)
+	}
 
 	return true
 }
@@ -89,22 +89,6 @@ func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 		handler.filterSelectorWidget.Render(contentWindow)
 
 		handler.totalTxCount, handler.filter = handler.filterSelectorWidget.GetSelectedFilter()
-		// handler.selectedFilter = selectedFilterAndCount[0]
-		// fmt.Println(handler.selectedFilter)
-
-		// if handler.selectedFilter == "All" {
-		// 	go handler.fetchTransactions(nil)
-		// }else{
-		// 	handler.filter = txindex.Filter()
-		// 	handler.filter = walletcore.BuildTransactionFilter(handler.selectedFilter)
-
-		// 	handler.totalTxCount, handler.fetchHistoryError = handler.wallet.TransactionCount(handler.filter)
-		// 	if handler.fetchHistoryError != nil {
-		// 		return
-		// 	}
-
-		go handler.fetchTransactions(handler.filter)
-		// }
 
 		// show transactions first, if any
 		if len(handler.transactions) > 0 {
--- a/nuklear/pagehandlers/history.go
+++ b/nuklear/pagehandlers/history.go
@@ -21,6 +21,7 @@ type HistoryHandler struct {
 	filter               *txindex.ReadFilter
 
 	totalTxCount int
+	txCountErr               error
 
 	currentPage            int
 	txPerPage              int
@@ -37,13 +38,37 @@ type HistoryHandler struct {
 func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *config.Settings, refreshWindowDisplay func()) bool {
 	handler.wallet = wallet
 	handler.refreshWindowDisplay = refreshWindowDisplay
-
-	handler.filterSelectorWidget = widgets.FilterSelectorWidget(wallet)
 	handler.currentPage = 1
 	handler.txPerPage = walletcore.TransactionHistoryCountPerPage
 	handler.transactions = nil
 	handler.fetchHistoryError = nil
 	handler.isFetchingTransactions = false
+
+	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, nil)
+
+	// fetch initial table data
+	handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
+	if handler.fetchHistoryError == nil {
+		// only fetch txs if there was no error getting tx count.
+		go handler.fetchTransactions(nil)
+	}
+
+	// fetch filtered data
+	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, func(int){
+		totalTxCount, filter, selectedFilter, err := handler.filterSelectorWidget.GetSelectedFilter()
+		handler.totalTxCount = totalTxCount
+		if err != nil {
+			handler.fetchHistoryError = err
+			return
+		}
+
+		if selectedFilter != handler.selectedFilter {
+			handler.selectedFilter = selectedFilter
+			handler.filter = filter
+			handler.transactions = nil
+			go handler.fetchTransactions(filter)
+		}
+	})
 
 	handler.clearTxDetails()
 
@@ -80,20 +105,8 @@ func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
 		handler.filterSelectorWidget.Render(contentWindow)
 
-		totalTxCount, filter, selectedFilter, err := handler.filterSelectorWidget.GetSelectedFilter()
-		handler.totalTxCount = totalTxCount
-
-		if err != nil {
-			contentWindow.DisplayErrorMessage("Error loading history", err)
-			return
-		}
-
-		if selectedFilter != handler.selectedFilter {
-			handler.selectedFilter = selectedFilter
-			handler.filter = filter
-			handler.transactions = nil
-			go handler.fetchTransactions(filter)
-			window.Master().Changed()
+		if handler.txCountErr != nil {
+			contentWindow.DisplayErrorMessage("Error loading history", handler.fetchHistoryError)
 		}
 
 		if len(handler.transactions) == 0 {
--- a/nuklear/pagehandlers/history.go
+++ b/nuklear/pagehandlers/history.go
@@ -14,6 +14,7 @@ import (
 type HistoryHandler struct {
 	wallet               walletcore.Wallet
 	refreshWindowDisplay func()
+	filterSelectorWidget *widgets.FilterSelector
 
 	totalTxCount int
 
@@ -32,6 +33,8 @@ type HistoryHandler struct {
 func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *config.Settings, refreshWindowDisplay func()) bool {
 	handler.wallet = wallet
 	handler.refreshWindowDisplay = refreshWindowDisplay
+
+	handler.filterSelectorWidget = widgets.FilterSelectorWidget(wallet)
 
 	handler.currentPage = 1
 	handler.txPerPage = walletcore.TransactionHistoryCountPerPage
@@ -77,6 +80,10 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
 
 func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
+		handler.filterSelectorWidget = widgets.FilterSelectorWidget(handler.wallet)
+		handler.filterSelectorWidget.Render(contentWindow)
+		handler.refreshWindowDisplay()
+
 		// show transactions first, if any
 		if len(handler.transactions) > 0 {
 			handler.displayTransactions(contentWindow)
--- a/nuklear/pagehandlers/history.go
+++ b/nuklear/pagehandlers/history.go
@@ -17,11 +17,11 @@ type HistoryHandler struct {
 	refreshWindowDisplay func()
 
 	filterSelectorWidget *widgets.FilterSelector
-	selectedFilter       string
-	filter               *txindex.ReadFilter
+	currentFilterText       string
+	selectedTxFilter               *txindex.ReadFilter
 
 	totalTxCount int
-	txCountErr               error
+	filterSelectorErr               error
 
 	currentPage            int
 	txPerPage              int
@@ -44,7 +44,9 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
 	handler.fetchHistoryError = nil
 	handler.isFetchingTransactions = false
 
-	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, nil)
+	handler.clearTxDetails()
+
+	handler.filterSelectorWidget, handler.filterSelectorErr = widgets.FilterSelectorWidget(wallet, nil)
 
 	// fetch initial table data
 	handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
@@ -54,23 +56,20 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
 	}
 
 	// fetch filtered data
-	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, func(int){
-		totalTxCount, filter, selectedFilter, err := handler.filterSelectorWidget.GetSelectedFilter()
-		handler.totalTxCount = totalTxCount
+	handler.filterSelectorWidget, handler.filterSelectorErr = widgets.FilterSelectorWidget(wallet, func(){
+		_, selectedTxFilter, currentFilterText, err := handler.filterSelectorWidget.GetSelectedFilter()
 		if err != nil {
-			handler.fetchHistoryError = err
+			handler.filterSelectorErr = err
 			return
 		}
 
-		if selectedFilter != handler.selectedFilter {
-			handler.selectedFilter = selectedFilter
-			handler.filter = filter
+		if currentFilterText != handler.currentFilterText {
+			handler.currentFilterText = currentFilterText
+			handler.selectedTxFilter = selectedTxFilter
 			handler.transactions = nil
-			go handler.fetchTransactions(filter)
+			go handler.fetchTransactions(selectedTxFilter)
 		}
 	})
-
-	handler.clearTxDetails()
 
 	return true
 }
@@ -80,7 +79,6 @@ func (handler *HistoryHandler) fetchTransactions(filter *txindex.ReadFilter) {
 	handler.refreshWindowDisplay() // refresh display to show loading indicator
 
 	txHistoryOffset := 0
-
 	if handler.transactions != nil {
 		txHistoryOffset = len(handler.transactions)
 	}
@@ -105,8 +103,8 @@ func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
 		handler.filterSelectorWidget.Render(contentWindow)
 
-		if handler.txCountErr != nil {
-			contentWindow.DisplayErrorMessage("Error loading history", handler.fetchHistoryError)
+		if handler.filterSelectorErr != nil {
+			contentWindow.DisplayErrorMessage("Error with filter selector", handler.filterSelectorErr)
 		}
 
 		if len(handler.transactions) == 0 {
@@ -193,7 +191,7 @@ func (handler *HistoryHandler) loadNextPage(window *widgets.Window) {
 	nextPageTxOffset := (nextPage - 1) * handler.txPerPage
 	if nextPageTxOffset >= len(handler.transactions) {
 		// we've not loaded txs for this page
-		go handler.fetchTransactions(handler.filter)
+		go handler.fetchTransactions(handler.selectedTxFilter)
 	}
 
 	handler.refreshWindowDisplay()
--- a/nuklear/pagehandlers/history.go
+++ b/nuklear/pagehandlers/history.go
@@ -2,6 +2,7 @@ package pagehandlers
 
 import (
 	"fmt"
+	// "strings"
 
 	"github.com/aarzilli/nucular"
 	"github.com/decred/dcrd/dcrutil"
@@ -14,7 +15,10 @@ import (
 type HistoryHandler struct {
 	wallet               walletcore.Wallet
 	refreshWindowDisplay func()
+
 	filterSelectorWidget *widgets.FilterSelector
+	// selectedFilter string
+	filter *txindex.ReadFilter
 
 	totalTxCount int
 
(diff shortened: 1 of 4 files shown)


nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,106 @@
...
+
+	for index, filter := range filterSelector.filters {
+		if filter == "All" {
+			filterSelector.filter = filter
+			filterSelector.txCount, filterSelector.txCountErr = wallet.TransactionCount(nil)
+			filterSelector.totalTxCount = filterSelector.txCount
+		}
+
+		filterSelector.txCount, filterSelector.txCountErr = wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		if filterSelector.txCountErr != nil {
```
barisere: This technique is returning an error implicitly. The `FilterSelectorWidget` function signature does not indicate the possibility of an error, so the caller has to be aware of the `FilterSelector.txCountErr` property. Can you make the error condition explicit? Preferably, modify `FilterSelectorWidget` to return `(*FilterSelector, error)`.

barisere (changes requested):

nuklear/pagehandlers/history.go
```diff
@@ -77,10 +77,30 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
...
+		if err != nil {
+			contentWindow.DisplayErrorMessage("Error loading history", err)
+			return
+		}
+
+		if selectedFilter != handler.selectedFilter {
+			handler.selectedFilter = selectedFilter
+			handler.filter = filter
+			handler.transactions = nil
+			go handler.fetchTransactions(filter)
```
itswisdomagain: Wallet operations should not be performed directly from the `Render` function, use the `BeforeRender` function instead. `Render` should only be for drawing/rendering views on the window; it is called repeatedly whenever there's a slight action such as moving the mouse. As a matter of fact, a single mouse move could cause the `Render` function to be called multiple times in a second.

The only safe places to call wallet functions is in `BeforeRender` and in event listeners such as button click event listeners.

itswisdomagain (changes requested):
I'll review again after this important change is addressed.

nuklear/pagehandlers/history.go
```diff
@@ -9,13 +9,19 @@ import (
...
 type HistoryHandler struct {
 	wallet               walletcore.Wallet
 	refreshWindowDisplay func()
 
+	filterSelectorWidget *widgets.FilterSelector
+	selectedFilter       string
+	filter               *txindex.ReadFilter
+
 	totalTxCount int
+	txCountErr               error
```
itswisdomagain: go fmt

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
 	handler.refreshWindowDisplay = refreshWindowDisplay
-
 	handler.currentPage = 1
 	handler.txPerPage = walletcore.TransactionHistoryCountPerPage
 	handler.transactions = nil
 	handler.fetchHistoryError = nil
 	handler.isFetchingTransactions = false
 
-	handler.clearTxDetails()
+	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, nil)
```
itswisdomagain: This call is rendered obsolete by similar function call in line 57. Please delete.

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
+	// fetch initial table data
 	handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
 	if handler.fetchHistoryError == nil {
 		// only fetch txs if there was no error getting tx count.
-		go handler.fetchTransactions()
+		go handler.fetchTransactions(nil)
 	}
 
+	// fetch filtered data
+	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, func(int){
```
itswisdomagain: Why is the error returned from the `widgets.FilterSelectorWidget` function assigned to `txCountErr`? The relattionship between tx count and FilterSelectorWidget (if such a relationship even exists) is not immediately obvious.

Should probably rename `txCountErr` to `filterSelectorErr`.

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
+		totalTxCount, filter, selectedFilter, err := handler.filterSelectorWidget.GetSelectedFilter()
+		handler.totalTxCount = totalTxCount
+		if err != nil {
+			handler.fetchHistoryError = err
+			return
+		}
+
+		if selectedFilter != handler.selectedFilter {
+			handler.selectedFilter = selectedFilter
+			handler.filter = filter
```
itswisdomagain: Rename `filter` and `handler.filter` to `selectedTxFilter` and `handler.selectedTxFilter` respectively. Also rename `selectedFilter` and `handler.selectedFilter` to `currentFilterText` and `handler.currentFilterText` respectively.

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
 	if handler.fetchHistoryError == nil {
 		// only fetch txs if there was no error getting tx count.
-		go handler.fetchTransactions()
+		go handler.fetchTransactions(nil)
 	}
 
+	// fetch filtered data
+	handler.filterSelectorWidget, handler.txCountErr = widgets.FilterSelectorWidget(wallet, func(int){
+		totalTxCount, filter, selectedFilter, err := handler.filterSelectorWidget.GetSelectedFilter()
+		handler.totalTxCount = totalTxCount
```
itswisdomagain: Does the `totalTxCount` count change? `handler.totalTxCount` is already assigned in line 50; don't see why it needs to be returned by `filterSelectorWidget.GetSelectedFilter()`.

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
+		if err != nil {
+			handler.fetchHistoryError = err
```
itswisdomagain: `err` isn't an history fetch error, is it?

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
+
+		if selectedFilter != handler.selectedFilter {
+			handler.selectedFilter = selectedFilter
+			handler.filter = filter
+			handler.transactions = nil
+			go handler.fetchTransactions(filter)
+		}
+	})
+
+	handler.clearTxDetails()
```
itswisdomagain: You want to clear tx details before calling `fetchTransactions`. Restore this `handler.clearTxDetails()` call to after line 45.

nuklear/pagehandlers/history.go
```diff
@@ -32,34 +38,54 @@ type HistoryHandler struct {
...
 	return true
 }
 
-func (handler *HistoryHandler) fetchTransactions() {
+func (handler *HistoryHandler) fetchTransactions(filter *txindex.ReadFilter) {
 	handler.isFetchingTransactions = true
 	handler.refreshWindowDisplay() // refresh display to show loading indicator
 
 	txHistoryOffset := 0
+
```
itswisdomagain: This space is unnecessary. The next set of lines directly relate to the preceding line.

nuklear/pagehandlers/history.go
```diff
@@ -77,10 +103,18 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
 
 func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
-		// show transactions first, if any
-		if len(handler.transactions) > 0 {
+		handler.filterSelectorWidget.Render(contentWindow)
+
+		if handler.txCountErr != nil {
+			contentWindow.DisplayErrorMessage("Error loading history", handler.fetchHistoryError)
```
itswisdomagain: How is a tx **COUNT** error same as an error loading history? Oh by the way, you're referencing `handler.fetchHistoryError` in the if statement body but your if statement checked for `handler.txCountErr != nil`

nuklear/widgets/accountselector.go
```diff
@@ -66,7 +66,7 @@ func (accountSelector *AccountSelector) Render(window *Window, addColumns ...int
 
 	if accountSelector.accountsFetchError != nil {
 		window.DisplayErrorMessage("Fetch accounts error", accountSelector.accountsFetchError)
-	} else if len(accountSelector.accounts) == 1 {
+	} else if len(accountSelector.accounts) != 1 {
```
itswisdomagain: Why is this change necessary?

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/raedahgroup/dcrlibwallet/txindex"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/aarzilli/nucular/label"
+)
+
+type FilterSelector struct {
+	wallet walletcore.Wallet
+
+	filters                  []string
+	filter                   string
```
itswisdomagain: `selectedFilter`

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,135 @@
...
+	transactionCountByFilter []string
+	txCountErr               error
+	selectedFilterIndex      int
+	totalTxCount             int
+	selectedFilter           string
+	lastfilter               string
+	changed                  bool
+	selectedFilterAndCount   string
+	txFilter                 *txindex.ReadFilter
+	selectionChanged         func(int)
```
itswisdomagain: The `int` is not needed obviously, can just use a simple `func()`.

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,135 @@
...
+func FilterSelectorWidget(wallet walletcore.Wallet, selectionChanged func(int)) (filterSelector *FilterSelector, error error) {
+	filterSelector = &FilterSelector{
+		selectionChanged: selectionChanged,
+	}
+
+	filterSelector.wallet = wallet
+
+	filterSelector.filters = walletcore.TransactionFilters
+
+	filterSelector.transactionCountByFilter = make([]string, len(filterSelector.filters))
```
itswisdomagain: The empty line spaces at lines 41 and 43 can be deleted without hurting readability.

nuklear/pagehandlers/history.go
```diff
@@ -9,13 +9,19 @@ import (
...
+	"github.com/raedahgroup/dcrlibwallet/txindex"
 )
 
 type HistoryHandler struct {
 	wallet               walletcore.Wallet
 	refreshWindowDisplay func()
 
+	filterSelectorWidget *widgets.FilterSelector
+	selectedFilter       string
+	filter               *txindex.ReadFilter
```
itswisdomagain: Shouldn't need to save this variable in this struct. Only save the selected filter text and whenever you need the readfilter, you call `walletcore.BuildTransactionFilter()`

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,135 @@
...
+
+	for index, filter := range filterSelector.filters {
+		if filter == "All" {
+			filterSelector.filter = filter
+			filterSelector.txCount, filterSelector.txCountErr = wallet.TransactionCount(nil)
+			filterSelector.totalTxCount = filterSelector.txCount
+		}
+
+		filterSelector.txCount, filterSelector.txCountErr = wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
```
itswisdomagain: Each time this loop runs, `filterSelector.txCount` is re-assigned. The final value of `filterSelector.txCount` is the value set by the last loop run. I believe this variable does not need to be a property of the `filterSelector` struct, but can be a local variable used here and here alone.

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,135 @@
...
+	"github.com/aarzilli/nucular/label"
+)
+
+type FilterSelector struct {
+	wallet walletcore.Wallet
+
+	filters                  []string
+	filter                   string
+	txCount                  int
+	transactionCountByFilter []string
```
itswisdomagain: Why is a transaction COUNT slice a string slice instead of a number slice? Probably need to rename your variable.

itswisdomagain (changes requested):
Will review again after these.

nuklear/pagehandlers/history.go
Sirmorrison: Not exactly... Line 47 generates the initial receive address on the first user account. The call on line 57 only comes to effect only after the dropdown option is selected or changed.

nuklear/pagehandlers/history.go
Sirmorrison: `filter` this is just a variable stored here not the filters themselves.

itswisdomagain: I did not debate that with you. Please read my comment again and apply accordingly.

nuklear/pagehandlers/history.go
```diff
@@ -9,30 +9,35 @@ import (
...
-	fetchTxDetailsError error
+	wallet               	walletcore.Wallet
+	refreshWindowDisplay 	func()
+
+	filterSelectorWidget 	*widgets.FilterSelector
+	currentFilterText    	string
+	selectedTxFilter      	*txindex.ReadFilter
+
+	totalTxCount 			int
+	filterSelectorErr       error
```
itswisdomagain: - go fmt please
- move this `filterSelectorErr` property close to the `filterSelectorWidget` property. Doesn't that just make more sense?

nuklear/pagehandlers/history.go
```diff
@@ -9,30 +9,35 @@ import (
...
-	selectedTxHash      string
-	selectedTxDetails   *walletcore.Transaction
-	isFetchingTxDetails bool
-	fetchTxDetailsError error
+	wallet               	walletcore.Wallet
+	refreshWindowDisplay 	func()
+
+	filterSelectorWidget 	*widgets.FilterSelector
+	currentFilterText    	string
+	selectedTxFilter      	*txindex.ReadFilter
```
itswisdomagain: I've mentioned in a previous review that this does not need to be stored. Only `currentFilterText` needs to be stored.

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +46,44 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
+	// fetch initial table data
 	handler.totalTxCount, handler.fetchHistoryError = wallet.TransactionCount(nil)
 	if handler.fetchHistoryError == nil {
 		// only fetch txs if there was no error getting tx count.
-		go handler.fetchTransactions()
+		go handler.fetchTransactions(nil)
 	}
 
+	// fetch filtered data
+	handler.filterSelectorWidget, handler.filterSelectorErr = widgets.FilterSelectorWidget(wallet, func(){
```
itswisdomagain: This code in line 57 does not fetch filtered data. It sets up the filter widget. Update your comment.

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +46,44 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
+		_, selectedTxFilter, currentFilterText, err := handler.filterSelectorWidget.GetSelectedFilter()
```
itswisdomagain: `handler.filterSelectorWidget.GetSelectedFilter` should only return `currentFilterText, err` or just `currentFilterText` even.

nuklear/pagehandlers/history.go
```diff
@@ -77,10 +101,18 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
 
 func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
-		// show transactions first, if any
-		if len(handler.transactions) > 0 {
+		handler.filterSelectorWidget.Render(contentWindow)
```
itswisdomagain: `handler.filterSelectorWidget` may be nil if line 57 returns an error. You should check that it is not nil before rendering it here.

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +46,44 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
 		// only fetch txs if there was no error getting tx count.
-		go handler.fetchTransactions()
+		go handler.fetchTransactions(nil)
 	}
 
+	// fetch filtered data
+	handler.filterSelectorWidget, handler.filterSelectorErr = widgets.FilterSelectorWidget(wallet, func(){
+		_, selectedTxFilter, currentFilterText, err := handler.filterSelectorWidget.GetSelectedFilter()
+		if err != nil {
+			handler.filterSelectorErr = err
```
itswisdomagain: This is not an error with the filter selector. If this code here executes, then it means the filter has been successfully rendered on the page. This sub-function only gets triggered if a user selects an item from the filter dropdown. This is probably more related to `fetchHistoryError` since the next thing after a user selects an item from the filter is an attempt to load transactions for the selected filter item.

itswisdomagain (changes requested):
Looking better but still some changes. I also replied to some comments from the previous review. Please take a look.

nuklear/pagehandlers/history.go
Sirmorrison: https://github.com/raedahgroup/godcr/pull/384#discussion_r327695663

already resolves this as no error is returned.

nuklear/pagehandlers/history.go
```diff
@@ -9,30 +9,33 @@ import (
...
-	isFetchingTxDetails bool
-	fetchTxDetailsError error
+	wallet               	walletcore.Wallet
+	refreshWindowDisplay 	func()
+
+	filterSelectorWidget 	*widgets.FilterSelector
+	filterSelectorErr       error
+	currentFilterText    	string
+
+	totalTxCount 			int
```
itswisdomagain: go fmt please

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +44,42 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
 		// only fetch txs if there was no error getting tx count.
-		go handler.fetchTransactions()
+		go handler.fetchTransactions(nil)
 	}
 
+	// sets up the filter widget
+	handler.filterSelectorWidget, handler.filterSelectorErr = widgets.FilterSelectorWidget(wallet, func(){
+		currentFilterText, err := handler.filterSelectorWidget.GetSelectedFilter()
+		if err != nil {
+			handler.fetchHistoryError = err
```
itswisdomagain: Again you mark a comment as resolved without actually resolving it or commenting on it. Please go back to https://github.com/raedahgroup/godcr/pull/384/files#r327002260

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +44,42 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
+			return
+		}
+		if currentFilterText != handler.currentFilterText {
```
itswisdomagain: Empty line space before this.

nuklear/pagehandlers/history.go
```diff
@@ -41,25 +44,42 @@ func (handler *HistoryHandler) BeforeRender(wallet walletcore.Wallet, settings *
...
+			handler.currentFilterText = currentFilterText
+			handler.transactions = nil
+			go handler.fetchTransactions(walletcore.BuildTransactionFilter(handler.currentFilterText))
+		}
+
+		refreshWindowDisplay()
```
itswisdomagain: This should only be called after the transactions are fetched, i.e. from inside the `handler.fetchTransactions` method. Please remove from here.

If you need to refresh the display, that should be after line 63 where you clear the previous list of transactions.

nuklear/pagehandlers/history.go
```diff
@@ -77,10 +97,20 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
 
 func (handler *HistoryHandler) renderHistoryPage(window *nucular.Window) {
 	widgets.PageContentWindowDefaultPadding("History", window, func(contentWindow *widgets.Window) {
-		// show transactions first, if any
-		if len(handler.transactions) > 0 {
+		if handler.fetchHistoryError == nil {
+			handler.filterSelectorWidget.Render(contentWindow)
```
itswisdomagain, line 99: How does `handler.fetchHistoryError` relate to `handler.filterSelectorWidget`?

nuklear/pagehandlers/history.go
```diff
@@ -77,10 +97,20 @@ func (handler *HistoryHandler) Render(window *nucular.Window) {
...
+		}
+
+		if handler.filterSelectorErr != nil {
+			contentWindow.DisplayErrorMessage("Error with filter selector", handler.filterSelectorErr)
+		}
+
+		if len(handler.transactions) == 0 {
+			contentWindow.AddWrappedLabel("No Transactions to display yet", widgets.CenterAlign)
```
itswisdomagain: Transactions -> transactions

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+
+	for index, filter := range filterSelector.filters {
+		if filter == "All" {
+			txCount, txCountErr := wallet.TransactionCount(nil)
+			filterSelector.totalTxCount = txCount
+			filterSelector.txCountErr = txCountErr
+		}
+
+		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		filterSelector.txCountErr = txCountErr
```
itswisdomagain: This assignment is unncessary. Just return the error!

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+		selectionChanged: selectionChanged,
+	}
+
+	filterSelector.wallet = wallet
+	filterSelector.filters = walletcore.TransactionFilters
+	filterSelector.transactionFilter = make([]string, len(filterSelector.filters))
+
+	for index, filter := range filterSelector.filters {
+		if filter == "All" {
+			txCount, txCountErr := wallet.TransactionCount(nil)
```
itswisdomagain: If filter is all, you call `wallet.TransactionCount` twice. This is unnecessary. See the similar comment on your other PR #383.

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+			filterSelector.txCountErr = txCountErr
+		}
+
+		txCount, txCountErr := wallet.TransactionCount(walletcore.BuildTransactionFilter(filter))
+		filterSelector.txCountErr = txCountErr
+		if filterSelector.txCountErr != nil {
+			return nil, filterSelector.txCountErr
+		}
+
+		filterSelector.transactionFilter[index] = fmt.Sprintf("%s (%d)", filter, txCount)
```
itswisdomagain: You should ignore 0 count filters. This is done better in #383

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+	return filterSelector, nil
+}
+
+func (filterSelector *FilterSelector) Render(window *Window, addColumns ...int) {
+	filterSelectorWidth := defaultFilterSelectorWidth
+
+	// row with fixed column widths to hold account selection prompt, the account widget, and any other widgets that may be added later
+	rowColumns := make([]int, 2)
+	rowColumns[0] = window.LabelWidth("filter")
+	rowColumns[1] = filterSelectorWidth
```
itswisdomagain: lines 64 to 66 can be simplified to 
```
rowColumns := []int {
     window.LabelWidth("filter"),
     filterSelectorWidth,
}
```

By the way, those are not columns! Please use proper variable names!

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+
+	"github.com/raedahgroup/dcrlibwallet/txindex"
+	"github.com/raedahgroup/godcr/app/walletcore"
+	"github.com/aarzilli/nucular/label"
+)
+
+type FilterSelector struct {
+	wallet walletcore.Wallet
+	filters                  []string
+	transactionFilter []string
```
itswisdomagain: - `transactionFilters`, this is a collection of items, not a single item
- what's the difference between `filters` and `transactionFilters`?
- go fmt

itswisdomagain (changes requested):

nuklear/widgets/dropDown.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	return filterSelector, nil
+}
+
+func (filterSelector *FilterSelector) Render(window *Window, addColumns ...int) {
+	filterSelectorWidth := defaultFilterSelectorWidth
+
+	// row with fixed column widths to hold account selection prompt, the account widget, and any other widgets that may be added later
+	rowColumns := []int {
```
Sirmorrison: here

itswisdomagain (approved):

PR #394, opened 2019-08-07
fyne: implement staking page, solo/vsp ticket purchase

itswisdomagain:
Implement ticket purchase feature on staking page.
Purchase tickets by completing the purchase ticket(s) form. VSP host is optional; if provided the tickets purchased after the form is submitted will share voting rights with the VSP whose host url was provided. A test vsp host is http://13.78.188.141:8000/

Screenshot 2019-08-07 at 7 22 53 PM
Screenshot 2019-08-07 at 7 23 02 PM


fyne/pages/staking.go
```diff
@@ -0,0 +1,143 @@
...
+		stakeInfoText = fmt.Sprintf("Error loading stake info: %s", err.Error())
+	} else {
+		stakeInfoText = fmt.Sprintf("unmined %d   immature %d   live %d   voted %d   missed %d   expired %d \n"+
+			"revoked %d   unspent %d   allmempooltix %d   poolsize %d   total subsidy %s",
+			stakeInfo.OwnMempoolTix, stakeInfo.Immature, stakeInfo.Live, stakeInfo.Voted, stakeInfo.Missed, stakeInfo.Expired,
+			stakeInfo.Revoked, stakeInfo.Unspent, stakeInfo.AllMempoolTix, stakeInfo.PoolSize, stakeInfo.TotalSubsidy)
+	}
+
+	if staking.stakeInfoLabel.Text == "" {
+		staking.stakeInfoLabel.Text = stakeInfoText
```
metaclips: If you aren't using SetText you should call widget refresh, also I dont think the crash is because the string is empty, so no need for a conditional statement.

itswisdomagain: If I use `SetText`, the program crashes because the label has not been added to the UI at this point.
If the text in the label is empty, then it has not been added to the page. Setting the `Text` property directly works without requiring a refresh.

metaclips: Yes, this has been fixed, please check https://github.com/fyne-io/fyne/issues/381

metaclips: add a layout.Spacer() before and after the submit button in a NewHBox, that should center it.

metaclips (approved):
Tested locally and runs fine.

PR #397, opened 2019-08-21
cli: implement wallet restore from seed

itswisdomagain:
Screenshot 2019-08-21 at 4 07 43 PM

### DIFF
--- a/cli/walletloader/createwallet.go
+++ b/cli/walletloader/createwallet.go
@@ -1,15 +1,18 @@
 package walletloader
 
 import (
+	"bufio"
 	"context"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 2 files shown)


Sirmorrison:
im not sure if this is normal but after inputting the wallet seed, i had to hit tab key twice before the prompt for a password showed up.

Sirmorrison (commented):
PR looks good to go but for the part where i had to hit tab key twice before the prompt to enter passphrase or and invalid seed notification prompt pops up.

itswisdomagain:
That behaviour is as expected @Sirmorrison, compare the screenshot on the PR description.

Sirmorrison (approved):
Good to go.

PR #400, opened 2019-08-22
Account page (with fyne library)

barisere:
This pull request implements the accounts page using the `fyne` library.
The page displays the list of accounts available, together with the accounts' balances.
Beside each account name, a toggle button is displayed. Clicking it will hide/show more information about the account.

The target functionality is that available in the mobile app, as seen in the following screenshots. Controls for changing account settings (i.e., setting an account as the default account, or hiding an account from the UI) will be done in another PR.

Mobile screenshot with account details collapsed

Mobile screenshot with account details expanded

The current functionality is as shown in the screenshot below.

Desktop screenshot with the fyne library

### DIFF
--- a/fyne/pages/accounts.go
+++ b/fyne/pages/accounts.go
@@ -1,0 +1,29 @@
+// Package pages contains implementations of the various pages that
+// can be viewed on the desktop GUI when using the "fyne" library.
+//
+// This file contains the implementation of the accounts overview page.
+package pages
+
+import (
+	"fmt"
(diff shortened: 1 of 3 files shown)


metaclips:
Screenshot from 2019-08-26 18-27-23
This are the issues noticed on your PR.

- Collapsible doesn't resize
-  There should be a checkbox to select default account which isn't implemented.

Account page is already being worked on by a developer.

PR #403, opened 2019-08-23
implement tickets list page

itswisdomagain:
Just a simple way to view hashes for tickets controlled by a wallet, no UI guide provided/worked with. Perhaps the **History** page could serve this purpose if tx filter is implemented.

Screenshot 2019-08-23 at 11 41 21 PM


itswisdomagain:
Redone in #408 with additional features.

PR #405, opened 2019-08-28
[WIP] Account and History page.

metaclips:
Implemented account page, history page and transaction details page. Receive page automatically generate a new address when an account is selected, icons are now hardcoded.
image
image
image
image

Issue #239, opened 2019-03-24
Go Install Breaks Modes


matthewjamesr:
Running after `go install` from anywhere breaks the following modes:

* http:
  ```
  C:\Users\me>godcr --mode http
  2019/03/24 19:52:45 error loading templates: open web/views/error.html: The system cannot find the path specified.
  ```
  **Fix:** Utilize GOPATH to point to resources
* nuklear: godcr.exe simply exits (panic?)
  Not sure of the cause

itswisdomagain:
The nuklear issue is same with the http issue - they both require access to some files at runtime (nuklear tries to load font files).

Using $GOPATH is a temporary fix, it implies that the user has to have the required files in $GOPATH - won't work when we start distributing the binary as a standalone file.

Long term fix is to find a means to bundle the needed static files with the binary when building. gobind came to mind before but we haven't looked more into it because there are other issues with higher priority.

We should keep this issue open though so we can document thoughts on a good fix going forward.

PR #419, opened 2019-09-04
fyne package set up to use dcrlibwallet/master

itswisdomagain:
The goal of this PR is to simplify the fyne interface and make it use dcrlibwallet/master while maintaining the codebase for the other interfaces and maintaining the ability to build/install/run each interface separately.

The ultimate goal (will be accomplished over a series of PRs) is to move all shared codes to dcrlibwallet so that the different interfaces rely solely on dcrlibwallet for shared code. This would mean move MOST of the code currently in the `./app` package over to dcrlibwallet/master. For example the `./app/config` package will need to be moved to dcrlibwallet such that saving and reading config values (aka settings) will be done using shared code in dcrlibwallet.

The fyne interface will lead this new direction and this PR is the first work in this direction. The `./fyne` package now no longer relies on the `./app` package and operations relating to logging and config reading are now done using dcrlibwallet.

The codebase in this PR may be a little messy (especially for the non-fyne interfaces). In line with the goal mentioned above, code-review attention should be mostly focused on the fyne interface while also testing to ensure that the other interfaces build and run as before.

The steps taken in this PR:
- make `./app`, `./cmd` and each interface package into (sub) modules, this is primarily to allow each interface define different sets of dependencies.
- set dcrlibwallet mod version in `./fyne/go.mod` to `dcrlibwallet/master` while the other interfaces continue to use `dcrlibwallet/wip`
- remove fyne features that require `dcrlibwallet/wip` which primarily entails not using the `./app` package from fyne since that package relies heavily on `dcrlibwallet/wip`
- re-implement wallet loading and fyne app launch features using `dcrlibwallet/master`
- add navigation menu and dummy page contents per menu item

It is (temporarily) no longer possible with this PR to build/install or run all interface binaries from the project root. This is because of the clashing/different dcrlibwallet versions used across the different interfaces.

Thus, in keeping with the desire to focus on the fyne interface for now, only the fyne interface is configured to be built/run or installed from the project root:
```
go run ./cmd/godcr-fyne
go build ./cmd/godcr-fyne
go install ./cmd/godcr-fyne
```
Also, select changes from https://github.com/raedahgroup/godcr/pull/405 are introduced into this PR to enable running fyne successfully regardless of where the fyne binary is placed.

To build/install or run other interfaces:
```
cd ./cmd/{godcr-interface}
go run .
go build
go install
```
Usually best to avoid go install for now since go install places the generated binaries in $GOPATH and as per https://github.com/raedahgroup/godcr/issues/239, the http and nuklear interfaces will not run from any directory other than `{project root}/cmd/godcr-web` and  `{project root}/cmd/godcr-nuklear` respectively.

Subsequent PRs will gradually complete the migration from dcrlibwallet-wip to dcrlibwallet/master for other interfaces (one interface at a time); after which such interface can be run/built or installed from the project root. Also, once all the interfaces have their codes modified to become fully standalone and compatible with dcrlibwallet/master, further cleanup of the sub modules and the different `go.mod` files will be done.


barisere (changes requested):
File `cmd/go.mod` should be removed, because there's no package providing that module.
```
can't load package: package github.com/raedahgroup/godcr/cmd: unknown import path "github.com/raedahgroup/godcr/cmd": cannot find module providing package github.com/raedahgroup/godcr/cmd
```
Also, the module replace directive at the root of the project causes the build to fail there, because there's no module in cmd/godcr-fyne.
```
can't load package: package github.com/raedahgroup/godcr/cmd: unknown import path "github.com/raedahgroup/godcr/cmd": cannot find module providing package github.com/raedahgroup/godcr/cmd
```

Sirmorrison (approved):
Tested all interfaces and PR looks good to go.

metaclips:
Looks good.

barisere (approved):

PR #424, opened 2019-09-04
Pack fyne icons into godcr-fyne binary

### LINKED ISSUE #423: Pack static assets into release binaries for fyne UI
The fyne UI depends on some PNG images that are used as icons on the UI. Previously they were read from a relative path on disk. That worked in development, but the images cannot be read when the binary is run from a different location on the filesystem.
A temporary workaround was introduced by @metaclips: the images were converted into bytes and hard-coded in the application. That workaround made its way into #419. 
However, hard-coding images as byte slices obscures the fact that they are images, and we cannot visually inspect the images on disk. We intend to restore the image files and, instead of hard-coding in the source code, pack them into the built binary using [Packr](https://github.com/gobuffalo/packr).
This issue will track the implementation of this approach across each interface where we adopt it.

barisere:
This PR packs the images needed by the fyne interface into the release binary. To do this we currently have to first run `packr2` in the `fyne` directory, then run `go build` in the `cmd/godcr-fyne` directory.
[Packr (v2)](https://github.com/gobuffalo/packr/tree/master/v2) is needed in the build step. Follow the steps below to build or install godcr-fyne.
```sh
# Install Packr2 binary to GOPATH. Run the command with module mode turned off.
GO111MODULE=off go get -u github.com/gobuffalo/packr/v2/packr2

# Generate Go files to pack fyne icons into the binary as byte slices.
(cd fyne && packr2)

# Then build godcr-fyne.
go build ./cmd/godcr-fyne

# or install into $GOBIN
go install ./cmd/godcr-fyne
```
At this time `godcr-fyne` binary can be built from only its own module directory. Building from any other directory will fail, because `godcr-fyne` currently depends on an untagged version of dcrlibwallet that is newer than that used by other packages. Therefore, it will always override the other versions in the build list if built together with them.
This may be resolved when all dependencies on dcrlibwallet are successfully switched to dcrlibwallet/master.
Fixes #423.


fyne/go.mod
```diff
@@ -1,12 +1,13 @@
...
 require (
 	fyne.io/fyne v1.1.0
 	github.com/decred/slog v1.0.0
+	github.com/gobuffalo/packr v1.30.1 // indirect
+	github.com/gobuffalo/packr/v2 v2.6.0
 	github.com/raedahgroup/dcrlibwallet v1.1.1-0.20190904121310-90259892e503
-	github.com/raedahgroup/godcr/fyne v0.0.0-00010101000000-000000000000
 )
 
 replace github.com/raedahgroup/godcr/fyne => ./
```
itswisdomagain: This replace directory would likely no longer be needed. I was quite surprised when I noticed that without it, fyne wouldn't build. Now I realize it was because of the incorrect module name at the top of this file.

fyne/icons.go
```diff
@@ -1,35 +1,89 @@
 package fyne
 
 import (
+	"fmt"
+	"path/filepath"
+	"sync"
+
 	"fyne.io/fyne"
+	"github.com/gobuffalo/packr/v2"
```
itswisdomagain: Separate with a space. Likely github imports first, space then fyne imports (g beore f alphabetically)

fyne/display.go
```diff
@@ -35,13 +35,17 @@ func (app *fyneApp) displayMainWindow() {
 }
 
 func (app *fyneApp) setupNavigationMenu() {
+	icons, err := getIcons(overviewIcon, historyIcon, sendIcon, receiveIcon, accountsIcon, stakeIcon)
+	if err != nil {
+		panic(err)
```
itswisdomagain: Should not panic. Use the `app. displayLaunchErrorAndExit` instead.

fyne/icons.go
```diff
@@ -1,35 +1,89 @@
...
+		iconBytes, err := imageBox.Find(name)
+		if err != nil {
+			return nil, err
+		}
+		icons[name] = &fyne.StaticResource{StaticName: name, StaticContent: iconBytes}
+	}
+	return icons, nil
+}
+
+// var overviewIcon = &fyne.StaticResource{
```
itswisdomagain: These can be deleted.

itswisdomagain (changes requested):
Need to include in your PR description and in readme that the `packr` cli utility needs to be installed using `go get -u github.com/gobuffalo/packr/v2/packr2`.

Also need to add `fyne/fyne-packr.go` and `fyne/packrd/` to gitignore.

Please while you're here, delete this dead code that made it's way into master.

itswisdomagain (changes requested):
Running `./godcr-fyne` from `./cmd/godcr-fyne` after building (`cd ./cmd/godcr-fyne && go build && ./godcr-fyne`) gives this error: 
`panic: stat /Users/itswisdomagain/go/src/github.com/raedahgroup/godcr/fyne/overview.png: no such file or directory`

When I run `./godcr-fyne` from the project root after building (`go build ./cmd/godcr-fyne && ./godcr-fyne`), I get:
`panic: stat /Users/itswisdomagain/go/src/github.com/raedahgroup/godcr/overview.png: no such file or directory`

fyne/icons.go
barisere: Generally, standard library >> GitHub >> custom domains?

ademuanthony:
I am unable to build the project
```
go: github.com/raedahgroup/dcrlibwallet@v1.1.1-0.20190904121310-90259892e503: unknown revision 90259892e503
go: error loading module requirements
```

barisere:
@ademuanthony the revision 90259892e503 was somehow erased from the dcrlibwallet history. I think someone force-pushed. I've replaced it with a compatible revision, at least the interfaces look the same and they're from close dates.
Try rebuilding, you may need to delete your local sources and pull this PR afresh.

itswisdomagain:
This PR should be set to fix/resolve #423 rather than be a partial solution. See https://github.com/raedahgroup/godcr/issues/423#issuecomment-530544179

README.md
```diff
@@ -43,8 +43,15 @@ git clone https://github.com/raedahgroup/godcr %GOPATH%/src/github.com/raedahgro
 * If you cloned to $GOPATH, set the `GO111MODULE=on` environment variable before building.
 Run `export GO111MODULE=on` in terminal (for Mac/Linux) or `setx GO111MODULE on` in command prompt for Windows.
 * `cd` to the cloned project directory.
-* To build/install godcr-fyne:
-`go build ./cmd/godcr-fyne` or `go install ./cmd/godcr-fyne`.
+* To build/install godcr-fyne, you need to install Packr also:
```
itswisdomagain: `you need to install Packr first`

README.md
```diff
@@ -43,8 +43,15 @@ git clone https://github.com/raedahgroup/godcr %GOPATH%/src/github.com/raedahgro
...
+```sh
+# Install Packr2 binary
+go get -u github.com/gobuffalo/packr/v2/packr2
+# Then build godcr-fyne
+cd cmd/godcr-fyne && go build .
```
itswisdomagain: It's also possible to build/install without changing the cwd using `go build ./cmd/godcr-fyne` or `go install ./cmd/godcr-fyne`.

Also no mention is made of using packr to pack the icons/assets before building.

And wouldn't running `go get -u github.com/gobuffalo/packr/v2/packr2` on the project directory give unintended results?

Makefile
```diff
@@ -0,0 +1,18 @@
+.POSIX:
```
itswisdomagain: Thought we were going with using a shell script?

itswisdomagain (changes requested):
I see changes to some go.mods and go.sums. The resulting effect is that it is no longer possible to build other interfaces aside fyne.

Also please include the new build steps in your PR description so we don't need to go searching for how to test the features introduced in your PR.

itswisdomagain (changes requested):
`fyne/fyne-packr.go` and `fyne/packrd/packed-packr.go` are generated when `packr2` is run. Those should be gitignored.

fyne/display.go
```diff
@@ -35,13 +36,17 @@ func (app *fyneApp) displayMainWindow() {
 }
 
 func (app *fyneApp) setupNavigationMenu() {
+	icons, err := getIcons(overviewIcon, historyIcon, sendIcon, receiveIcon, accountsIcon, stakeIcon)
+	if err != nil {
+		app.displayLaunchErrorAndExit(fmt.Sprintf("An error occured while load app icons: %s", err))
```
itswisdomagain: while **loading** app icons

itswisdomagain (changes requested):
There are a couple of unused icons, let's take them out and only (re) introduce them when they will be needed in order to avoid forgetfully leaving stuff in the branch that we never use.

README.md
```diff
@@ -43,8 +43,20 @@ git clone https://github.com/raedahgroup/godcr %GOPATH%/src/github.com/raedahgro
...
+# Install Packr2 binary to GOPATH. Run the command with module mode turned off.
+GO111MODULE=off go get -u github.com/gobuffalo/packr/v2/packr2
+```
+Then run the build script `build.sh` as follows.
+```sh
+./build.sh fyne
+```
+Or install `godcr-fyne` into $GOBIN as follows.
+```sh
+(cd cmd/godcr-fyne && go install)
```
itswisdomagain: This would fail without running `packr2` first.
Also, should be able to run `go install ./cmd/godcr-fyne` from the root dir.

fyne/icons.go
```diff
@@ -1,35 +1,41 @@
...
+const (
+	aboutIcon       = "about.png"
+	accountsIcon    = "account.png"
+	decredDarkIcon  = "decredDark.png"
```
itswisdomagain: These and others like it that are not used should be removed along with the png files.

itswisdomagain (changes requested):

README.md
barisere: For `godcr-fyne`, this is temporarily not possible, until the dependencies on dcrlibwallet are synced. For the other binaries it is possible.

itswisdomagain (changes requested):
https://github.com/raedahgroup/godcr/pull/424/commits/1888dbac8b5a346a4b79475e77d1cd9d7b2d1575 cleared the go mod changes that do not relate to this PR. This PR now builds as documented in the project's readme. Any further changes to the modules setup should be done in a different PR.

For now, just the readme and build.sh files need some corrections to the build instructions and this should be good to merge.

itswisdomagain (approved):

PR #439, opened 2019-11-05
Fyne History Page Implementation

Sirmorrison:
Fixes #439 

image

image

image

image

image

image

image


oshorefueled:
The history page seems to work when using one wallet. However, it's not fully functional with multiple wallets. First, how multiple wallets are displayed. A dropdown would be more suitable for displaying users' wallets. You can use the "receive" page as a reference for that. 

Screenshot 2019-12-15 at 1 55 53 AM

I get an empty transactions result when I try switching between wallets and that's just one of the issues. Kindly create multiple wallets and test.

Screenshot 2019-12-15 at 2 09 57 AM

terminal/pages/history.go
```diff
@@ -20,7 +20,7 @@ const (
 	MessageKindInfo
 )
 
-const txPerPage int32 = 25
+const txPerPage int32 = 10
```
oshorefueled: You're working on the history page for fyne. You shouldn't make changes to other files that don't affect the feature you're adding. If there's a need to clean up code or make modifications to unrelated parts of the code base, that should be done on another pull request.  Please fix other parts of this PR that this applies to.

cli/go.sum
```diff
@@ -14,6 +14,7 @@ github.com/agl/ed25519 v0.0.0-20170116200512-5312a6153412/go.mod h1:WPjqKcmVOxf0
 github.com/akavel/rsrc v0.8.0/go.mod h1:uLoCtb9J+EyAqh+26kdrTgmzRBFPGOolLWKpdxkKq+c=
 github.com/asdine/storm v0.0.0-20190216191021-fe89819f6282 h1:DmSVc81daQAPvXwcCZi0W6A14sTCYQ1QI21C0E37KoY=
 github.com/asdine/storm v0.0.0-20190216191021-fe89819f6282/go.mod h1:cMLKpjHSP4q0P133fV15ojQgwWWB2IMv+hrFsmBF/wI=
+github.com/atotto/clipboard v0.1.2 h1:YZCtFu5Ie8qX2VmVTBnrqLSiU9XOWwqNRmdT3gIQzbY=
```
oshorefueled: unrelated file to the pull request, this file shouldn't be part of the PR. When working on a feature, you should make modifications to only areas that affect or are related to what you're working on.

fyne/helpers/amount.go
```diff
@@ -0,0 +1,59 @@
+package helpers
```
oshorefueled: none of this code seems to be used anywhere, was it previously used somewhere?

oshorefueled (changes requested):

terminal/pages/history.go
Sirmorrison: Please you may need to be more explicit as to the pages.

fyne/helpers/amount.go
Sirmorrison: it came in after rebase. will remove it for now.

oshorefueled:
Please remove this file https://github.com/raedahgroup/godcr/blob/28a531acf89b55f9ab87a2e920240d443c6cef9b/cmd/godcr-fyne/.~bash-profile.swp from your PR

cli/go.sum
oshorefueled: > unrelated file to the pull request, this file shouldn't be part of the PR. When working on a feature, you should make modifications to only areas that affect or are related to what you're working on.

This comment also applies to the files where you applied "go fmt" like the transactionList code. You also made modifications to the receives page code. If a code file doesn't relate to the feature you're working on, try not to make modifications to it. If you feel any area needs modification or improvement, you can bring up an issue and another PR can be created for that. Merge conflicts would be avoided this way. Kindly remove unrelated files to this PR.

fyne/pages/history.go
```diff
@@ -1,10 +1,681 @@
...
+
+func updateTable(multiWallet *dcrlibwallet.MultiWallet, window fyne.Window, tabMenu *widget.TabContainer) {
+	size := txHistory.txTable.Container.Content.Size().Height - txHistory.txTable.Container.Size().Height
+	scrollPosition := float64(txHistory.txTable.Container.Offset.Y) / float64(size)
+	txTableRowCount := txHistory.txTable.NumberOfColumns()
+
+	if txHistory.allTxCount > int(txHistory.TotalTxFetched) {
+		if txHistory.txTable.Container.Offset.Y == 0 {
+			if txTableRowCount < int(txPerPage) {
+				return
```
metaclips: If I am right this checks if the scroller is at the starting point and then if the amount of transactions in the table is less than the min tx per scroll which is 25 it returns, this shouldn't be because if it's a new wallet and there is an incoming transactions the table won't be updated.

fyne/pages/history.go
```diff
@@ -1,10 +1,681 @@
...
+			}
+			// table not yet scrolled wait 4 secs and update
+			time.AfterFunc(time.Second*4, func() {
+				updateTable(multiWallet, window, tabMenu)
+			})
+		} else if scrollPosition < 0.5 {
+			if txHistory.TotalTxFetched == txPerPage {
+				time.AfterFunc(time.Second*4, func() {
```
metaclips: You shouldn't update every 4 seconds, on a fast scroll <0.5 parameter might not be met, also I really don't see why we are calling a time.AfterFunc recursively, for every time users switch to history page a new time.AfterFunc call is made this can better be passed here https://github.com/raedahgroup/godcr/blob/0dc5f3cac669f276f1244f061976bdcf17212380/fyne/pages/display.go#L88 making sure to return function if history page is not in view

fyne/pages/history.go
```diff
@@ -1,10 +1,681 @@
...
+
+		return widgets.NewClickableBox(widget.NewHBox(textToCopy),
+			func() {
+				messageLabel.SetText("Data Copied")
+				clipboard := window.Clipboard()
+				clipboard.SetContent(text)
+
+				time.AfterFunc(time.Second*2, func() {
+					if tabMenu.CurrentTabIndex() == 1 {
+						messageLabel.SetText("")
```
metaclips: This should be messageLabel.Hide()

metaclips (changes requested):
You should try to split components to separate functions and place them in the handler package for easy review also scrolling seems very slow, can we reduce transaction table per scroll to 15 instead of 25?

Sirmorrison:
> You should try to split components to separate functions and place them in the handler package for easy review also scrolling seems very slow, can we reduce transaction table per scroll to 15 instead of 25?

As a matter of uniformity, across all godcr interfaces, the txperpage is 25.

fyne/pages/history.go
Sirmorrison: The idea was not to hide the message label. Reason is it distorts the page when ever the message displays.

metaclips: While splitting components to handler package, place a variable of *widget.VBox in a struct that should house all component on history page then when you make changes, only the Box should be refreshed, this was done in send page and fixes issues with distortion.

metaclips:
> > You should try to split components to separate functions and place them in the handler package for easy review also scrolling seems very slow, can we reduce transaction table per scroll to 15 instead of 25?
> 
> As a matter of uniformity, across all godcr interfaces, the txperpage is 25.

Scrolling is slow and table contents distorts while scrolling we should reduce table contents.

oshorefueled:
Is it necessary that we use a horizontal scroll?
Screenshot 2019-12-24 at 9 45 56 AM

cmd/go.mod
```diff
@@ -6,10 +6,10 @@ require (
 	github.com/decred/slog v1.0.0
 	github.com/jessevdk/go-flags v1.4.0
 	github.com/jrick/logrotate v1.0.0
+	github.com/raedahgroup/dcrlibwallet v1.1.1-0.20190928085114-bcc6e6b7769a // indirect
```
oshorefueled: why do we need dcrlibwallet as a dependency here?

metaclips:
> Is it necessary that we use a horizontal scroll?
> Screenshot 2019-12-24 at 9 45 56 AM

There are contents we can do away with.
I don't see why the texts are needed with the icons in Direction line, type can also be removed and also hash text can be reduced so that the width wont be jumpy when users switch pages.

Sirmorrison:
> > Is it necessary that we use a horizontal scroll?
> > Screenshot 2019-12-24 at 9 45 56 AM
> 
> There are contents we can do away with.
> I don't see why the texts are needed with the icons in Direction line, type can also be removed and also hash text can be reduced so that the width wont be jumpy when users switch pages.

Is the recommendation that i remove the icon?
About the jumpy pages why dont we give the fyne interface a fixed min width?

Sirmorrison:
> Is it necessary that we use a horizontal scroll?
> Screenshot 2019-12-24 at 9 45 56 AM

resolved

oshorefueled:
The transaction details popup seems to be losing part of its content to the edges of the window.

Screenshot 2019-12-27 at 10 31 14 AM

cmd/go.mod
```diff
@@ -9,7 +9,6 @@ require (
 	github.com/raedahgroup/godcr/app v0.0.0-00010101000000-000000000000
 	github.com/raedahgroup/godcr/cli v0.0.0-00010101000000-000000000000
 	github.com/raedahgroup/godcr/nuklear v0.0.0-00010101000000-000000000000
-	github.com/raedahgroup/godcr/terminal v0.0.0-00010101000000-000000000000 // indirect
```
oshorefueled: Why was this line deleted?

fyne/widgets/table.go
```diff
@@ -11,67 +12,157 @@ type Table struct {
...
+	table.Container = widget.NewScrollContainer(table.Result)
 	table.tableData = []*widget.Box{heading}
 	table.tableData = append(table.tableData, data...)
-	table.Result = widget.NewHBox()
-	table.Refresh()
+	table.set()
 }
 
+// Append widget adds to the bottom row of a table.
 func (table *Table) Append(data ...*widget.Box) {
```
oshorefueled, line 26: You changed quite a number of code in the table.go. It's a widget I used on the overview page. It seemed to work well, any reason for the change? Would my existing implementation work without an issue if I used this?

Sirmorrison: I had this these codes already before the overview page was implemented. Anyways have you tested your implementation on the overview page just to be sure there are no issues because overview page seems fine to me.?

Sirmorrison:
> The transaction details popup seems to be losing part of its content to the edges of the window.
> 
> Screenshot 2019-12-27 at 10 31 14 AM

so basically the issue is the change @metaclips suggested that we reduce the width of the history page. this also affected the txdetails popup as its wider than the current history page.
https://github.com/raedahgroup/godcr/pull/439#issuecomment-568803534

Two things can done here. 
1) increase the width of the default history page to accommodate the details page.
2) reduce the width of the details popup and add a horizontal scroll to view the hidden details to the right. 

for now i will make it scroll horizontally.
image

metaclips (changes requested):
scrolling seems to be much better, still needs little changes to UI.
- Status table contents should be centered.
Screenshot 2019-12-28 at 20 04 13
-  On transaction details page, confirmations-Data contents are not aligned, can we place them in a form widget?
- All tables on transaction details page should be properly aligned
Screenshot 2019-12-28 at 20 08 13
- Info icon should be right aligned, and popup width should be reduced.

fyne/pages/handler/historypagehandler/txtablecomponents.go
```diff
@@ -0,0 +1,138 @@
...
+			widget.NewLabelWithStyle(dcrutil.Amount(tx.Amount).String(), fyne.TextAlignTrailing, fyne.TextStyle{}),
+			widget.NewLabelWithStyle(dcrutil.Amount(tx.Fee).String(), fyne.TextAlignCenter, fyne.TextStyle{}),
+			widgets.NewClickableBox(widget.NewHBox(widget.NewLabelWithStyle(trimmedHash, fyne.TextAlignCenter, fyne.TextStyle{Italic: true})), func() {
+				historyPage.fetchTxDetails(txForTrimmedHash)
+			}),
+		))
+	}
+
+	if prepend {
+		txTable.Prepend(txBox...)
```
metaclips: You should try as much as possible not to use Prepend and Append function especially when users change to history page, can we instead call A NewTable instead for faster rendering , if users view history page?

fyne/pages/handler/historypagehandler/filterselector.go
```diff
@@ -0,0 +1,133 @@
...
+		"Staking":     dcrlibwallet.TxFilterStaking,
+	}
+
+	if walletId != historyPage.selectedWalletID {
+		historyPage.selectedWalletID = walletId
+	}
+
+	txCountForFilter, err := historyPage.MultiWallet.WalletWithID(historyPage.selectedWalletID).CountTransactions(allTxFilters["All"])
+	if err != nil {
+		historyPage.errorMessage = fmt.Sprintf("Cannot load history page page. Error getting transaction count for filter All: %s", err.Error())
```
metaclips, line 34: why are we passing error messages here? isn't there already a widget that handles that? Also, if err!=nil parameter is met, the app crashes, all component functions that require using dcrlibwallet should have a return parameter "error", you should stop all history page initialisation if any function throws an error.

metaclips (changes requested):
You should add an init prefix to component on history page and make the component that is to be initialised the first calling function and other helper functions below on each files for easy review and so that others wont have issues implementing features on the page.

fyne/pages/handler/historypagehandler/filterselector.go
Sirmorrison: this is not true if there is an error at any point the error is returned the app does not panic... you should do well to test this properly as i have done so from my end...

if you are changing `!=` to `==` you should comment out `err.Error()` else it will panic as there is no error to be returned. 

image

fyne/pages/handler/historypagehandler/txtablecomponents.go
Sirmorrison: This will function more like pagination reason is for every update a new table will be called and this will have only data from the last tx fetched.

fyne/pages/handler/historypagehandler/filterselector.go
metaclips: Screenshot 2019-12-30 at 19 58 06

This is what is shown when I make `err == nil` [here](https://github.com/Sirmorrison/godcr/blob/08c678bbe25a6b7609f1372848bae39567b60472/fyne/pages/handler/historypagehandler/txtablecomponents.go#L43), you are showing parts of history page components even when there was an error.
You shouldn't be passing errors to variables and then making a [return](https://github.com/Sirmorrison/godcr/blob/08c678bbe25a6b7609f1372848bae39567b60472/fyne/pages/handler/historypagehandler/filterselector.go#L34)

You are still calling up functions even if there was error on other functions called recently and then checking if any of the functions returned an error [here](https://github.com/Sirmorrison/godcr/blob/08c678bbe25a6b7609f1372848bae39567b60472/fyne/pages/handler/historypagehandler/inithistorypage.go#L81)

This shouldn't be as any of the functions that errors and returns early without initialising struct variable "HistoryPageData struct" could cause a panic in other functions if they use the uninitialised variable, you should instead pass return parameter "errors" to all functions that uses dcrlibwallet also InitHistoryPage function and then if any of the dcrlibwallet library functions return an error,
- Stop the failed function, return the error
- Stop history page initialisation in the handler package, return the error passed to InitHistoryPage
- Check if InitHistoryPage function which is to be called in history.go file returns an `err` if there was an error, only show on history page a label indicating that there was an error initialising history page.

PR #468, opened 2019-12-25
implement exit page

### LINKED ISSUE #467: Implement exit page to godcr-fyne
Fyne interface should have a button on the tab menu that properly exits the wallet.

metaclips:
Screenshot 2019-12-25 at 12 12 47
closes #467


oshorefueled:
The app successfully exits when I click "Yes" on the exit popup. However, when I click "No" on a page other than the overview page, the popup disappears but takes me to the overview page. The app should remain on the same page after clicking the "No" button.

fyne/pages/exit.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/raedahgroup/godcr/fyne/pages/handler/values"
+	"github.com/raedahgroup/godcr/fyne/widgets"
+)
+
+func (app *AppInterface) exitPageContent() fyne.CanvasObject {
+	var popup *widget.PopUp
+
+	yesButton := widget.NewButtonWithIcon("Yes", theme.ConfirmIcon(), func() {
+		app.Window.Close()
+		fmt.Println("Exited fyne")
```
oshorefueled: This should be logged

oshorefueled (changes requested):

metaclips:
> The app successfully exits when I click "Yes" on the exit popup. However, when I click "No" on a page other than the overview page, the popup disappears but takes me to the overview page. The app should remain on the same page after clicking the "No" button.

What would be shown on exit page if the popup closes?  Showing an empty page on clicking No doesn't really seem right to be honest.

oshorefueled:
> > The app successfully exits when I click "Yes" on the exit popup. However, when I click "No" on a page other than the overview page, the popup disappears but takes me to the overview page. The app should remain on the same page after clicking the "No" button.
> 
> What would be shown on exit page if the popup closes? Showing an empty page on clicking No doesn't really seem right to be honest.

ideally, the popup should not be a page, but just a popup capable of being visible on any page. But in a case where this isn't possible as a result of fyne's limitations, you can just go back to the previous page. It should give the same effect as if the user never left the page.

oshorefueled (approved):

PR #469, opened 2019-12-26
Add live data to wallet sync details

### LINKED ISSUE #447: add live block sync status
replace dummy data for block sync status with users's live data. Start with "block syncing" status and sync details. "Synced" and "Not Synced" status can then be added after.

oshorefueled:
This PR closes #447. It adds live wallet sync details to the overview page.

Screenshot 2019-12-26 at 10 28 45 AM

Screenshot 2019-12-26 at 10 27 36 AM

### DIFF
--- a/fyne/pages/display.go
+++ b/fyne/pages/display.go
@@ -10,6 +10,7 @@ import (
 	"fyne.io/fyne/theme"
 	"fyne.io/fyne/widget"
 
+
 	"github.com/decred/slog"
 	"github.com/raedahgroup/dcrlibwallet"
 
@@ -56,6 +57,7 @@ func (app *AppInterface) DisplayMainWindow() {
 	app.Window.SetFixedSize(true)
 	app.Window.CenterOnScreen()
 	fyne.CurrentApp().Settings().SetTheme(theme.LightTheme())
+	go overviewHandler.PreserveSyncSteps()
 	app.Window.ShowAndRun()
 	app.tearDown()
 }
--- a/fyne/pages/handler/overview.go
+++ b/fyne/pages/handler/overview.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"fyne.io/fyne/layout"
 	"github.com/gen2brain/beeep"
+	"github.com/raedahgroup/godcr/fyne/pages/handler/values"
 	"image/color"
 	"log"
 	"sort"
@@ -212,10 +213,19 @@ func (handler *OverviewHandler) blockSyncStatus() (string, color.Color) {
 }
 
 func (handler *OverviewHandler) SyncTrigger(wallet *dcrlibwallet.MultiWallet) {
-	if wallet.IsSyncing() {
+	var notifyTitle, notifyMessage string
+	synced, syncing := wallet.IsSynced(), wallet.IsSyncing()
+	if wallet.ConnectedPeers() > 0 && (synced || syncing) {
+		if synced {
+			notifyTitle = "Wallet Disconnected"
+			notifyMessage = "wallet has been disconnected from peers"
+		} else if syncing {
+			notifyTitle = "Sync Canceled"
+			notifyMessage = "wallet sync stopped until wallet reconnect"
+		}
 		wallet.CancelSync()
 		handler.ConnectedPeers = 0
-		err := beeep.Notify("Sync Canceled", "wallet sync stopped until wallet reconnect", "assets/information.png")
+		err := beeep.Notify(notifyTitle, notifyMessage, "assets/information.png")
 		if err != nil {
 			log.Println("error initiating alert:", err.Error())
 		}
@@ -362,11 +372,11 @@ func newTransactionRow(transactionType, amount, fee, direction, status, date str
 	icon := canvas.NewImageFromResource(icons[transactionType])
 	icon.SetMinSize(fyne.NewSize(5, 20))
 	iconBox := widget.NewVBox(widgets.NewVSpacer(4), icon)
-	amountLabel := widget.NewLabel(amount)
-	feeLabel := widget.NewLabel(fee)
-	dateLabel := widget.NewLabel(date)
-	statusLabel := widget.NewLabel(status)
-	directionLabel := widget.NewLabel(direction)
+	amountLabel := widget.NewLabelWithStyle(amount, fyne.TextAlignTrailing, fyne.TextStyle{})
+	feeLabel := widget.NewLabelWithStyle(fee, fyne.TextAlignCenter, fyne.TextStyle{})
+	dateLabel := widget.NewLabelWithStyle(date, fyne.TextAlignCenter, fyne.TextStyle{})
+	statusLabel := widget.NewLabelWithStyle(status, fyne.TextAlignCenter, fyne.TextStyle{})
+	directionLabel := widget.NewLabelWithStyle(direction, fyne.TextAlignCenter, fyne.TextStyle{})
 	column := widget.NewHBox(iconBox, amountLabel, feeLabel, directionLabel, statusLabel, dateLabel)
 	return column
 }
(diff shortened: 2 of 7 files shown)


metaclips (changes requested):
image
There seems to be spacing between trailing and leading amount text, you should check how send page confirmation window implements amount texts https://github.com/raedahgroup/godcr/blob/0dc5f3cac669f276f1244f061976bdcf17212380/fyne/pages/handler/sendpagehandler/confirmationwindow.go#L47

image
We should add padding

image
Overview text should be titled
image
All header text should be titled.
image
Amount table contents should be right aligned
image
table contents should be centered

metaclips (commented):
You omitted wallet sync connect/disconnect button on overview page.

fyne/pages/handler/overview.go
```diff
@@ -0,0 +1,407 @@
...
+
+	for _, txn := range handler.Transactions {
+		amount := dcrutil.Amount(txn.Amount).String()
+		fee := dcrutil.Amount(txn.Fee).String()
+		timeDate := dcrlibwallet.ExtractDateOrTime(txn.Timestamp)
+		status := transactionStatus(height, txn)
+		transactionList = append(transactionList, newTransactionRow(transactionIcon(txn.Direction), amount, fee,
+			dcrlibwallet.TransactionDirectionName(txn.Direction), status, timeDate))
+	}
+	handler.Table.Append(transactionList...)
```
metaclips, line 141: You should try as much as possible not to use Append/Prepend table method so as to render pages faster, Instead create an iTable variable that would initialize a new table with heading and contents then pass the children of the iTable to that of handler.Table 

   var iTable widgets.Table{}

   iTable.NewTable(heading,contents)

   handler.Table.Result.Children = iTable.Result.Children

fyne/pages/overview.go
```diff
@@ -1,208 +1,188 @@
...
 
-// todo: display overview page (include sync progress UI elements)
-// todo: register sync progress listener on overview page to update sync progress views
+var overviewHandler = &handler.OverviewHandler{}
+
 func overviewPageContent(app *AppInterface) fyne.CanvasObject {
 	ov := &overview{}
 	ov.app = app
-	// app.Window.Resize(fyne.NewSize(650, 650))
+	//app.Window.Resize(fyne.NewSize(650, 680))
```
metaclips: remove commented code

fyne/pages/overview.go
```diff
@@ -1,208 +1,188 @@
...
-	return widget.NewHBox(widgets.NewHSpacer(values.Padding), ov.container(), widgets.NewHSpacer(values.Padding))
+
+	defer func() {
+		go overviewHandler.UpdateBalance(app.MultiWallet)
+		go overviewHandler.UpdateBlockStatusBox(app.MultiWallet)
+	}()
+	go func() {
+		time.Sleep(200 * time.Millisecond)
+		overviewHandler.UpdateWalletsSyncBox(app.MultiWallet)
+	}()
```
metaclips: Fyne renders widgets quickly so we shouldn't be using a go routine for speed, instead, isolate your code to see where the issue lies with slow rendering.

fyne/pages/handler/overview.go
```diff
@@ -0,0 +1,407 @@
...
+	BlockStatusSyncing   fyne.CanvasObject
+	Table                *widgets.Table
+	ConnectedPeersWidget *widget.Label
+	SyncStepWidget       *widget.Label
+	BlockHeadersWidget   *widget.Label
+	WalletSyncInfo       *fyne.Container
+	Scroll               *widget.ScrollContainer
+	CancelButton         *widget.Button
+
+	StepsChannel chan int32
```
metaclips: Do we need to make all of this exportable?

fyne/pages/display.go
```diff
@@ -56,6 +57,7 @@ func (app *AppInterface) DisplayMainWindow() {
 	app.Window.SetFixedSize(true)
 	app.Window.CenterOnScreen()
 	fyne.CurrentApp().Settings().SetTheme(theme.LightTheme())
+	go overviewHandler.PreserveSyncSteps()
```
metaclips, line 60: can we do away with this go routine?

metaclips (changes requested):

fyne/pages/display.go
oshorefueled: No we can't, did you check what the code does?

oshorefueled: > can we do away with this go routine?

No, we can't, did you check what that method does?

fyne/pages/handler/overview.go
oshorefueled: > Do we need to make all of this exportable?

Yes, we do.

oshorefueled:
> image
> There seems to be spacing between trailing and leading amount text, you should check how send page confirmation window implements amount texts
> 
> https://github.com/raedahgroup/godcr/blob/0dc5f3cac669f276f1244f061976bdcf17212380/fyne/pages/handler/sendpagehandler/confirmationwindow.go#L47
> 
> image
> We should add padding
> 
> image
> Overview text should be titled
> image
> All header text should be titled.
> image
> Amount table contents should be right aligned
> image
> table contents should be centered

Kindly clarify what you mean by titled.

fyne/pages/handler/overview.go
oshorefueled: the Append method and NewTable method both add data to the list the same way. Or is there something I'm missing? 

> You should try as much as possible not to use Append/Prepend table method so as to render pages faster, Instead create an iTable variable that would initialize a new table with heading and contents then pass the children of the iTable to that of handler.Table
> 
> var iTable widgets.Table{}
> 
> iTable.NewTable(heading,contents)
> 
> handler.Table.Result.Children = iTable.Result.Children

the Append method and NewTable method both add data to the list the same way. Or is there something I'm missing?

fyne/pages/overview.go
oshorefueled: apparently Fyne doesn't render a bunch of widgets that fast when they're outside a container.

metaclips:
> > image
> > There seems to be spacing between trailing and leading amount text, you should check how send page confirmation window implements amount texts
> > https://github.com/raedahgroup/godcr/blob/0dc5f3cac669f276f1244f061976bdcf17212380/fyne/pages/handler/sendpagehandler/confirmationwindow.go#L47
> > 
> > image
> > We should add padding
> > image
> > Overview text should be titled
> > image
> > All header text should be titled.
> > image
> > Amount table contents should be right aligned
> > image
> > table contents should be centered
> 
> Kindly clarify what you mean by titled.

First character should be upper case

fyne/pages/overview.go
metaclips: Can you be explicit on the specific widgets that are rendered slow?

oshorefueled:
Screenshot 2020-01-07 at 12 45 25 PM

fyne/pages/overview.go
oshorefueled: The box widgets render slow quite a number of content are added to its children and the UI is refreshed.

metaclips (changes requested):
UI Changes

Screenshot 2020-01-08 at 16 51 39

- We need to remove horizontal scroll, confirmed/pending text needs to be labeled

- Bottom padding doesn't seem right, can we add decred logo at the top of overview text?

- Table looks off, can we instead start with Direction Icon, Date,  Amount, Fee, Status ("Left Aligned") check below image for reference and also I guess Direction is same thing as # on the table
Screenshot 2020-01-08 at 16 51 39

- When sync completes and objects are hidden, the window size isn't calibrated and table centered
Screenshot 2020-01-08 at 16 16 00

- When wallet is disconnected sync steps and wallet info bar are shown
 Color when wallet is synced, syncing and not synced seems off
 Texts in bordered wallet info can be enlarged so as to be clear and border can be padded
Screenshot 2020-01-08 at 16 16 11

-   while syncing it doesn't report how many minutes left for sync to be done but instead there's a timer  and also the scroller housing wallet info moves back to the starting if scrolled
Screenshot 2020-01-08 at 17 55 41

fyne/pages/handler/overview.go
```diff
@@ -0,0 +1,440 @@
...
+	Table                 *widgets.Table
+	TableHeader           *widget.Box
+	ConnectedPeersWidget  *widget.Label
+	SyncStepWidget        *widget.Label
+	BlockHeadersWidget    *widget.Label
+	WalletSyncInfo        *fyne.Container
+	Scroll                *widget.ScrollContainer
+	SyncButton            *widget.Button
+
+	StepsChannel chan int32
```
metaclips, line 54: Do we need to make all variables exportable? can we only make variables that are dynamic exportable? we should try our best to bring about encapsulation of variables and functions/methods
This are the variables that are needed 
Sync label
Table widget
Time left label
Last time of block fetched
Balance label (we just pass the *fyne.Container object and then if we need to dynamically change this we can pass the values to the struct field "Objects" to dynamically change the balance label)
Button to enable/disable syncing
Steps label and Header/address discovery label
Connected peer count label
Block count label
contents in bordered wallet info (synced, block header fetched and syncing progress label) we can pass them as an array of *canvas.Text

You only need to make the functions/methods that initializes overview page and that are passed to txnotifier.go exportable in the handler package so that all other helper functions/methods in handler package are private to pages package bringing about encapsulation also, functions to be passed to txnotifier can be moved to the handler package.

fyne/pages/handler/overview.go
```diff
@@ -0,0 +1,440 @@
...
+		return "Synced", color.Gray{Y: 123}
+	}
+
+	return "Not Synced", color.Gray{Y: 123}
+}
+
+func (handler *OverviewHandler) SyncTrigger(wallet *dcrlibwallet.MultiWallet) {
+	var notifyTitle, notifyMessage string
+	synced, syncing := wallet.IsSynced(), wallet.IsSyncing()
+	if wallet.ConnectedPeers() > 0 && (synced || syncing) {
```
metaclips, line 218: What if there's no internet connection and connected peers is 0? I really dont think this conditions are necessary, only if wallet.IsSyncing or not should do

fyne/pages/display.go
metaclips: This isn't needed there are already methods in txnotifier that handles this
step 1/3 OnHeadersFetchProgress 
step 2/3 OnAddressDiscoveryProgress
step 3/3 OnHeadersRescanProgress
if all steps are completed we hide all objects that are dynamically changed by them

metaclips (changes requested):
When I created a new wallet and tried canceling sync, the window stalls and I have to force close the app also, progress bar isn't updated while syncing
sync/waiting for sync seems not to be accurate
Screenshot 2020-01-08 at 19 06 23
show/hide details button is missing 
Progress bar should be updated for all steps
Lastly can we do away with the goroutines used in overview.go? I still see no reason for this, we initialize all dynamic objects then pass them to the txnotifier which should be capable of handling all dynamic objects

